PR #13, opened 2025-04-02
Add opensearch index

### LINKED ISSUE #5: Add opensearch index
## tl;dr

feature flag: `opensearch`

The vector-store is rooted in a use the Usearch library. We want to support also Opensearch as  an indexing service. This issue should implement support for this.

## Design

The architecture of the service is based on the actor framework. The Usearch index functionality is implemented in the index actor. The Opensearch functionality should be implemented in a separate actor. API for both should be same, only a backend should be different. There should be only one running actor, there should be an environment value to choose which index functionality should be used. All additional parameters for Opensearch should be provided by environment values. For integration tests use a mockup of Opensearch. Real integration tests should be done in a scylladb repository.

## PRs

- [x] #34 
- [x] Refactor the Usearch index actor. Copy a backend from index.rs into index/usearch.rs.
  - [x] #36
- [x] Create empty backend for the Opensearch. Create backend in index/opensearch.rs.
  - [x] #133
  - [x] #125 
- [x] Implement Index::Add for the Opensearch.
  - [x] #126  
- [x] Implement Index::Ann for the Opensearch.
  - [x] #141 
- [x] Implement other necessary functions for the OpenSearch.
  - [x] #140   
  - [x] #144 
- [x] Create integration tests with mockup
  - [x] #145  
- [ ] Remove `opensearch` flag feature
  - [ ] #147  

Migrated to Jira: [VS-74](https://scylladb.atlassian.net/browse/VS-74)

[VS-74]: https://scylladb.atlassian.net/browse/VS-74?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

QuerthDP:
Fixes #5


swasik:
@QuerthDP could we already share it with @ewienik for the first round of review?

swasik:
Why was it closed?

swasik:
Thanks for the explanation! BTW, it's worth leaving one sentence comment when closing the issue without clear reason so that we can recall what happened.

Issue #3, opened 2025-03-19
Add integration tests for devs


ewienik:
## tl;dr

feature flag: `integ_tests` 

There should be integration tests for the development. Integration test should use std test infrastructure from Rust. This issue will be a poc for feature development methodology using compile time rust features (similar to unstable features in rust compiler): the reason is to try to make smaller PRs which could land faster in master. The last stage of work on the issue there will be stabilizing feature by removing compilation flag from the source code.

## Design

We should run integration tests mocking up external services. Full integration between scylladb and vector-store should be done in scylladb repository tests. We need to mockup scylladb cql service and we should do this by mockuping internal api to db. We could create external service, but we don't have such service. Mockuping rust driver to scylla seems to be too big effort to develop.

There should be specific db actor which we could mockup by providing specific actor for our tests. The crate should export running service with an ability to provide own db actor.

## PRs

- [x] #6
- [x] #11 
- [x] #12 
- [x] Move a monitor_indexes `struct Db` into the db actor.
  - [x] #14 
  - [x] #15 
  - [x] #17 
  - [x] #18
  - [x] #19  
- [x] Move a modify_indexes `struct Db` into the db actor.
  - [x] #20  
  - [x] #25 
- [x] Move a monitor_items `struct Db` into the db actor.
  - [x] #26 
  - [x] #29 
  - [x] #30 
  - [x] #31 
- [x] Refactor db actor before implementing tests
  - [x] #32 
  - [x] #35 
  - [x] #37 
  - [x] #38 
  - [x] #39 
- [x] #40 
- [ ] #41 
- [x] ~Remove `integ_tests` flag feature~

dani-tweig:
Migrated to Jira: [VS-146](https://scylladb.atlassian.net/browse/VS-146)

PR #19, opened 2025-04-03
Move a monitor_indexes::Db::get_index_target_type into a db index actor

ewienik:
This is a part of #3.

The change moves get_index_target_type method from monitor_indexes into a db actor. This will also remove monitor_indexes empty Db struct. Next paches will move other methods into a db actor.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -7,6 +7,7 @@ use crate::db_index;
 use crate::db_index::DbIndex;
 use crate::ColumnName;
 use crate::Connectivity;
+use crate::Dimensions;
 use crate::ExpansionAdd;
 use crate::ExpansionSearch;
 use crate::IndexId;
@@ -16,10 +17,12 @@ use crate::KeyspaceName;
 use crate::TableName;
 use anyhow::Context;
 use futures::TryStreamExt;
+use regex::Regex;
 use scylla::client::session::Session;
 use scylla::statement::prepared::PreparedStatement;
 use scylla::value::CqlTimeuuid;
 use std::collections::BTreeMap;
+use std::num::NonZeroUsize;
 use std::sync::Arc;
 use tokio::sync::mpsc;
 use tokio::sync::oneshot;
@@ -49,6 +52,13 @@ pub(crate) enum Db {
         tx: oneshot::Sender<anyhow::Result<Option<IndexVersion>>>,
     },
 
+    GetIndexTargetType {
+        keyspace: KeyspaceName,
+        table: TableName,
+        target_column: ColumnName,
+        tx: oneshot::Sender<anyhow::Result<Option<Dimensions>>>,
+    },
+
     GetIndexParams {
         id: IndexId,
         #[allow(clippy::type_complexity)]
@@ -68,6 +78,13 @@ pub(crate) trait DbExt {
         keyspace: KeyspaceName,
         index: TableName,
     ) -> anyhow::Result<Option<IndexVersion>>;
+
+    async fn get_index_target_type(
+        &self,
+        keyspace: KeyspaceName,
+        table: TableName,
+        target_column: ColumnName,
+    ) -> anyhow::Result<Option<Dimensions>>;
 
     async fn get_index_params(
         &self,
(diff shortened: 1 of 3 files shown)


src/db.rs
```diff
@@ -49,6 +52,13 @@ pub(crate) enum Db {
         tx: oneshot::Sender<anyhow::Result<Option<IndexVersion>>>,
     },
 
+    GetIndexTargetType {
+        keyspace: KeyspaceName,
+        table: TableName,
+        target: ColumnName,
```
swasik, line 58: I would call it column instead of target. Then the target is keyspace.table.column and it will make clear what this field maps to.

ewienik: The `column` name is a general one. In this case it should mean a column which is a target for indexing. Scylla labels it as `target`, so that the name `target` is consistent with ScyllaDB:

```
cqlsh:vector_benchmark> SELECT * FROM system_schema.indexes WHERE kind = 'CUSTOM' ALLOW FILTERING;                       
                                                                                                                         
 keyspace_name    | table_name     | index_name | kind   | options                                                       
------------------+----------------+------------+--------+---------------------------------------------------------------
 vector_benchmark | vector_indexes | ann2_index | CUSTOM | {'class_name': 'dummy-vector-backend', 'target': 'dimension'} 
 vector_benchmark |   vector_items |  ann_index | CUSTOM | {'class_name': 'dummy-vector-backend', 'target': 'embedding'} 
```

swasik: So maybe target_column? I do not suppose we will have different type of targets for  vector data and this will make the code much easier to understand.

ewienik:
Uploaded v2
<details>
<summary>Range-diff against v1:</summary>

```
1:  3b67252 ! 1:  6023a71 Move a monitor_indexes::Db::get_index_target_type into a db index actor
    @@ src/db.rs: pub(crate) enum Db {
     +    GetIndexTargetType {
     +        keyspace: KeyspaceName,
     +        table: TableName,
    -+        target: ColumnName,
    ++        target_column: ColumnName,
     +        tx: oneshot::Sender<anyhow::Result<Option<Dimensions>>>,
     +    },
     +
    @@ src/db.rs: pub(crate) trait DbExt {
     +        &self,
     +        keyspace: KeyspaceName,
     +        table: TableName,
    -+        target: ColumnName,
    ++        target_column: ColumnName,
     +    ) -> anyhow::Result<Option<Dimensions>>;
     +
          async fn get_index_params(
    @@ src/db.rs: impl DbExt for mpsc::Sender<Db> {
     +        &self,
     +        keyspace: KeyspaceName,
     +        table: TableName,
    -+        target: ColumnName,
    ++        target_column: ColumnName,
     +    ) -> anyhow::Result<Option<Dimensions>> {
     +        let (tx, rx) = oneshot::channel();
     +        self.send(Db::GetIndexTargetType {
     +            keyspace,
     +            table,
    -+            target,
    ++            target_column,
     +            tx,
     +        })
     +        .await?;
    @@ src/db.rs: async fn process(statements: Arc<Statements>, msg: Db) {
     +        Db::GetIndexTargetType {
     +            keyspace,
     +            table,
    -+            target,
    ++            target_column,
     +            tx,
     +        } => tx
     +            .send(
     +                statements
    -+                    .get_index_target_type(keyspace, table, target)
    ++                    .get_index_target_type(keyspace, table, target_column)
     +                    .await,
     +            )
     +            .unwrap_or_else(|_| warn!("db::process: Db::GetIndexVersion: unable to send response")),
    @@ src/db.rs: impl Statements {
     +        &self,
     +        keyspace: KeyspaceName,
     +        table: TableName,
    -+        target: ColumnName,
    ++        target_column: ColumnName,
     +    ) -> anyhow::Result<Option<Dimensions>> {
     +        Ok(self
     +            .session
     +            .execute_iter(
     +                self.st_get_index_target_type.clone(),
    -+                (keyspace, table, target),
    ++                (keyspace, table, target_column),
     +            )
     +            .await?
     +            .rows_stream::<(String,)>()?
```

src/db.rs
ewienik: fixed in v2

swasik (approved):

PR #20, opened 2025-04-04
Move a modify_indexes::Db::update_items_count into a db actor

ewienik:
This is a part of #3.

The change moves update_items_count method from modify_indexes into a db actor. Next paches will move other methods into a db actor.


src/db.rs
```diff
@@ -410,4 +444,21 @@ impl Statements {
             .try_next()
             .await?)
     }
+
+    const UPDATE_ITEMS_COUNT: &str = "
```
swasik, line 448: Nit: I already noticed this in one of the previous PRs but did not comment as the CQL will be changed.

I think that it would be nice to have something pointing to the fact that this is CQL statement in the name of the constant, e.g. UPDATE_ITEMS_COUNT_CQL (here and for other CQLs.)

swasik (approved):

ewienik:
Uploaded v2

<details>
<summary>Range-diff against v1:</summary>

```
1:  38ca4a7 ! 1:  84d1fd6 Move a modify_indexes::Db::update_items_count into a db actor
    @@ src/db.rs: impl Statements {                                                    
                      .context("ST_GET_INDEX_PARAMS")?,                                
~~~~~~                                                                                 
     +            st_update_items_count: session                                       
    -+                .prepare(Self::UPDATE_ITEMS_COUNT)                               
    ++                .prepare(Self::ST_UPDATE_ITEMS_COUNT)                            
     +                .await                                                           
    -+                .context("UPDATE_ITEMS_COUNT")?,                                 
    ++                .context("ST_UPDATE_ITEMS_COUNT")?,                              
     +                                                                                 
                  session,                                                             
              })                                                                       
    @@ src/db.rs: impl Statements {                                                    
                  .await?)                                                             
          }                                                                            
     +                                                                                 
    -+    const UPDATE_ITEMS_COUNT: &str = "                                           
    ++    const ST_UPDATE_ITEMS_COUNT: &str = "                                        
     +        UPDATE vector_benchmark.vector_indexes                                   
     +            SET indexed_elements_count = ?                                       
     +            WHERE id = ?                                                         
```

src/db.rs
ewienik: fixed in v2

swasik (approved):

Issue #4, opened 2025-03-19
Add integration tests for CI


ewienik:
## tl;dr

~feature flag: `ci_tests` ~

This issue depends on #3. CI should use github actions. On the master CI should check the product without any features in progress. PRs should be able to enable checking product with specified feature.

## PRs

- ~Add `ci_tests` flag feature~
- Add CI configuration for master
- Add CI configuration for PR
- ~Remove `ci_tests` flag feature~

## Final tests

After merging with master needs to check CI master actions.

dani-tweig:
Migrated to Jira: [VS-145](https://scylladb.atlassian.net/browse/VS-145)

PR #21, opened 2025-04-04
Introduce a github rust worflow for CI tests

ewienik:
CI should use github actions for static checks and for tests. This change introduces fmt, clippy and build checks as static_checks and running all tests as tests. It contains also all necessary fixes in formatting to pass these tests.

Fixes: #4

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,0 +1,57 @@
+name: Rust
+
+on:
+  push:
+    branches:
+    - 'master'
+    - '**'
+
+  pull_request:
+    branches:
+    - 'master'
+
+env:
+  CARGO_TERM_COLOR: always
+  RUSTFLAGS: -Dwarnings
+  RUST_BACKTRACE: full
+
+jobs:
+  static_checks:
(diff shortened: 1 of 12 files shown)


.github/workflows/rust.yml
```diff
@@ -0,0 +1,60 @@
...
+
+    - name: Format check
+      run: cargo fmt --all -- --check
+
+    - name: Clippy check
+      run: cargo clippy --all-targets
+
+    - name: Build
+      run: cargo build --verbose
+
```
Lorak-mmk, line 46: What is the purpose of having a build step?
- It is not useful for catching errors - earlier step (clippy) takes care of it
- Artifact is not saved anywhere, so this is also not the use case for `build` here.

We've removed a similar step some time ago in Rust Driver because it was significantly slowing down CI.

ewienik: You are right, the artifact of build step is not used anywhere later. The building step is also run in `cargo test`. Fixed now.

Lorak-mmk (approved):

PR #29, opened 2025-04-07
Move a monitor_items::Db::get_items into a db index actor

ewienik:
This is a part of #3.

The change moves get_items method from monitor_tems into a db index actor. Next paches will move other methods into a db index actor.

### DIFF
--- a/src/db_index.rs
+++ b/src/db_index.rs
@@ -4,6 +4,7 @@
  */
 
 use crate::ColumnName;
+use crate::Embeddings;
 use crate::IndexMetadata;
 use crate::Key;
 use crate::TableName;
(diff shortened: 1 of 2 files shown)


src/db_index.rs
```diff
@@ -86,6 +111,15 @@ impl Statements {
                 .await
                 .context("get_processed_ids_query")?,
 
+            st_get_items: session
+                .prepare(Self::get_items_query(
+                    &metadata.table_name,
+                    &metadata.key_name,
+                    &metadata.target_name,
```
swasik, line 118: We changed this to target_column.

ewienik: Because it is change in `IndexMetadata` struct, so as to be done separately. It will touch a lot of places. I am going to make this change after `db_index` refactoring.

PR #34, opened 2025-04-08
Add opensearch feature flag

QuerthDP:
This is a part of #5.

This change introduce `opensearch` feature flag, adds OpenSearch dependency and updates `Cargo.lock` file accordingly.


Cargo.toml
```diff
@@ -8,13 +8,18 @@ name = "vector-store"
...
+opensearch = ["dep:opensearch"]
+
 [dependencies]
 anyhow = "1.0.97"
 axum = { version = "0.8.1", features = ["macros"] }
 derive_more = { version = "2.0.1", features = ["full"] }
 dotenvy = "0.15.7"
 futures = "0.3.31"
 itertools = "0.14.0"
+opensearch = { version = "1.0.0", optional = true }
```
swasik, line 22: The newest version is 2.19.1 do why do we use 1.0.0?

QuerthDP: I haven't thought of checking the latest version.

It was "recommended" in the OpenSearch Rust client [documentation](https://opensearch.org/docs/latest/clients/rust/), probably just for the example.

We can update to the newest I think.

swasik: Ok, let us do it.

QuerthDP: One more question: how do you know that 2.19.1 is the latest version?
I've checked many sources ([docs.rs](https://docs.rs/opensearch/latest/opensearch/), [github](https://github.com/opensearch-project/opensearch-rs), [crates.io](https://crates.io/crates/opensearch)) and all of them state that 2.3.0 is the latest. Maybe I don't see something..

swasik: But maybe Rust versioning is different :-)

QuerthDP: Yes it is 🙂
Will change to 2.3.0

QuerthDP:
Changelog:
- updated opensearch dependency version from 1.0.0 to 2.3.0 (latest)

swasik:
LGTM, but let us wait for @ewienik as I am not Rust expert (or even regular user).

swasik (approved):

PR #36, opened 2025-04-09
Move index.rs to index/usearch.rs

QuerthDP:
Create new index module with it's proper mod.rs file. Move and rename index.rs file to make a setup for opensearch.rs index file upcoming.

### DIFF
--- a/src/index/mod.rs
+++ b/src/index/mod.rs
@@ -1,0 +1,5 @@
+pub mod usearch;
+
+pub use usearch::Index;
+pub use usearch::IndexExt;
+pub use usearch::new;
--- a/src/index/usearch.rs
+++ b/src/index/usearch.rs
@@ -42,7 +42,7 @@ const RESERVE_INCREMENT: usize = 1000000;
 // The ratio was taken for initial benchmarks
 const RESERVE_THRESHOLD: usize = RESERVE_INCREMENT / 3;
 
-pub(crate) enum Index {
+pub enum Index {
     Add {
         key: Key,
         embeddings: Embeddings,
@@ -54,7 +54,7 @@ pub(crate) enum Index {
     },
 }
 
-pub(crate) trait IndexExt {
+pub trait IndexExt {
     async fn add(&self, key: Key, embeddings: Embeddings);
     async fn ann(
         &self,
@@ -86,7 +86,7 @@ impl IndexExt for mpsc::Sender<Index> {
     }
 }
 
-pub(crate) fn new(
+pub fn new(
     id: IndexId,
     db: mpsc::Sender<Db>,
     dimensions: Dimensions,


src/index/mod.rs
```diff
@@ -0,0 +1,5 @@
+pub mod usearch;
```
ewienik, line 1: `src/index/mod.rs` could be `src/index.rs` - file with basename same as directory has the semantic of `mod.rs`.

mod usearch should be not public - it is optional dependency, it should be visible in index module when feature is enabled (use `#[cfg(not(feature = "opensearch"))]`

src/index/usearch.rs
```diff
@@ -42,7 +42,7 @@ const RESERVE_INCREMENT: usize = 1000000;
 // The ratio was taken for initial benchmarks
 const RESERVE_THRESHOLD: usize = RESERVE_INCREMENT / 3;
 
-pub(crate) enum Index {
+pub enum Index {
```
ewienik, line 45: This is interface to index actor. It should be defined in `actor.db`
The visibility `pub(crate)` should be enough

src/index/usearch.rs
```diff
@@ -54,7 +54,7 @@ pub(crate) enum Index {
     },
 }
 
-pub(crate) trait IndexExt {
+pub trait IndexExt {
```
ewienik, line 57: This is interface to index actor. It should be defined in `actor.db`
The visibility `pub(crate)` should be enough

src/index/usearch.rs
```diff
@@ -86,7 +86,7 @@ impl IndexExt for mpsc::Sender<Index> {
     }
 }
 
-pub(crate) fn new(
+pub fn new(
```
ewienik, line 89: The visibility `pub(crate)` should be enough

src/index/mod.rs
```diff
@@ -0,0 +1,5 @@
...
+
+pub use usearch::Index;
+pub use usearch::IndexExt;
```
ewienik, line 4: Both struct are interface to index actor, should be defined in `actor.rs`

src/index/mod.rs
```diff
@@ -0,0 +1,5 @@
...
+pub use usearch::new;
```
ewienik, line 5: This function should depends on feature defined (opensearch or usearch) - it should forward call to the `new` function from specified backend for this actor.

ewienik (changes requested):

src/index/mod.rs
QuerthDP: As mentioned before, I wanted to wrap it in `#[cfg...]` macro in upcoming commit(s).

src/index/mod.rs
ewienik: > Why is it better to use the index.rs file instead of index/mod.rs?
Technically - there is no difference (It seems to be my preference). I think we should stay with consistency with rust driver and I've checked that it uses `module/mod.rs` template, so let's leave this as you wrote it.

QuerthDP:
Changelog:
- rebased on master
- moved Index, IndexExt to `actor.rs`
- wrapped non-common elements in cfg feature macro

smoczy123:
Changelog:

- added opensearch.rs with empty functions
- ran cargo fmt

smoczy123:
Changelog:

- Added copyright to mod.rs

ewienik (approved):

PR #89, opened 2025-04-14
Refactor usearch index using task per message

ewienik:
This is a part of #23.

This patch introduce internal task per each message received by usearch actor to simplify usage on consumer side.

Issue #23, opened 2025-04-07
Use the ScyllaDB table's multicolumn primary key


ewienik:
Current version of vector-store has been  built for a specific benchmark framework and depends on single i64 as PK in a table with embeddings. Real tables could have multicolumn PK (partition + cluster keys' columns). Vector-store should map ScyllaDB table's PK into internal key for vectors internally.

## PRs

- #88 
- #89 
- #90 
- #91 
- #106 
- #92

dani-tweig:
Migrated to Jira: [VS-143](https://scylladb.atlassian.net/browse/VS-143)

PR #90, opened 2025-04-14
Refactor db index reset and update items

ewienik:
This is a part of #23.

This is a preparation for multicolumn primary keys support. Previously getting or resettings embeddings in a table was based on single i64 primary key, so it could be done as a full scan. In a multicolumn primary keys it is more convenient to use a specific primary key in queries.

---

It depends on #89. Only the last commit is valid.


swasik (approved):

ewienik:
There are no changes in patch 6041f4d

swasik (approved):

PR #91, opened 2025-04-14
Introduce vector of i64s as a base for multicolumn primary keys

ewienik:
This is a part of #23.

This is first step in introduction multicolumn primary keys. This change introduces a vector storage of i64s for a primary key as a collection of several columns. Currently it will use only first item in this vector (the size will be always 1) - the one which is retrieved from IndexMetadata. In the next patch there will be support more dimensions. It changes HTTP API according to this change also.

### DIFF
--- a/src/httproutes.rs
+++ b/src/httproutes.rs
@@ -3,6 +3,7 @@
  * SPDX-License-Identifier: LicenseRef-ScyllaDB-Source-Available-1.0
  */
 
+use crate::ColumnName;
 use crate::Distance;
 use crate::Embeddings;
 use crate::IndexId;
@@ -10,9 +11,11 @@ use crate::Key;
 use crate::KeyspaceName;
 use crate::Limit;
 use crate::TableName;
+use crate::db_index::DbIndexExt;
 use crate::engine::Engine;
 use crate::engine::EngineExt;
 use crate::index::IndexExt;
+use anyhow::bail;
 use axum::Router;
 use axum::extract;
 use axum::extract::Path;
@@ -21,8 +24,11 @@ use axum::http::StatusCode;
 use axum::response;
 use axum::response::IntoResponse;
 use axum::response::Response;
+use itertools::Itertools;
+use std::collections::HashMap;
 use tokio::sync::mpsc::Sender;
 use tower_http::trace::TraceLayer;
+use tracing::error;
 use utoipa::OpenApi;
 use utoipa_axum::router::OpenApiRouter;
 use utoipa_axum::routes;
@@ -72,13 +78,13 @@ pub struct PostIndexAnnRequest {
 
 #[derive(serde::Deserialize, serde::Serialize, utoipa::ToSchema)]
 pub struct PostIndexAnnResponse {
-    pub keys: Vec<Key>,
+    pub primary_keys: HashMap<ColumnName, Vec<Key>>,
     pub distances: Vec<Distance>,
 }
 
 #[utoipa::path(
     post,
-    path = "/api/v1/indexes/{keyspace}/{table}/ann",
+    path = "/api/v1/indexes/{keyspace}/{index}/ann",
     description = "Ann search in the index",
     params(
         ("keyspace" = KeyspaceName, Path, description = "Keyspace name for the table to search"),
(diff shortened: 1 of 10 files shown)


src/httproutes.rs
```diff
@@ -95,15 +101,67 @@ async fn post_index_ann(
     Path((keyspace, index)): Path<(KeyspaceName, TableName)>,
```
swasik, line 101: In the line 87 above we have {keyspace} and {table} parameters. But in the params() section we have index parameter instead of table.

src/httproutes.rs
```diff
@@ -95,15 +101,67 @@ async fn post_index_ann(
...
     extract::Json(request): extract::Json<PostIndexAnnRequest>,
```
swasik, line 102: Above, why is index mapped to TableName? Do we assume that there is just one index per table?

src/httproutes.rs
```diff
@@ -95,15 +101,67 @@ async fn post_index_ann(
...
+        Err(err) => {
+            let msg = format!("index.ann request error: {err}");
+            error!("post_index_ann: {msg}");
+            (StatusCode::INTERNAL_SERVER_ERROR, msg).into_response()
+        }
+
+        Ok((primary_keys, distances)) => {
+            if primary_keys.len() != distances.len() {
+                let msg = format!(
+                    "wrong dimensions of an ann response: number of primary_keys = {}, number of distances = {}",
```
swasik, line 118: `wrong dimensions of an ann response: primary_keys.len() = {}, distances.len() = {}`?

There is just one primary key. 

Also `dimension` without s.

tests/integration/usearch.rs
```diff
@@ -70,9 +71,9 @@ async fn simple_create_search_delete_index() {
         &index.table_name,
         &index.target_column,
         vec![
-            (1.into(), vec![1., 1., 1.].into()),
-            (2.into(), vec![2., -2., 2.].into()),
-            (3.into(), vec![3., 3., 3.].into()),
+            (vec![1.into()].into(), vec![1., 1., 1.].into()),
```
swasik, line 74: Could you add some test with a primary key with more than 1 column?

src/httproutes.rs
ewienik: I am going to change `dimensions` into `size`. From the search function there could be several `primary_keys` - up to the query limit.

src/httproutes.rs
ewienik: I've created #105

swasik: Great, thanks!

tests/integration/usearch.rs
ewienik: This patch is for introduction vector of columns, but using only one of it. I am changing the description of this patch and providing intermediate patch to introduce several columns with i64 type.

ewienik:
Changlog for 60c4db7:
- Update patch description
- Update utoipa endpoint description
- Update logging messages

<details>

<summary>range diff</summary>

```
1:  6011f09 ! 1:  60c4db7 Implement multicolumn primary keys
    @@ Metadata
     Author: Pawel Pery <pawel.pery@scylladb.com>
     
      ## Commit message ##
    -    Implement multicolumn primary keys
    +    Introduce vector of i64s as a base for multicolumn primary keys
     
         This is a part of #23.
     
         This is first step in introduction multicolumn primary keys. This change
    -    introduces a primary key as a collection of several columns with
    -    limitation of support for i64 type only. It changes HTTP API according
    -    to this change also. Internally it use only one column in the primary
    -    key collection - the one which is retrieved from IndexMetadata.
    +    introduces a vector storage of i64s for a primary key as a collection of
    +    several columns. Currently it will use only first item in this vector
    +    (the size will be always 1) - the one which is retrieved from
    +    IndexMetadata. In the next patch there will be support more dimensions.
    +    It changes HTTP API according to this change also.
     
      ## src/db_index.rs ##
     @@
    @@ src/httproutes.rs: pub struct PostIndexAnnRequest {
          pub distances: Vec<Distance>,
      }
      
    + #[utoipa::path(
    +     post,
    +-    path = "/api/v1/indexes/{keyspace}/{table}/ann",
    ++    path = "/api/v1/indexes/{keyspace}/{index}/ann",
    +     description = "Ann search in the index",
    +     params(
    +         ("keyspace" = KeyspaceName, Path, description = "Keyspace name for the table to search"),
     @@ src/httproutes.rs: async fn post_index_ann(
          Path((keyspace, index)): Path<(KeyspaceName, TableName)>,
          extract::Json(request): extract::Json<PostIndexAnnRequest>,
    @@ src/httproutes.rs: async fn post_index_ann(
     +        Ok((primary_keys, distances)) => {
     +            if primary_keys.len() != distances.len() {
     +                let msg = format!(
    -+                    "wrong dimensions of an ann response: number of primary_keys = {}, number of distances = {}",
    ++                    "wrong size of an ann response: number of primary_keys = {}, number of distances = {}",
     +                    primary_keys.len(),
     +                    distances.len()
     +                );
    @@ src/httproutes.rs: async fn post_index_ann(
     +                            .map(|primary_key| {
     +                                if primary_key.0.len() != primary_key_columns.len() {
     +                                    bail!(
    -+                                        "wrong dimensions of a primary key: {}, {}",
    ++                                        "wrong size of a primary key: {}, {}",
     +                                        primary_key_columns.len(),
     +                                        primary_key.0.len()
     +                                    );
```

swasik (approved):

PR #92, opened 2025-04-14
Implement CQL types for a primary key

ewienik:
This is a part of #23.

This change implements support for different CQL types for a primary key. Specific actor for index must support internal bimap between general primary key with cql values and internal key representation. This change provides custom DeserializeRow, as scylla driver's Row doesn't support a vector type. It updates HTTP API to support general json value as a value of a primary key's column.

Fixes: #23


src/db_index.rs
```diff
@@ -152,6 +159,61 @@ async fn process(statements: Arc<Statements>, msg: DbIndex) {
...
+
+impl<'frame, 'metadata> DeserializeRow<'frame, 'metadata> for PrimaryKeyWithEmbeddings {
+    fn type_check(specs: &[ColumnSpec]) -> Result<(), TypeCheckError> {
+        if specs.len() < 2 {
+            return Err(TypeCheckError::new(
+                DeserializeError::InvalidQuerySelectLength,
+            ));
+        }
+        let ColumnType::Vector { typ, .. } = specs.last().unwrap().typ() else {
+            return Err(TypeCheckError::new(DeserializeError::InvalidEmbeddingsType));
```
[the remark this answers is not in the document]
ewienik: `type` is a keyword in Rust so the common way to avoid such names as functions or variables is to use `typ` name.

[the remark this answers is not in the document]
swasik: I do not like it :-) But if this is standard used over Rust, I will not fight for change.

src/db_index.rs
```diff
@@ -152,6 +159,61 @@ async fn process(statements: Arc<Statements>, msg: DbIndex) {
...
+enum DeserializeError {
+    #[error("Query for primary key & embeddings should contains at least two elements")]
+    InvalidQuerySelectLength,
+    #[error("Invalid embeddings type")]
+    InvalidEmbeddingsType,
+}
+
+struct PrimaryKeyWithEmbeddings {
+    primary_key: PrimaryKey,
+    embeddings: Embeddings,
```
swasik, line 172: For some time already I was wondering what is the correct form - embeddings vs embedding. And it looks like (see https://tanelpoder.com/posts/embedding-vectors-vs-vector-embeddings/) **embedding** will be a correct form here when we speak about single vector. I think it is worth to update it but I will leave you decision if you want to do it here or in the separate GH issue.

src/db_index.rs
```diff
@@ -152,6 +159,61 @@ async fn process(statements: Arc<Statements>, msg: DbIndex) {
     }
 }
 
+#[derive(thiserror::Error, Debug)]
+enum DeserializeError {
+    #[error("Query for primary key & embeddings should contains at least two elements")]
```
swasik, line 164: should contain

src/db_index.rs
```diff
@@ -163,7 +225,21 @@ struct Statements {
 
 impl Statements {
     async fn new(session: Arc<Session>, metadata: IndexMetadata) -> anyhow::Result<Self> {
-        let primary_key_columns = metadata.key_names;
+        let cluster_state = session.get_cluster_state();
+        let table = cluster_state
+            .get_keyspace(metadata.keyspace_name.as_ref())
+            .ok_or_else(|| anyhow!("keyspace {} not exists", metadata.keyspace_name))?
```
swasik, line 231: does not exist (here and below)

src/httproutes.rs
```diff
@@ -165,3 +172,110 @@ async fn post_index_ann(
...
+                .expect("CqlValue::Timestamp should be correct")
+                .format(&Iso8601::DEFAULT)
+                .expect("OffsetDateTime should be correct"),
+        ),
+
+        _ => unimplemented!(),
+    }
+}
+
+#[cfg(test)]
```
swasik, line 221: Should not tests go in a separate file?

src/db_index.rs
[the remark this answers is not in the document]
ewienik: `ColumnType::Vector::type` comes from scylla driver :-)

src/httproutes.rs
ewienik: > The convention is to create a module named tests in each file to contain the test functions and to annotate the module with cfg(test).

https://doc.rust-lang.org/book/ch11-03-test-organization.html#unit-tests

src/db_index.rs
ewienik: Created #107

ewienik:
Changelog for dc57386:

- fixes logging messages

<details>
<summary>range diff</summary>

```
1:  c70d510 ! 1:  dc57386 Implement CQL types for a primary key
    @@ src/db_index.rs: async fn process(statements: Arc<Statements>, msg: DbIndex) {
      
     +#[derive(thiserror::Error, Debug)]
     +enum DeserializeError {
    -+    #[error("Query for primary key & embeddings should contains at least two elements")]
    ++    #[error("Query for primary key & embeddings should contain at least two elements")]
     +    InvalidQuerySelectLength,
     +    #[error("Invalid embeddings type")]
     +    InvalidEmbeddingsType,
    @@ src/db_index.rs: struct Statements {
     +        let cluster_state = session.get_cluster_state();
     +        let table = cluster_state
     +            .get_keyspace(metadata.keyspace_name.as_ref())
    -+            .ok_or_else(|| anyhow!("keyspace {} not exists", metadata.keyspace_name))?
    ++            .ok_or_else(|| anyhow!("keyspace {} does not exist", metadata.keyspace_name))?
     +            .tables
     +            .get(metadata.table_name.as_ref())
    -+            .ok_or_else(|| anyhow!("table {} not exists", metadata.table_name))?;
    ++            .ok_or_else(|| anyhow!("table {} does not exist", metadata.table_name))?;
     +
     +        let primary_key_columns = table
     +            .partition_key
```

src/db_index.rs
ewienik: fixed in dc57386

src/db_index.rs
ewienik: fixed in dc57386

swasik (approved):

Issue #57, opened 2025-04-14
Support setting index parametrs and similarity function in CREATE INDEX


swasik:
Parameters should be set using `WITH OPTIONS` syntax, e.g.: 

`WITH OPTIONS = { 'similarity_function': 'DOT_PRODUCT' }; `
 
- Supported similarity functions are: DOT_PRODUCT, COSINE, or EUCLIDEAN. Default is COSINE.

- Vectors will not be normalized so if users want to use quick DOT_PRODUCT function, they have to insert normalized vectors (of length 1)

- We should support options that are [supported](https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/index/sai/disk/v1/IndexWriterConfig.java#L52) but not documented for Cassandra. In our case, they should be documented. Default values to be validated during performance tests.
  - maximum_node_connections (parameter m in HNSW) - max 512, default 16
  - construction_beam_width (parameter ef_construct) - max 4096, default 128
 
- We should support options that are not supported by Cassandra - we should introduce new name for them:
  - search_beam_width - parameter ef_search used during searching for similar vectors, max 4096, default 128 - even though it is used at search time some indexes such as usearch require this at index creation time

Migrated to Jira: [VS-55](https://scylladb.atlassian.net/browse/VS-55)

[VS-55]: https://scylladb.atlassian.net/browse/VS-55?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

dani-tweig:
Migrated to Jira: [VS-125](https://scylladb.atlassian.net/browse/VS-125)

dani-tweig:
Migrated to Jira: [VS-450](https://scylladb.atlassian.net/browse/VS-450)

dani-tweig:
Closing this issue as it was moved to Jira.
Please continue the thread in [VS-55](https://scylladb.atlassian.net/browse/VS-55)

Issue #54, opened 2025-04-14
Implement full scan when the index is created to index the whole table


swasik:

ewienik:
Current list of PRs:
- #108 
- #109 
- #110 
- #111 
- #112 
- #113 
- #114

dani-tweig:
Migrated to Jira: [VS-128](https://scylladb.atlassian.net/browse/VS-128)

PR #111, opened 2025-04-22
Remove temporary storage of index parameters

ewienik:
This is a part of #54.

The old workaround for storing metadata of an index in specific table in the ScyllaDB has been removing. This patch removes getting parameters of an index from specific table and replace them with default values. Setting index parameters from the custom index in CREATE INDEX will be a part of #57.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -9,7 +9,6 @@ use crate::DbCustomIndex;
 use crate::Dimensions;
 use crate::ExpansionAdd;
 use crate::ExpansionSearch;
-use crate::IndexId;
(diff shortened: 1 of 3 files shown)


src/db.rs
```diff
@@ -70,18 +68,10 @@ pub enum Db {
     },
 
     GetIndexParams {
```
swasik, line 70: Typos in the PR description: has been removing instead of has been removed, replace instead of replaces.

ewienik: Fixed in 4bcb54d

swasik (approved):

PR #112, opened 2025-04-22
Improve logging levels and messages

ewienik:
This is a part of #54.

This patch cleans up logging levels - error, warn & info should be used with messages meaningful for administrators, debug & trace could store more debugging information, useful rather for developers.

### DIFF
--- a/src/monitor_items.rs
+++ b/src/monitor_items.rs
@@ -3,6 +3,7 @@
  * SPDX-License-Identifier: LicenseRef-ScyllaDB-Source-Available-1.0
  */
 
+use crate::IndexId;
 use crate::db_index::DbIndex;
 use crate::db_index::DbIndexExt;
 use crate::index::Index;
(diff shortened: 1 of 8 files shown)


src/monitor_items.rs
```diff
@@ -13,30 +14,38 @@ use tokio::sync::mpsc::Receiver;
...
-        async move {
-            initial_read(&mut rx, &db_index, &index)
-                .await
-                .unwrap_or_else(|err| {
-                    warn!("monitor_items: unable to do initial read: {err}");
-                });
+        {
+            let id = id.clone();
+            async move {
+                debug!("starting");
```
swasik, line 35: starting what?

src/monitor_items.rs
```diff
@@ -13,30 +14,38 @@ use tokio::sync::mpsc::Receiver;
...
-            while rx.recv().await.is_some() {}
+                initial_read(&mut rx, &db_index, &index)
+                    .await
+                    .unwrap_or_else(|err| {
+                        warn!("unable to do initial read from db for {id}: {err}");
+                    });
+
+                while rx.recv().await.is_some() {}
+
+                debug!("finished");
```
swasik, line 45: finished what?

src/monitor_items.rs
ewienik: There is a debug_span around these events, so eventually there will be a log line like this: `...: monitor items{id}: starting`

```
     tokio::spawn(                                         
         async move {                                      
             debug!("starting");                           
...                                 
             debug!("finished");                           
         }                                                 
         .instrument(debug_span!("monitor items", "{id}")),
     );                                                    
```

swasik (approved):

PR #113, opened 2025-04-22
Refactor reading embeddings as a channel

ewienik:
This is a part of #54.

This patch creates a channel of embeddings from db index while constructing its owner, instead of stream returned from API. There should be the only one channel/stream in the lifetime of db index, so it is better to create it during construction. A mpsc channel is better solution for concurrent scan ranges introduces in the next patch.

There should be also a limitation of concurrent requests by the index actor, so there is implemented a semaphore to limit that.

### DIFF
--- a/src/db_index.rs
+++ b/src/db_index.rs
@@ -4,6 +4,7 @@
  */
 
 use crate::ColumnName;
+use crate::DbEmbeddings;
 use crate::Embeddings;
 use crate::IndexMetadata;
 use crate::KeyspaceName;
@@ -11,9 +12,9 @@ use crate::PrimaryKey;
 use crate::TableName;
 use anyhow::Context;
 use anyhow::anyhow;
+use futures::Stream;
 use futures::StreamExt;
 use futures::TryStreamExt;
-use futures::stream::BoxStream;
 use itertools::Itertools;
 use scylla::client::session::Session;
 use scylla::cluster::metadata::ColumnType;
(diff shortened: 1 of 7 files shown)


src/index/usearch.rs
```diff
@@ -86,14 +87,19 @@ pub(crate) fn new(
             // Incremental key for a usearch index
             let usearch_key = Arc::new(AtomicU64::new(0));
 
+            let semaphore = Arc::new(Semaphore::new(rayon::current_num_threads() * 2));
```
swasik, line 90: Why do you multiply by 2? Would be worth adding comment or constant.

src/db_index.rs
```diff
@@ -72,29 +60,50 @@ impl DbIndexExt for mpsc::Sender<DbIndex> {
 pub(crate) async fn new(
     db_session: Arc<Session>,
     metadata: IndexMetadata,
-) -> anyhow::Result<mpsc::Sender<DbIndex>> {
+) -> anyhow::Result<(mpsc::Sender<DbIndex>, mpsc::Receiver<DbEmbeddings>)> {
     let id = metadata.id();
     let statements = Arc::new(Statements::new(db_session, metadata).await?);
-    let (tx, mut rx) = mpsc::channel(10);
+    let (tx_index, mut rx_index) = mpsc::channel(10);
```
swasik, line 66: Could you use constant instead of 10 for channel size? Maybe even on global (it is used in usearch as well).

ewienik:
Changelog for 78f3446:
- add const for CHANNEL_SIZE in db_index
- add comment for the Semaphore size

<details>
<summary>range diff</summary>

```
1:  1721895 ! 1:  78f3446 Refactor reading embeddings as a channel
    @@ src/db_index.rs: impl DbIndexExt for mpsc::Sender<DbIndex> {
          let id = metadata.id();
          let statements = Arc::new(Statements::new(db_session, metadata).await?);
     -    let (tx, mut rx) = mpsc::channel(10);
    -+    let (tx_index, mut rx_index) = mpsc::channel(10);
    -+    let (tx_embeddings, rx_embeddings) = mpsc::channel(10);
    ++
    ++    // TODO: The value of channel size was taken from initial benchmarks. Needs more testing
    ++    const CHANNEL_SIZE: usize = 10;
    ++    let (tx_index, mut rx_index) = mpsc::channel(CHANNEL_SIZE);
    ++    let (tx_embeddings, rx_embeddings) = mpsc::channel(CHANNEL_SIZE);
    ++
     +    let mut items = statements.initial_scan().await?;
          tokio::spawn(
              async move {
    @@ src/index/usearch.rs: pub(crate) fn new(
                  // Incremental key for a usearch index
                  let usearch_key = Arc::new(AtomicU64::new(0));
      
    ++            // This semaphore decides how many tasks are queued for an usearch process. It is
    ++            // calculated as a number of threads multiply 2, to be sure that there is always a new
    ++            // task waiting in the queue.
     +            let semaphore = Arc::new(Semaphore::new(rayon::current_num_threads() * 2));
     +
                  while let Some(msg) = rx.recv().await {
```

src/index/usearch.rs
ewienik: Added a comment in 78f3446

src/db_index.rs
ewienik: Added a const in 78f3446. After some experience in testing/production there will be need to correctly setup channels bounds sizes, maybe adding some configuration for a back pressure.

swasik (approved):

PR #114, opened 2025-04-22
Implement token range aware initial scan

ewienik:
This is a part of #54.

The first full scan of embeddings has been done using schema `SELECT pk, embeddings FROM keyspace.table` and it is not optimal solution (it uses one shard as a coordinator for all nodes scan). This patch implements token range concurrent scans like `SELECT pk, embeddings FROM keyspace.table WHERE token(pk) >= ? AND token(pk) <= ?`.  Token ranges are calculated using a scylladb driver token ring functionality. The max scans concurrency is calculated based on number of shards in a cluster.

### DIFF
--- a/src/db_index.rs
+++ b/src/db_index.rs
@@ -12,9 +12,8 @@ use crate::PrimaryKey;
 use crate::TableName;
 use anyhow::Context;
 use anyhow::anyhow;
-use futures::Stream;
 use futures::StreamExt;
-use futures::TryStreamExt;
+use futures::stream::BoxStream;
 use itertools::Itertools;
 use scylla::client::session::Session;
 use scylla::cluster::metadata::ColumnType;
@@ -25,15 +24,20 @@ use scylla::deserialize::value::DeserializeValue;
 use scylla::errors::DeserializationError;
 use scylla::errors::TypeCheckError;
 use scylla::frame::response::result::ColumnSpec;
+use scylla::routing::Token;
 use scylla::statement::prepared::PreparedStatement;
 use scylla::value::CqlValue;
+use std::iter;
+use std::num::NonZeroUsize;
 use std::sync::Arc;
+use tokio::sync::Semaphore;
 use tokio::sync::mpsc;
 use tokio::sync::oneshot;
 use tracing::Instrument;
 use tracing::debug;
 use tracing::debug_span;
 use tracing::trace;
+use tracing::warn;
 
 type GetPrimaryKeyColumnsR = Vec<ColumnName>;
 
@@ -69,23 +73,14 @@ pub(crate) async fn new(
     let (tx_index, mut rx_index) = mpsc::channel(CHANNEL_SIZE);
     let (tx_embeddings, rx_embeddings) = mpsc::channel(CHANNEL_SIZE);
 
-    let mut items = statements.initial_scan().await?;
     tokio::spawn(
         async move {
             debug!("starting");
 
             while !rx_index.is_closed() {
                 tokio::select! {
-                    item = items.next() => {
-                        let Some(item) = item else {
-                            break;
-                        };
-                        let Ok(item) = item else {
-                            break;
-                        };
-                        if tx_embeddings.send(item).await.is_err() {
-                            break;
-                        }
+                    _ = statements.initial_scan(tx_embeddings.clone()) => {
+                        break;
                     }
                     Some(msg) = rx_index.recv() => {
                         tokio::spawn(process(Arc::clone(&statements), msg));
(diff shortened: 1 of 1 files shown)


src/db_index.rs
```diff
@@ -218,32 +215,115 @@ impl Statements {
...
         embeddings: &ColumnName,
     ) -> String {
         format!(
             "
-            SELECT {st_primary_key_select}, {embeddings}
+            SELECT {st_primary_key_list}, {embeddings}
             FROM {keyspace}.{table}
+            WHERE
+                token({st_partition_key_list}) >= ?
+                AND token({st_partition_key_list}) <= ?
```
swasik, line 231: Should both of them be closed intervals? Should not one of the end skip =?

src/db_index.rs
```diff
@@ -218,32 +215,115 @@ impl Statements {
...
+    fn nr_parallel_queries(&self) -> NonZeroUsize {
+        const SMUGE_FACTOR: NonZeroUsize = NonZeroUsize::new(3).unwrap();
+        self.nr_shards_in_cluster()
+            .checked_mul(SMUGE_FACTOR)
+            .unwrap()
+    }
+
+    fn fullscan_ranges(&self) -> impl Iterator<Item = (Token, Token)> {
+        const TOKEN_MAX: i64 = i64::MAX;
+        const TOKEN_MIN: i64 = -TOKEN_MAX;
```
swasik, line 281: Why not `i64::MIN`? Are we sure this is correct, as `::MIN = -::MAX - 1` (no need to represent 0 for negatives).

src/db_index.rs
swasik: Ok, I found later that you subtract 1 from the right end.

src/db_index.rs
ewienik: i64::MIN means infinity in ScyllaDB. See https://docs.rs/scylla/latest/scylla/routing/struct.Token.html

ewienik:
Changelog for ad1624c:
- update comments for initial_scan function

<details>
<summary>range diff</summary>

```
1:  5b613fa ! 1:  ad1624c Implement token range aware initial scan
    @@ src/db_index.rs: impl Statements {
          }
      
     -    async fn initial_scan(
    ++    /// The initial full scan of embeddings stored in a ScyllaDB table. It scans concurrently using
    ++    /// token ranges read from a rust driver. At first it prepares ranges, limits concurrent scans
    ++    /// using semaphore, and runs each scan in separate concurrent task using cloned mpsc channel
    ++    /// to send read embeddings into the pipeline.
     +    async fn initial_scan(&self, tx: mpsc::Sender<DbEmbeddings>) {
     +        let semaphore = Arc::new(Semaphore::new(self.nr_parallel_queries().get()));
     +
```

src/db_index.rs
```diff
@@ -214,32 +211,115 @@ impl Statements {
         self.primary_key_columns.clone()
     }
 
-    fn initial_scan_query(
+    fn range_scan_query(
```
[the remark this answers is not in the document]
ewienik: fixed in ad1624c

src/db_index.rs
swasik: Could you add comment about it here? It is definitely not intuitive.

swasik (dismissed):

ewienik:
Changelog for e94d089:
- added a comment for fullscan_ranges function

<details>
<summary>range diff</summary>

```
1:  ad1624c ! 1:  e94d089 Implement token range aware initial scan
    @@ src/db_index.rs: impl Statements {
     +            .unwrap()
     +    }
     +
    ++    /// Creates an iterator over all tokens ranges available in a cluster. A token ring is taken
    ++    /// from the rust driver as a list of tokens. A token range is build from a token pair
    ++    /// (tokens[idx], tokens[idx+1]-1) to be able to use WHERE token >= ? AND token <= ? in CQL
    ++    /// queries - the next token starts the new range. Additionally to the list of tokens taken
    ++    /// from the token ring, the first item is the lowest possible token, and the last item is the
    ++    /// highest possible token - for support the specific token range after the highest token to
    ++    /// the lowest token. The highest possible token value is not decremented, because it doesn't
    ++    /// start a new range.
     +    fn fullscan_ranges(&self) -> impl Iterator<Item = (Token, Token)> {
    -+        const TOKEN_MAX: i64 = i64::MAX;
    -+        const TOKEN_MIN: i64 = -TOKEN_MAX;
    ++        const TOKEN_MAX: i64 = i64::MAX; // the highest possible token value in the ScyllaDB
    ++        const TOKEN_MIN: i64 = -TOKEN_MAX; // the lowest possible token value in the ScyllaDB
     +
     +        let tokens = iter::once(Token::new(TOKEN_MIN))
     +            .chain(
```

src/db_index.rs
ewienik: fixed in e94d089

Issue #24, opened 2025-04-07
Use CDC for table to add/remove/modify data to index


ewienik:
Current version of vector-store has been built for a specific benchmark framework and depends on custom tables for storing status of processed items in a table. Vector-store should depend on CDC functionality to lookup for modifications to the table: adding or removing, or modifying rows in index.

The scope of this issue covers just **the regular CDC implementation** that exists in ScyllaDB for vnodes. The custom modifications of CDC for vector search service will be added in https://github.com/scylladb/vector-store/milestone/2.

ewienik:
Current list of PRs
- #117
- #118
- #119
- #120
- #121
- #122
- #123
- #124

PR #121, opened 2025-04-28
gh-24: implement a db index validation

ewienik:
Schema changes are concurrent processes without an atomic view from the
client/driver side. A process of retrieving an index metadata from the
vector-store could be faster than similar process in a driver itself.
The vector-store reads some schema metadata from system tables directly,
because they are not available from a rust driver, and it reads some
other schema metadata from the rust driver, so there must be an
agreement between data read directly from a db and a driver. This patch
checks if index metadata are correct and if there is an agreement of a
db schema in the rust driver.

---

### List of PRs for #24
- #117
- #118
- #119
- #120
- -> implement a db index validation
- #122
- #123
- #124

PR #124, opened 2025-04-28
gh-24: implement reading new embeddings from the CDC

ewienik:
This patch implements support for reading CDC in a concurrent task and
push outcome data into copy of the mpsc channel. The task is started
before initial full scan, to be able to catch all data written into the
table after the fullscan finish. There is a possibility that some data
will be duplicated (same in fullscan and in CDC) but merging will be
solved by using timestamps. Timestamps from the CDC functionality in the
scylla-cdc crate are compatible with UUIDv1 timestamp format, which is
hunderds of nanosecods since gregorian epoch.

Fixes: #24

---

### List of PRs for #24
- #117
- #118
- #119
- #120
- #121
- #122
- #123
- -> implement reading new embeddings from the CDC

### DIFF
--- a/src/db_index.rs
+++ b/src/db_index.rs
@@ -11,6 +11,7 @@ use crate::TableName;
 use anyhow::Context;
 use anyhow::anyhow;
 use anyhow::bail;
+use async_trait::async_trait;
 use futures::StreamExt;
 use futures::TryStreamExt;
 use futures::stream::BoxStream;
@@ -20,11 +21,17 @@ use scylla::routing::Token;
 use scylla::statement::prepared::PreparedStatement;
 use scylla::value::CqlValue;
 use scylla::value::Row;
+use scylla_cdc::consumer::CDCRow;
+use scylla_cdc::consumer::Consumer;
+use scylla_cdc::consumer::ConsumerFactory;
+use scylla_cdc::log_reader::CDCLogReaderBuilder;
 use std::iter;
 use std::num::NonZeroUsize;
 use std::sync::Arc;
 use std::time::Duration;
+use time::Date;
 use time::OffsetDateTime;
+use time::Time;
 use tokio::sync::Semaphore;
 use tokio::sync::mpsc;
 use tokio::sync::oneshot;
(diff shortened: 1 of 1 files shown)


src/db_index.rs
```diff
@@ -339,3 +374,131 @@ impl Statements {
...
+        let target_column = self.0.target_column.as_ref();
+        if !row.column_deletable(target_column) {
+            bail!("CDC error: target column {target_column} should be deletable");
+        }
+
+        let embedding = row
+            .take_value(target_column)
+            .map(|value| {
+                let CqlValue::Vector(value) = value else {
+                    bail!("CDC error: target column {target_column} should be VECTOR type");
```
swasik, line 404: What happens to those errors? Do we forward them to the client?

src/db_index.rs
```diff
@@ -339,3 +374,131 @@ impl Statements {
...
+        );
+
+        Ok(Self(Arc::new(CdcConsumerData {
+            primary_key_columns,
+            target_column: metadata.target_column.clone(),
+            tx,
+            gregorian_epoch,
+        })))
+    }
+}
```
swasik, line 504: Do we have some issue to remember about adding unit/integration tests for CDC?

src/db_index.rs
ewienik: Reading from CDC is a concurrent task and this error should be only visible in log file in debug level. There is no client which is able to receive such errors:

```
91     tokio::spawn(                                 
92         async move {                              
93             debug!("starting");                   
94                                                   
95             if let Err(err) = cdc_handler.await { 
96                 debug!("handler: {err}");         
97             }                                     
98                                                   
99             debug!("finished");                   
00         }                                         
01         .instrument(debug_span!("cdc", "{}", id)),
02     );                                            
```

src/db_index.rs
ewienik: `vector-store` service has integration tests with mocked `Db` and `DbIndex` actors, because there is no ScyllaDB available. The same is for unit tests - no possible to mockup (or run in memory) the ScyllaDB. There is a possibility to make unit tests for specific functions, independent on an external db.

src/db_index.rs
swasik: Right, this was stupid question. But nevertheless, I do not like the answer. I added #132 to add vector-service log to our monitoring stack so that we can be notified when something is broken.

src/db_index.rs
swasik: I remember the mockup challenges but nevertheless I think we should have some approved approach to testing. Let us discuss it offline next week.

swasik (approved):

PR #125, opened 2025-04-28
Add opensearch index: create backend for Opensearch

smoczy123:
This is a part of #5.

Adds the ability to connect to Opensearch database and create an empty vector index there.

PR #130, opened 2025-04-29
Add CI for opensearch feature

QuerthDP:
According to [Paweł's comment](https://github.com/scylladb/vector-store/pull/125#pullrequestreview-2802631857) this patch introduces github workflow adjustments to check and test the code written under the opensearch feature.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -42,6 +42,9 @@ jobs:
     - name: Clippy check
       run: cargo clippy --all-targets
 
+    - name: Clippy check with opensearch feature
+      run: cargo clippy --all-targets --features opensearch
+
   tests:
     runs-on: ubuntu-latest
     steps:
@@ -55,4 +58,7 @@ jobs:
 
     - name: Run tests
       run: cargo test --verbose
+
+    - name: Run tests with opensearch feature
+      run: cargo test --verbose --features opensearch
 
--- a/src/index/opensearch.rs
+++ b/src/index/opensearch.rs
@@ -3,30 +3,41 @@
  * SPDX-License-Identifier: LicenseRef-ScyllaDB-Source-Available-1.0
  */
 
+// TODO: Please remove if necessary implementation is provided.
+#![allow(dead_code)]
+
 use crate::Connectivity;
 use crate::Dimensions;
-use crate::Distance;
-use crate::Embeddings;
 use crate::ExpansionAdd;
 use crate::ExpansionSearch;
 use crate::IndexId;
-use crate::IndexItemsCount;
-use crate::Key;
-use crate::Limit;
-use crate::db::Db;
+use crate::PrimaryKey;
 use crate::index::actor::Index;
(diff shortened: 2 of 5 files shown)


QuerthDP:
My concerns here are that we do not have any automation introduced to run OpenSearch instance, however it is required for testing the opensearch feature (opensearch client must connect to the running instance, so far being setup manually via docker / local setup / `test/pylib/opensearch_cluster.py`).

ewienik:
> My concerns here are that we do not have any automation introduced to run OpenSearch instance, however it is required for testing the opensearch feature (opensearch client must connect to the running instance, so far being setup manually via docker / local setup / `test/pylib/opensearch_cluster.py`).

I think starting opensearch instance should be avoided. This service could be mocked by simple http server with specific endpoints used by an opensearch index. There could be some predefined/mocked responses for HTTP request, but it is also possible to use an usearch index as an mocked service, and eventually there will be mocked opensearch endpoints + usearch index behind.

.github/workflows/rust.yml
```diff
@@ -42,6 +42,9 @@ jobs:
     - name: Clippy check
       run: cargo clippy --all-targets
 
+    - name: Clippy check with all features
+      run: cargo clippy --all-targets --all-features
```
ewienik, line 46:
```suggestion
    - name: Clippy check with opensearch feature
      run: cargo clippy --all-targets --features opensearch
```
The `default` and the `opensearch` features are exclusive, so it is better to run steps for them separately, and name them accordingly. There should be also additional steps for `cargo fmt` and `cargo test`

ewienik (changes requested):

.github/workflows/rust.yml
ewienik: The commit and PR description should be also changed as this patch should introduce only CI tests for `opensearch`

ewienik: The commit and PR description should be also changed as this patch should introduce only CI tests for `opensearch`

.github/workflows/rust.yml
```diff
@@ -42,6 +42,9 @@ jobs:
     - name: Clippy check
       run: cargo clippy --all-targets
 
+    - name: Clippy check with opensearch feature
+      run: cargo clippy --all-targets --features opensearch
```
ewienik, line 46: There is a need for a cargo fmt check with an opensearch feature

ewienik (changes requested):

.github/workflows/rust.yml
QuerthDP: AFAIK and as I checked manually, the `cargo fmt --all -- --check` does check the formatting behind `[cfg(feature = "...")]`.

ewienik (changes requested):
Need to resolve failing tests

QuerthDP:
Changelog:

- introduced minor changes to the code including Rust attributes allowing the warnings to make the CI pass

ewienik (approved):

PR #144, opened 2025-05-28
VECTOR-74 Add Opensearch index: implement Index::Count for Opensearch

QuerthDP:
Add an implementation of Index::Count in the Opensearch backend.


smoczy123 (dismissed):
Please adress the comment before merging

smoczy123 (approved):

PR #148, opened 2025-06-03
VECTOR-74: Remove opensearch feature flag

QuerthDP:
Removed `opensearch` feature.
Now both the code for Usearch and OpenSearch are compiled and available to use.

This change is required to build docker image able to use either provider. 
One of the above is now chosen setting `SCYLLA_INDEXING_PROVIDER` environmental variable to `opensearch` or `usearch` respectively. By default the Usearch is being used.

Fixes: VECTOR-74 (Add opensearch index)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -27,6 +27,7 @@ use std::hash::Hasher;
 use std::net::SocketAddr;
 use std::num::NonZeroUsize;
 use std::str::FromStr;
+use std::sync::Once;
 use time::OffsetDateTime;
 use tokio::signal;
 use tokio::sync::mpsc::Sender;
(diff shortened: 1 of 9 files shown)


src/lib.rs
```diff
@@ -450,14 +456,11 @@ pub async fn new_db(uri: ScyllaDbUri) -> anyhow::Result<Sender<Db>> {
...
-}
-
-#[cfg(feature = "opensearch")]
-pub fn new_index_factory(addr: String) -> anyhow::Result<impl IndexFactory> {
-    index::opensearch::new_opensearch(&addr)
+pub fn new_index_factory(addr: Option<String>) -> anyhow::Result<Box<dyn IndexFactory + Send>> {
+    if let Some(addr) = addr {
+        return Ok(Box::new(index::opensearch::new_opensearch(&addr)?));
+    }
+    Ok(Box::new(index::usearch::new_usearch()?))
```
ewienik, line 463: `addr` parameter shouldn't have meaning of 'choose usearch of opensearch'. I'd rather use two functions: `new_index_factory_usearch` and `new_index_factory_opensearch`.

src/main.rs
```diff
@@ -37,15 +37,28 @@ async fn main() -> anyhow::Result<()> {
...
-    let index_factory = {
-        let addr = dotenvy::var("OPENSEARCH_ADDRESS").unwrap_or("http://localhost".to_string());
-        let port = dotenvy::var("OPENSEARCH_PORT").unwrap_or("9200".to_string());
-        let addr = format!("{addr}:{port}");
-        vector_store::new_index_factory(addr)?
-    };
-    #[cfg(not(feature = "opensearch"))]
-    let index_factory = vector_store::new_index_factory()?;
+    let indexing_provider =
+        dotenvy::var("SCYLLA_INDEXING_PROVIDER").unwrap_or("usearch".to_string());
```
ewienik, line 41: Do we need to add additional env variable? Maybe enough is to check if `OPENSEARCH_ADDRESS` and `OPENSEARCH_PORT` exist.

ewienik (changes requested):

ewienik (approved):

PR #153, opened 2025-06-23
Update update_jira_status_in_review.yml - Detect the jira ticket from the PR body

dani-tweig:
- Update the logic so that the action will update one or several Jira tickets to 'in review' status. 
- The Jira ticket number can be either in the PR title or the PR body.


dani-tweig:
I know that it does nothing of there is no jira ticket. I don't know why it failed.

PR #155, opened 2025-06-23
Update update_jira_status_in_review.yml- Detect the jira ticket from the PR body

dani-tweig:
- Update the logic so that the action will update one or several Jira tickets to 'in review' status. 
- The Jira ticket number can be either in the PR title or the PR body.
- Exit gracefully if no Jira key was found without doing anything.


dani-tweig:
@swasik can you review and merge?

PR #158, opened 2025-06-26
Add Prometheus metrics infrastructure with index latency tracking

amnonh:
This series introduces basic Prometheus observability to the vector search service. The following changes are included:

Prometheus dependencies were added to support metrics collection and export.

A new metrics module defines a Metrics object containing a latency histogram (request_latency_seconds) labeled by keyspace and index name.

The HTTP server and route handlers were updated to:
Wire the Metrics object through the application state.
Expose a /metrics endpoint for Prometheus scraping.
Record latency for vector search requests in the post_index_ann handler.

This lays the groundwork for more comprehensive metric tracking per index in future changes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,8 @@ utoipa = { version = "5.3.1", features = ["axum_extras"] }
 utoipa-axum = "0.2.0"
 utoipa-swagger-ui = { version = "9.0.0", features = ["axum"] }
 uuid = "1.16.0"
+prometheus = "0.14"
+
 
 [patch.crates-io]
 scylla = { git = "https://github.com/ewienik/scylla-rust-driver.git", rev = "5fef6ae" }
--- a/src/httproutes.rs
+++ b/src/httproutes.rs
@@ -2,7 +2,8 @@
  * Copyright 2025-present ScyllaDB
  * SPDX-License-Identifier: LicenseRef-ScyllaDB-Source-Available-1.0
  */
-
+use crate::metrics::Metrics;
+use std::sync::Arc;
 use crate::ColumnName;
 use crate::Distance;
 use crate::Embedding;
@@ -39,7 +40,7 @@ use utoipa::OpenApi;
 use utoipa_axum::router::OpenApiRouter;
 use utoipa_axum::routes;
 use utoipa_swagger_ui::SwaggerUi;
-
+use axum::routing::get;
 #[derive(OpenApi)]
 #[openapi(
     tags(
@@ -48,8 +49,10 @@ use utoipa_swagger_ui::SwaggerUi;
 )]
 // TODO: modify HTTP API after design
 struct ApiDoc;
-
-pub(crate) fn new(engine: Sender<Engine>) -> Router {
+#[derive(Clone)]
+struct RoutesInnerState { engine: Sender<Engine>, metrics: Arc<Metrics>}
+pub(crate) fn new(engine: Sender<Engine>, metrics: Arc<Metrics>) -> Router {
+    let state = RoutesInnerState { engine, metrics: metrics.clone() };
     let (router, api) = OpenApiRouter::with_openapi(ApiDoc::openapi())
         .merge(
             OpenApiRouter::new()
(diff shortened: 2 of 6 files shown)


Cargo.toml
ewienik, line 1: Commit description:
>  Cargo: Add prometheus and protobuf support
>
> This patch adds the dependencies for Prometheus and Protobuf support to
> allow metrics reporting.

I think `Protobuf` should be removed from description - there is no such dependency in the patch

src/metrics.rs
```diff
@@ -0,0 +1,55 @@
...
+use std::sync::Arc;
+
+#[derive(Clone)]
+pub struct Metrics {
+    pub registry: Registry,
+    pub latency: HistogramVec,
+}
+
+impl Metrics {
+    pub fn new() -> Arc<Self> {
```
ewienik, line 12: `new` should return only `Self`. If there is a need for and `Arc` it could be chained `Arc::new(Metrics::new)`

src/metrics.rs
```diff
@@ -0,0 +1,55 @@
...
+        ).unwrap();
+
+        registry.register(Box::new(latency.clone())).unwrap();
+
+        Arc::new(Self {
+            registry,
+            latency,
+        })
+    }
+     pub async fn handler(self: Arc<Self>) -> impl IntoResponse {
```
ewienik, line 27: Handler should go into `http_routes.rs` for consistence with other handlers (probably as `get_metrics`)

If you want to return http response from different sources you can use `Response` as a result instead of `impl IntoResponse`, then there won't be need for use `.to_vec()` in one of the responses.

src/metrics.rs
```diff
@@ -0,0 +1,55 @@
+use prometheus::{Encoder, TextEncoder, Registry, HistogramVec};
+use axum::{response::IntoResponse, http::{StatusCode, header, HeaderMap}};
```
ewienik, line 2: To be consistent with other sources `use` should use one item per line.

src/httproutes.rs
ewienik, line 1: You can sync with master - there are added test for openapi documentation.

src/httproutes.rs
```diff
@@ -48,20 +49,31 @@ use utoipa_swagger_ui::SwaggerUi;
...
         )
         .split_for_parts();
 
+    let metrics_route = Router::new().route(
+        "/metrics",
+        get({
+            let metrics = metrics.clone();
+            move || metrics.clone().handler()
+        }),
+    );
```
ewienik, line 74: To be consistent with other routes you can prepare a new function handler with an openapi documentation instead of inplace lambda

src/httproutes.rs
```diff
@@ -134,16 +146,30 @@ pub struct PostIndexAnnResponse {
...
+    // Start timing
+    let timer = state.metrics
+            .latency
+            .with_label_values(&[
+                keyspace.as_ref().as_str(),
+                index.as_ref().as_str()
+            ])
+            .start_timer();
+
+    let Some((index, db_index)) = state.engine.get_index(IndexId::new(&keyspace, &index)).await else {
```
ewienik, line 164: When the routine finishes early in this `let else` what will happen to the timer?

src/lib.rs
```diff
@@ -40,6 +41,8 @@ use utoipa::openapi::SchemaFormat;
 use utoipa::openapi::schema::Type;
 use uuid::Uuid;
 
+use crate::metrics::Metrics;
```
ewienik, line 44: let's connect this `use` with the block of all `uses`

src/lib.rs
```diff
@@ -11,7 +11,7 @@ mod httpserver;
 mod index;
 mod monitor_indexes;
 mod monitor_items;
-
+mod metrics;
```
ewienik, line 14: Let's leave the space between `mods` and `uses`

ewienik (changes requested):

amnonh:
$ git log --oneline master..HEAD
62f448a (HEAD -> prometheus_metrics) Connect the prometheus interface and metrics
bc2c812 Add metrics.rs to hold vector metrics
3884906 Cargo: Add prometheus and protobuf support
48ed9e9 (origin/master, origin/HEAD) Repeat adding index when there is an error
3880a83 Update scylla rust driver to 1.2.0


Only the last 3 commits are mine

ewienik (approved):

PR #162, opened 2025-07-01
REST API refactoring/redesign

knowack1:
Fixes: VECTOR-32

### DIFF
--- a/api/openapi.json
+++ b/api/openapi.json
@@ -2,7 +2,7 @@
   "openapi": "3.1.0",
   "info": {
     "title": "ScyllaDB Vector Store API",
-    "description": "REST API for ScyllaDB Vector Store - provides vector search and index management",
+    "description": "REST API for ScyllaDB Vector Store nodes. Provides capabilities for executing vector search queries, managing indexes, and checking status of Vector Store nodes.",
     "license": {
       "name": "LicenseRef-ScyllaDB-Source-Available-1.0"
     },
@@ -11,6 +11,9 @@
   "paths": {
     "/api/v1/indexes": {
       "get": {
+        "tags": [
+          "scylla-vector-store-index"
+        ],
         "description": "Get list of current indexes",
         "operationId": "get_indexes",
         "responses": {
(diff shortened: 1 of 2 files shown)


tests/integration/usearch.rs
```diff
@@ -122,13 +122,11 @@ async fn simple_create_search_delete_index() {
...
-            task::yield_now().await;
-        }
-    })
-    .await
-    .unwrap();
+    wait_for(
+        || async { client.count(&index).await == Some(3) },
+        "Waiting for 3 vectors to be indexed",
+    )
+    .await;
```
Lorak-mmk, line 127: Curious: why are you using timeouts for specific operations, instead of whole test? You could use `ntest` crate for that, as we do in rust driver. Example usage: https://github.com/scylladb/scylla-rust-driver/blob/7d1f45380da0d1bf5d7d52c07af63cb2caea571b/scylla/tests/integration/session/cluster_reachability.rs#L11-L14

Lorak-mmk: > It seems that ntest does not point to exact place in test case where timeout occurred

Right, this is a very good point. I wonder if its even solvable, or it is some inherent limitation that ntest or other macro could not fix.

> One must remember to add #timeout macro while using wait_for to not hand forever in case of failure.
> I haven't found option (with ntest) to add global timeout for all test cases - without explicitly specifying by macro.

There are 2 nice solutions for that.

First is using nextest as your test runner. It has support for identifying slow tests and setting timeout globally: https://nexte.st/docs/features/slow-tests/
It also has other advantages, so you may want to consider it.

The other solutions is to use your own macro: add new crate to workspace, it in write a macro, lets call it scylla::test, that has an optional timout parameter, so you would use it like this:
```rust
#[scylla::test]
fn some_test_function() {

}
```

```rust
#[scylla::test(30000)]
fn some_test_function() {

}
```

For case without explicit timeout, you can then take default from e.g. env variable.
Then you need to remember to use that instead of `tokio::test`, but this is a matter of putting a right grep in ci.

PR #167, opened 2025-07-03
Add bad request to search response

knowack1:
Add 400 response when embedding in search request has wrong size

This commit adds a 400 (Bad Request) response to the /ann search endpoint when the embedding vector size does not match the index dimension.

Reference: VECTOR-32

### DIFF
--- a/api/openapi.json
+++ b/api/openapi.json
@@ -84,6 +84,9 @@
               }
             }
           },
+          "400": {
+            "description": "Bad request. Possible causes: invalid embedding size, malformed input, or missing required fields."
+          },
           "404": {
             "description": "Index not found. Possible causes: index does not exist, or is not built yet."
           },
--- a/crates/vector-store/src/httproutes.rs
+++ b/crates/vector-store/src/httproutes.rs
@@ -14,6 +14,7 @@ use crate::db_index::DbIndexExt;
 use crate::engine::Engine;
 use crate::engine::EngineExt;
 use crate::index::IndexExt;
+use crate::index::validator;
 use crate::info::Info;
 use crate::metrics::Metrics;
 use anyhow::bail;
@@ -282,6 +283,7 @@ The similarity metric is determined at index creation and cannot be changed per 
             description = "Successful ANN search. Returns a list of primary keys and their corresponding distances for the most similar vectors found.",
             body = PostIndexAnnResponse
         ),
+        (status = 400, description = "Bad request. Possible causes: invalid embedding size, malformed input, or missing required fields."),
         (status = 404, description = "Index not found. Possible causes: index does not exist, or is not built yet."),
         (status = 500, description = "Ann search error. Possible causes: internal error, or search engine issues."),
     )
@@ -312,12 +314,14 @@ async fn post_index_ann(
     timer.observe_duration();
 
     match search_result {
-        Err(err) => {
-            let msg = format!("index.ann request error: {err}");
-            debug!("post_index_ann: {msg}");
-            (StatusCode::INTERNAL_SERVER_ERROR, msg).into_response()
-        }
-
+        Err(err) => match err.downcast_ref::<validator::Error>() {
+            Some(err) => (StatusCode::BAD_REQUEST, err.to_string()).into_response(),
+            None => {
+                let msg = format!("index.ann request error: {err}");
+                debug!("post_index_ann: {msg}");
+                (StatusCode::INTERNAL_SERVER_ERROR, msg).into_response()
+            }
+        },
         Ok((primary_keys, distances)) => {
             if primary_keys.len() != distances.len() {
                 let msg = format!(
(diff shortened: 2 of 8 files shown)


src/httproutes.rs
```diff
@@ -14,6 +14,7 @@ use crate::db_index::DbIndexExt;
 use crate::engine::Engine;
 use crate::engine::EngineExt;
 use crate::index::IndexExt;
+use crate::index::actor::AnnError;
```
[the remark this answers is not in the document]
knowack1: This validation was already done by the business logic itself – I didn't change that. Nevertheless, I think it's a somewhat good place. I would expect that the business logic function IndexExt::ann will fail with a descriptive error in this case.

However, I can move this check up to the handler.
Should I then, remove validation from the business logic to not duplicate it? 
I think yes. Even if IndexExt::ann is somehow called with an incorrect embedding, it will eventually fail on search engine level.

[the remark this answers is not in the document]
ewienik: You're right, now there is validation on the index side. The best way could be some newtype which will be be invariant and will store correct dimensions. IMHO it is impossible, because there could be a few indexes with different dimensions, so some validation must be done in index.

src/index/actor.rs
```diff
@@ -7,10 +7,40 @@ use crate::Distance;
...
+                    f,
+                    "Wrong embedding dimension: expected {expected}, got {actual}"
+                )
+            }
+            AnnError::OtherError(err) => write!(f, "Other error: {err}"),
+        }
+    }
+}
+
+pub(crate) type AnnR = anyhow::Result<(Vec<PrimaryKey>, Vec<Distance>), AnnError>;
```
ewienik, line 43:
```suggestion
pub(crate) type AnnR = anyhow::Result<(Vec<PrimaryKey>, Vec<Distance>)>;
```

I would propose to use [anyhow::bail!](https://docs.rs/anyhow/latest/anyhow/macro.bail.html) to send custom error as anyhow::Error

src/index/opensearch.rs
```diff
@@ -14,7 +14,9 @@ use crate::IndexId;
 use crate::Limit;
 use crate::PrimaryKey;
 use crate::SpaceType;
+use crate::index::actor::AnnError;
 use crate::index::actor::Index;
+use crate::index::validator::validate_embedding_dimensions;
```
ewienik, line 19: In Rust convention is to import modules, not functions and call functions using module name: `validator::validate_embedding_dimensions`. Same for importing in `index::usearch`.
```suggestion
use crate::index::validator;
```

src/index/validator.rs
```diff
@@ -0,0 +1,88 @@
+use crate::Dimensions;
+use crate::Embedding;
+use crate::index::actor::AnnError;
+
+pub fn validate_embedding_dimensions(
```
ewienik, line 5: As module is called `validator`
```suggestion
pub fn embedding_dimensions(
```

src/index/actor.rs
```diff
@@ -7,10 +7,40 @@ use crate::Distance;
 use crate::Embedding;
 use crate::Limit;
 use crate::PrimaryKey;
+use std::fmt;
 use tokio::sync::mpsc;
 use tokio::sync::oneshot;
 
-pub(crate) type AnnR = anyhow::Result<(Vec<PrimaryKey>, Vec<Distance>)>;
+#[derive(Debug)]
+pub enum AnnError {
```
ewienik, line 15: Semantically, it is a validator error, not an actor error. It could be moved into a `index::validator` module.

src/index/actor.rs
```diff
@@ -7,10 +7,40 @@ use crate::Distance;
...
+    WrongEmbeddingDimension { expected: usize, actual: usize },
+    OtherError(anyhow::Error),
+}
+
+impl<E> From<E> for AnnError
```
ewienik, line 20: You can use crate [thiserror](https://docs.rs/thiserror/latest/thiserror/) for definition of the error

src/index/mod.rs
```diff
@@ -3,6 +3,8 @@
  * SPDX-License-Identifier: LicenseRef-ScyllaDB-Source-Available-1.0
  */
 
+mod validator;
+
```
ewienik, line 7: nit: remove empty line

tests/integration/main.rs
```diff
@@ -30,3 +33,17 @@ fn enable_tracing() {
             .init();
     });
 }
+
+async fn wait_for<F, Fut>(mut condition: F, msg: &str)
```
ewienik, line 37: Introducing `wait_for` could be done in a separate commit.

ewienik (approved):

src/index/actor.rs
ewienik: Returning custom errors is mostly done on libraries (crates). On the application side using anyhow::Error is a common path. We could think how to refactor error maintenance as we need a way to log errors, send back http errors. But for this we need to get some experience what is really needed in this application.

PR #171, opened 2025-07-04
vsc: read from vsc log

QuerthDP:
Set `CDCLogReader` to read from `_scylla_vsc_log` tables.

Fixes: VECTOR-91

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2020,8 +2020,7 @@ dependencies = [
 [[package]]
 name = "scylla-cdc"
 version = "0.4.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0ef47312c2d9ab846dd5117773dd01df4122de8b36f54945970c4f25614f4f80"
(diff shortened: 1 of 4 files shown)


QuerthDP:
> Well I believe we have to keep the `scylla-cdc` dependency based on a github branch for now as the official release is not coming soon I suppose.

As we talked with @wprzytula, we probably will release the new version with support for reading the VSC log.
Then I believe this PR would benefit from splitting the commits (update dependency and use the new feature).

QuerthDP:
Changelog:

- split commits into update dependency and read from vsc log
- rebase on main

src/db.rs
```diff
@@ -479,10 +479,10 @@ impl Statements {
             return false;
         }
 
-        // check a cdc log table
+        // check a vsc log table
         if !keyspace
             .tables
-            .contains_key(&format!("{}_scylla_cdc_log", metadata.table_name))
+            .contains_key(&format!("{}_scylla_vsc_log", metadata.table_name))
```
ewienik, line 485: I'd rather introduce special feature to switch from `cdc_log` into `vsc_log` as we need to support normal cdc for a moment.

ewienik (changes requested):

src/db.rs
ewienik: ok, let's keem this PR opened till change in the scylladb

PR #180, opened 2025-07-10
fix: remove index metadata

QuerthDP:
The index metadata in `MonitorIndexes` was never removed, which led to storing unnecessary metadata.
Furthermore it caused logging multiple communicates of removing the indexes that were already removed.

Remove index metadata from `MonitorIndexes` after deleting the index.

Fixes: VECTOR-138

### DIFF
--- a/crates/vector-store/src/monitor_indexes.rs
+++ b/crates/vector-store/src/monitor_indexes.rs
@@ -63,7 +63,7 @@ pub(crate) async fn new(
                         node_state.send_event(
                             Event::IndexesDiscovered(new_indexes.clone()),
                         ).await;
-                        del_indexes(&engine, indexes.difference(&new_indexes)).await;
+                        del_indexes(&engine, indexes.extract_if(|idx| !new_indexes.contains(idx))).await;
                         let AddIndexesR {added, has_failures} = add_indexes(
                             &engine,
                             new_indexes.into_iter().filter(|idx| !indexes.contains(idx))
@@ -197,7 +197,7 @@ async fn add_indexes(
     }
 }
 
-async fn del_indexes(engine: &Sender<Engine>, idxs: impl Iterator<Item = &IndexMetadata>) {
+async fn del_indexes(engine: &Sender<Engine>, idxs: impl Iterator<Item = IndexMetadata>) {
     for idx in idxs {
         engine.del_index(idx.id()).await;
     }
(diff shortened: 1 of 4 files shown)


crates/vector-store/src/monitor_indexes.rs
```diff
@@ -53,7 +53,7 @@ pub(crate) async fn new(
                             schema_version.reset();
                             continue;
                         };
-                        del_indexes(&engine, indexes.difference(&new_indexes)).await;
+                        del_indexes(&engine, indexes.extract_if(|idx| !new_indexes.contains(idx))).await;
```
knowack1, line 56: Is it possible to create reproduction test?

knowack1: It looks that unit test must assert what is passed to engine.del_index(...)

QuerthDP: Done, please take a look.

crates/vector-store/src/monitor_indexes.rs
```diff
@@ -292,4 +292,237 @@ mod tests {
...
+        *state.error_tx.lock().unwrap() = Some(error_tx);
+
+        let tx_db = new_db(state.clone()).await.unwrap();
+        let tx_eng = new_engine(state.clone()).await.unwrap();
+
+        // Start the monitor
+        let _monitor = new(tx_db.clone(), tx_eng.clone()).await.unwrap();
+
+        // Wait a bit for initial processing
+        tokio::time::sleep(std::time::Duration::from_millis(100)).await;
```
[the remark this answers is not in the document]
Lorak-mmk: And at some point you will notice your tests taking absurd amounts of time, 95% of which spent in sleeps.

crates/vector-store/src/monitor_indexes.rs
```diff
@@ -292,4 +292,219 @@ mod tests {
...
+                    })
+                    .collect()
+            }
+        }
+
+        // Engine mock
+        async fn new_engine(state: TestState) -> anyhow::Result<mpsc::Sender<Engine>> {
+            let (tx_eng, mut rx_eng) = mpsc::channel(10);
+
+            tokio::spawn(async move {
```
knowack1, line 366: We should mock EngineExt and DbExt. Then treads will be avoidable in unittests.

crates/vector-store/src/monitor_indexes.rs
```diff
@@ -292,4 +292,219 @@ mod tests {
...
+            panic!("Unexpected error during setup: {error}");
+        }
+
+        // Add index1 to the database
+        let index1 = sample_db_index("index1");
+        state.set_db_indexes(vec![index1]);
+        state.notify.notified().await;
+
+        // Check for errors
+        if let Ok(error) = error_rx.try_recv() {
```
knowack1, line 478: What we would like to verify here?

crates/vector-store/src/monitor_indexes.rs
```diff
@@ -292,4 +292,219 @@ mod tests {
...
+            panic!("Unexpected error: {error}");
+        }
```
knowack1, line 480: assert that index was added to the engine

crates/vector-store/src/monitor_indexes.rs
```diff
@@ -292,4 +292,219 @@ mod tests {
...
+
+        // Remove the index from the database
+        state.set_db_indexes(vec![]);
+        state.notify.notified().await;
```
knowack1, line 484: assert the index has been removed from the engine

crates/vector-store/src/monitor_indexes.rs
```diff
@@ -292,4 +292,219 @@ mod tests {
...
+
+        // Check for errors
+        if let Ok(error) = error_rx.try_recv() {
```
knowack1, line 487: assert that index was removed

crates/vector-store/src/monitor_indexes.rs
```diff
@@ -292,4 +292,219 @@ mod tests {
...
+            panic!("Unexpected error: {error}");
+        }
```
knowack1, line 508: This test tests exactly the same as part for index 1. What is the difference?
```
 // Add index2 to the database
        let index2 = sample_db_index("index2");
        state.set_db_indexes(vec![index2]);
        state.notify.notified().await;

        // Check for errors
        if let Ok(error) = error_rx.try_recv() {
            panic!("Unexpected error: {error}");
        }

        // Remove the index from the database
        state.set_db_indexes(vec![]);
        state.notify.notified().await;

        // Check for errors
        if let Ok(error) = error_rx.try_recv() {
            panic!("Unexpected error: {error}");
        }
```

crates/vector-store/src/monitor_indexes.rs
QuerthDP: The whole issue is about that the `engine.del_index` function is called multiple times, because the monitor indexes had not been removing the metadata of already removed indexes.
The tests were written having that in mind, so I wanted to check if the calls are not excessive, not that they happen at all.

It is a good idea to have the tests enhanced to check both the cases.

crates/vector-store/src/monitor_indexes.rs
QuerthDP: The point of this PR is fixing a bug that would cause calling `engine::del_index` excessive number of times (on already removed indexes). There would appear multiple logs about removing the index, which was already removed.

This bug was reproduced by adding and removing the indexes more than once, that's why the action is repeated here.
I used "index2" here, just to check if the bug happens also when adding and removing different indexes (it did).

knowack1: I would prefer to create more explicit/specific test case for the issue. Maybe something like (this is pseudo code the express the idea):
```
test index_metadata_are_removed_once {
//Arrage: Add two indexes to the db and wait for tick.
db.add_indexes([index1, index2]);
wait_for( engine.indexes().len() == 2)

//Act: Remove indexes one by one. Wait for the tick after each removal
db.remove_index(index2]);
wait_for(engine.indexes().len() == 1)
db.remove_index(index1]);
wait_for(engine.indexes().len() == 0)

//Assert: Ensure that call to del_indexes was called only once for specific index
assert(engine.del_indexes.calls.with(index") == 1)
assert(engine.del_indexes.calls.with(index2) == 1)
}
```
Does it make sense?

crates/vector-store/src/monitor_indexes.rs
ewienik: The design is to use `Sender<Msg>` as a handle for communication with actors, it simplifies actors communication and implementation. In other case we should go with traits - but we need to consider starting/stopping/async dropping actors. And by design each actor is a `task` (async thread), so we need to think about the system as an interactions between independent actors/tasks.

crates/vector-store/src/monitor_indexes.rs
```diff
@@ -292,4 +292,219 @@ mod tests {
...
+
+        // Check for any errors that occurred during setup
+        if let Ok(error) = error_rx.try_recv() {
+            panic!("Unexpected error during setup: {error}");
+        }
+
+        // Add index1 to the database
+        let index1 = sample_db_index("index1");
+        state.set_db_indexes(vec![index1]);
+        state.notify.notified().await;
```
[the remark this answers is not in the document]
ewienik: We could use timeout for whole test function using `ntest` crate

QuerthDP:
Changelog:

- add ntest dependency and use it's timeout
- derive Clone trait for DbCustomIndex
- assert insertion and deletion in engine indexes
- check for number of calls to `del_index`
- adjust the test to be more specific

QuerthDP:
Changelog:

- rebased on master

crates/vector-store/src/monitor_indexes.rs
```diff
@@ -302,4 +302,215 @@ mod tests {
...
+            "Both indexes should be present"
+        );
+
+        indexes.pop(); // Remove index1 from the list
+        state.set_db_indexes(&indexes).await;
+
+        let engine_indexes = state.engine_indexes.lock().unwrap().clone();
+        assert!(
+            !engine_indexes.contains(&index1_id) && !engine_indexes.contains(&index2_id),
+            "Both indexes should be present"
```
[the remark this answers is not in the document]
QuerthDP: Oh good catch - it's not. Same for the message above. Copied by mistake.

crates/vector-store/src/lib.rs
```diff
@@ -404,7 +404,7 @@ impl IndexMetadata {
     }
 }
 
-#[derive(Debug)]
+#[derive(Clone, Debug)]
```
knowack1, line 407: We added extra behavior (Clone) to the production code that is needed only for testing.
I would prefer to avoid this.

QuerthDP: As I talked with @ewienik we decided it's reasonable to derive this trait in this case.
I couldn't find any simpler solution for this than cloning.
I'll try to refactor it slightly. I believe it might be possible to change the set_db_indexes to add/remove db indexes and it should be ok without cloning.

QuerthDP:
Changelog:

- removed derive Clone
- fixed assertion messages

PR #185, opened 2025-07-11
Implement status endpoint

smoczy123:
This PR implements a status endpoint. For the feature to work properly, the initialization of the HTTP server had to be moved before the connection to the database.

### DIFF
--- a/crates/vector-store/src/engine.rs
+++ b/crates/vector-store/src/engine.rs
@@ -9,6 +9,7 @@ use crate::db::Db;
 use crate::db::DbExt;
 use crate::db_index::DbIndex;
 use crate::factory::IndexFactory;
+use crate::httpserver::ServerMsg;
 use crate::index::Index;
 use crate::monitor_indexes;
 use crate::monitor_items;
@@ -26,7 +27,7 @@ type GetIndexIdsR = Vec<IndexId>;
 type AddIndexR = anyhow::Result<()>;
 type GetIndexR = Option<(mpsc::Sender<Index>, mpsc::Sender<DbIndex>)>;
 
-pub(crate) enum Engine {
+pub enum Engine {
     GetIndexIds {
         tx: oneshot::Sender<GetIndexIdsR>,
     },
@@ -97,10 +98,11 @@ type IndexesT = HashMap<
 pub(crate) async fn new(
     db: mpsc::Sender<Db>,
     index_factory: Box<dyn IndexFactory + Send + Sync>,
+    server_actor: mpsc::Sender<ServerMsg>,
 ) -> anyhow::Result<mpsc::Sender<Engine>> {
     let (tx, mut rx) = mpsc::channel(10);
 
-    let monitor_actor = monitor_indexes::new(db.clone(), tx.clone()).await?;
+    let monitor_actor = monitor_indexes::new(db.clone(), tx.clone(), server_actor).await?;
 
     tokio::spawn(
         async move {
--- a/crates/vector-store/src/httproutes.rs
+++ b/crates/vector-store/src/httproutes.rs
@@ -12,8 +12,8 @@ use crate::KeyspaceName;
 use crate::Limit;
 use crate::Progress;
 use crate::db_index::DbIndexExt;
-use crate::engine::Engine;
 use crate::engine::EngineExt;
+use crate::httpserver::WrapEngine;
 use crate::index::IndexExt;
 use crate::index::validator;
 use crate::info::Info;
@@ -45,7 +45,7 @@ use time::Date;
 use time::OffsetDateTime;
 use time::Time;
 use time::format_description::well_known::Iso8601;
-use tokio::sync::mpsc::Sender;
+use tokio::sync::Mutex;
 use tower_http::trace::TraceLayer;
 use tracing::debug;
 use utoipa::OpenApi;
@@ -86,14 +86,20 @@ struct ApiDoc;
 
 #[derive(Clone)]
 struct RoutesInnerState {
-    engine: Sender<Engine>,
+    engine: WrapEngine,
     metrics: Arc<Metrics>,
+    node_state: Arc<Mutex<Status>>,
 }
 
-pub(crate) fn new(engine: Sender<Engine>, metrics: Arc<Metrics>) -> Router {
+pub(crate) fn new(
+    engine: WrapEngine,
+    metrics: Arc<Metrics>,
+    node_state: Arc<Mutex<Status>>,
+) -> Router {
     let state = RoutesInnerState {
         engine,
         metrics: metrics.clone(),
+        node_state,
     };
     let (router, api) = new_open_api_router();
     let router = router
@@ -157,12 +163,30 @@ impl IndexInfo {
             status = 200,
             description = "Successful operation. Returns an array of index information representing all indexes managed by the Vector Store.",
             body = [IndexInfo]
+        ),
+        (
+            status = 503,
+            description = "Service Unavailable. Indicates that the Vector Store node is not ready to serve requests",
+            content_type = "application/json",
+            body = ErrorMessage
         )
     )
 )]
-async fn get_indexes(State(state): State<RoutesInnerState>) -> response::Json<Vec<IndexInfo>> {
-    let ids = state.engine.get_index_ids().await;
-    let indexes = ids
+async fn get_indexes(State(state): State<RoutesInnerState>) -> Response {
+    let engine = state.engine.read().await;
+    if engine.is_none() {
+        return (
+            StatusCode::SERVICE_UNAVAILABLE,
+            "Vector Store node is not ready to serve requests",
+        )
+            .into_response();
+    }
+
+    let indexes: Vec<_> = engine
+        .as_ref()
+        .unwrap()
+        .get_index_ids()
+        .await
         .iter()
         .map(|id| IndexInfo {
             keyspace: id.keyspace(),
@@ -170,7 +194,7 @@ async fn get_indexes(State(state): State<RoutesInnerState>) -> response::Json<Ve
             quantization: Quantization::F32, // currently the only supported quantization by Vector Store
         })
         .collect();
-    response::Json(indexes)
+    (StatusCode::OK, response::Json(indexes)).into_response()
 }
 
 /// A human-readable description of the error that occurred.
@@ -206,6 +230,12 @@ struct ErrorMessage(#[allow(dead_code)] String);
             description = "Error while counting embeddings. Possible causes: internal error, or issues accessing the database.",
             content_type = "application/json",
             body = ErrorMessage
+        ),
+        (
+            status = 503,
+            description = "Service Unavailable. Indicates that the Vector Store node is not ready to serve requests",
+            content_type = "application/json",
+            body = ErrorMessage
         )
     )
 )]
@@ -213,8 +243,17 @@ async fn get_index_count(
     State(state): State<RoutesInnerState>,
     Path((keyspace, index)): Path<(KeyspaceName, IndexName)>,
 ) -> Response {
-    let Some((index, _)) = state
-        .engine
+    let engine = state.engine.read().await;
+    if engine.is_none() {
+        return (
+            StatusCode::SERVICE_UNAVAILABLE,
+            "Vector Store node is not ready to serve requests",
+        )
+            .into_response();
+    }
+    let Some((index, _)) = engine
+        .as_ref()
+        .unwrap()
         .get_index(IndexId::new(&keyspace, &index))
         .await
     else {
@@ -318,7 +357,7 @@ The similarity metric is determined at index creation and cannot be changed per 
         ),
         (
             status = 503,
-            description = "Service Unavailable. Indicates that a full scan of the index is in progress and the search cannot be performed at this time.",
+            description = "Service Unavailable. Indicates that the Vector Store node is not ready to serve requests or that a full scan of the index is in progress and the search cannot be performed at this time.",
             content_type = "application/json",
             body = ErrorMessage
         )
@@ -330,14 +369,23 @@ async fn post_index_ann(
     extract::Json(request): extract::Json<PostIndexAnnRequest>,
 ) -> Response {
     // Start timing
+    let engine = state.engine.read().await;
+    if engine.is_none() {
+        return (
+            StatusCode::SERVICE_UNAVAILABLE,
+            "Vector Store node is not ready to serve requests",
+        )
+            .into_response();
+    }
     let timer = state
         .metrics
         .latency
         .with_label_values(&[keyspace.as_ref().as_str(), index.as_ref().as_str()])
         .start_timer();
 
-    let Some((index, db_index)) = state
-        .engine
+    let Some((index, db_index)) = engine
+        .as_ref()
+        .unwrap()
         .get_index(IndexId::new(&keyspace, &index))
         .await
     else {
@@ -493,10 +541,10 @@ async fn get_info() -> response::Json<InfoResponse> {
     })
 }
 
-#[derive(ToEnumSchema, serde::Deserialize, serde::Serialize)]
+#[derive(ToEnumSchema, serde::Deserialize, serde::Serialize, Clone, PartialEq, Eq, Debug)]
 #[serde(rename_all = "SCREAMING_SNAKE_CASE")] // This line makes all variants uppercase in the schema
 /// Operational status of the Vector Store node.
-enum Status {
+pub enum Status {
     /// The node is starting up.
     Initializing,
     /// The node is establishing a connection to ScyllaDB.
@@ -518,8 +566,33 @@ enum Status {
         (status = 200, description = "Successful operation. Returns the current operational status of the Vector Store node.", body = Status),
     )
 )]
-async fn get_status() -> Response {
-    (StatusCode::NOT_IMPLEMENTED, "").into_response()
+async fn get_status(State(state): State<RoutesInnerState>) -> Response {
+    let mut status = state.node_state.lock().await;
+    if *status == Status::IndexingEmbeddings {
+        let ids = state
+            .engine
+            .read()
+            .await
+            .as_ref()
+            .unwrap()
+            .get_index_ids()
+            .await;
+
+        for id in ids {
+            let engine = state.engine.read().await.clone();
+            if let Some(index) = engine.unwrap().get_index(id).await {
+                if matches!(index.1.full_scan_progress().await, Progress::InProgress(..)) {
+                    return (
+                        StatusCode::SERVICE_UNAVAILABLE,
+                        response::Json(Status::IndexingEmbeddings),
+                    )
+                        .into_response();
+                }
+            }
+        }
+        *status = Status::Serving;
+    }
+    (StatusCode::OK, response::Json(status.clone())).into_response()
 }
 
 #[cfg(test)]
--- a/crates/vector-store/src/httpserver.rs
+++ b/crates/vector-store/src/httpserver.rs
@@ -10,17 +10,23 @@ use crate::metrics::Metrics;
 use std::net::SocketAddr;
 use std::sync::Arc;
 use tokio::net::TcpListener;
+use tokio::sync::Mutex;
 use tokio::sync::Notify;
+use tokio::sync::RwLock;
 use tokio::sync::mpsc;
 use tokio::sync::mpsc::Sender;
 
-pub(crate) enum HttpServer {}
+pub(crate) type WrapEngine = Arc<RwLock<Option<Sender<Engine>>>>;
+
+pub enum ServerMsg {
+    SetEngine(Sender<Engine>),
+    ChangeState(httproutes::Status),
+}
 
 pub(crate) async fn new(
     addr: HttpServerAddr,
-    engine: Sender<Engine>,
     metrics: Arc<Metrics>,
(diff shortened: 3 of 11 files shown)


crates/vector-store/src/monitor_indexes.rs
```diff
@@ -5,10 +5,12 @@
 
 use crate::IndexMetadata;
 use crate::SpaceType;
+use crate::change_state;
 use crate::db::Db;
 use crate::db::DbExt;
 use crate::engine::Engine;
 use crate::engine::EngineExt;
+use crate::httpserver::ServerMsg;
```
knowack1, line 13: I think engine (business logic) must not depend on httpserver.
Instead HttpServer should listen to engine state.

It looks like httproutes::Status should be moved to enging (business logic). Or Engine should have dedicated enum.

crates/vector-store/src/httpserver.rs
```diff
@@ -30,16 +36,52 @@ pub(crate) async fn new(
...
+            while let Some(msg) = rx.recv().await {
+                match msg {
+                    ServerMsg::SetEngine(engine_sender) => {
+                        let mut engine_lock = engine.write().await;
+                        *engine_lock = Some(engine_sender);
+                    }
+                    ServerMsg::ChangeState(state) => {
+                        let mut state_lock = node_state.lock().await;
+                        match *state_lock {
+                            httproutes::Status::Initializing => {
```
knowack1, line 55: what is the reason to check previous state?
I assume that any transition is allowed, when reflects current state. E.g. Serving -> ConnectingToDb after connection loss.

crates/vector-store/src/httpserver.rs
```diff
@@ -30,16 +36,52 @@ pub(crate) async fn new(
...
+    let node_state = Arc::new(Mutex::new(httproutes::Status::Initializing));
     tokio::spawn({
+        let engine = engine.clone();
+        let node_state = node_state.clone();
         let notify = Arc::clone(&notify);
         async move {
-            while rx.recv().await.is_some() {}
+            while let Some(msg) = rx.recv().await {
+                match msg {
+                    ServerMsg::SetEngine(engine_sender) => {
```
knowack1, line 48: I preferred the previous approach when engine was passed to new(). We do not need to force INITIALIZING state if it does not make sense, from businesses logic perspective.

crates/vector-store/src/main.rs
```diff
@@ -52,15 +58,10 @@ async fn main() -> anyhow::Result<()> {
         vector_store::new_index_factory_usearch()?
     };
 
+    change_state(&server_actor, httproutes::Status::ConnectingToDb).await?;
```
knowack1, line 61: this state must come from businesses logic  implementation. Should be strictly correlated with the code performing connection:
```
let session = Arc::new(
            SessionBuilder::new()
                .known_node(uri.0.as_str())
                .build()
                .await?,
        );
```

knowack1 (changes requested):

crates/vector-store/src/httpserver.rs
smoczy123: I think that the bigger issue here would be the lack of CONNECTING_TO_DB state, as before we needed the connection to already be established to create the HTTP server. I also don't like this new approach, but I'm not sure if there's a better way to execute this.

crates/vector-store/src/httpserver.rs
```diff
@@ -30,16 +36,52 @@ pub(crate) async fn new(
...
+                        let mut engine_lock = engine.write().await;
+                        *engine_lock = Some(engine_sender);
+                    }
+                    ServerMsg::ChangeState(state) => {
+                        let mut state_lock = node_state.lock().await;
+                        match *state_lock {
+                            httproutes::Status::Initializing => {
+                                if state == httproutes::Status::ConnectingToDb {
```
knowack1, line 56: I think Serving -> ConnectingToDb will also be valid transition. After connection loss to DB. However this behavior is not in the scope of this task.

smoczy123:
Changelog:

- Added NodeState actor to serve as a intermediate between other actors to keep the node state

crates/vector-store/src/state.rs
```diff
@@ -0,0 +1,41 @@
...
+pub enum Status {
+    Initializing,
+    ConnectingToDb,
+    DiscoveringIndexes,
+    IndexingEmbeddings,
+    Serving,
+}
+
+pub enum NodeState {
+    ChangeState(Status),
```
ewienik, line 16: The status should be calculated from `NodeState`, it shouldn't be provided to it. Only events should be provided to it. Actors should send events to the `NodeState` while they reached specific milestones in their life:
- db started connecting
- db finished connecting
- monitor indexes started to initial discovery
- monitor indexes finished initial discovery
- monitor indexes found such indexes initially
- db_index started to initial scan for specific index
- db_index finished initial scan for specific index
- etc...
These are example events, you can change them accordingly. And `NodeState` should calculate status based on these events.

ewienik (changes requested):

crates/vector-store/src/httpserver.rs
```diff
@@ -7,20 +7,26 @@ use crate::HttpServerAddr;
...
 use tokio::sync::Notify;
+use tokio::sync::RwLock;
 use tokio::sync::mpsc;
 use tokio::sync::mpsc::Sender;
 
-pub(crate) enum HttpServer {}
+pub(crate) type WrapEngine = Arc<RwLock<Option<Sender<Engine>>>>;
+
+pub enum ServerMsg {
+    SetEngine(Sender<Engine>),
```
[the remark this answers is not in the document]
ewienik: We can solve this in other way. Creating a db actor should not wait for connection to the db - in case there is no db ready we're finishing vector-store, which is bad for production, we should wait for the db. So probably we need to refactor db.rs that creating `Statements` will be in the `tokio::spawn` async context, not before.

crates/vector-store/src/state.rs
smoczy123: Hmm, I'm not sure about this, I think there easily could be events that are shared between actors and I don't think we could completely get rid of the dependency on the node_state as the actors have to know what are they allowed to send

crates/vector-store/src/lib.rs
```diff
@@ -445,13 +449,29 @@ pub async fn run(
...
 
-pub async fn new_db(uri: ScyllaDbUri) -> anyhow::Result<Sender<Db>> {
-    db::new(uri).await
+pub async fn send_event(node_state: &Sender<NodeState>, event: Event) -> anyhow::Result<()> {
+    node_state
+        .send(NodeState::ChangeState(event))
+        .await
+        .map_err(|_| anyhow::anyhow!("Failed to change state in the HTTP server"))
+}
+
```
ewienik, line 468: It should be moved to `NodeStateExt` trait in `node_state.rs`

crates/vector-store/src/node_state.rs
```diff
@@ -0,0 +1,76 @@
...
+    ConnectingToDb,
+    DiscoveringIndexes,
+    IndexingEmbeddings,
+    Serving,
+}
+
+pub enum Event {
+    StartedConnectingToDb,
+    ConnectedToDb,
+    StartedDiscoveringIndexes,
```
knowack1, line 24: DiscoveringIndexes,

crates/vector-store/src/node_state.rs
```diff
@@ -0,0 +1,76 @@
...
+    DiscoveredIndexes(HashSet<IndexMetadata>),
```
knowack1, line 25: IndexesDiscovered?

crates/vector-store/src/node_state.rs
```diff
@@ -0,0 +1,76 @@
...
+    FinishedFullScan(IndexMetadata),
```
knowack1, line 26: FullScanFinished

crates/vector-store/src/db.rs
```diff
@@ -197,24 +201,34 @@ impl DbExt for mpsc::Sender<Db> {
...
+) -> anyhow::Result<mpsc::Sender<Db>> {
     let (tx, mut rx) = mpsc::channel(10);
     tokio::spawn(
         async move {
+            let _ = send_event(&node_state, Event::StartedConnectingToDb).await;
+            let statements = Statements::new(uri).await;
+            if statements.is_err() {
+                tracing::error!("Failed to create statements");
+                return;
+            }
```
ewienik, line 216: It should be a loop with processing messages - returning error on every request and retrying to create Satements. For clarity it could be separate commit/PR.

crates/vector-store/tests/integration/usearch.rs
```diff
@@ -319,3 +338,12 @@ async fn ann_fail_while_building() {
         "Full scan is in progress, percentage: 33.33%"
     );
 }
+
+#[tokio::test]
+async fn status_returns_expected_values() {
```
knowack1, line 343: could we be more specific on the test case name ?

crates/vector-store/src/db_index.rs
```diff
@@ -280,6 +289,13 @@ impl Statements {
                         end.value().abs_diff(begin.value() - 1),
                         std::sync::atomic::Ordering::Relaxed,
                     );
+                    if scan_length.load(std::sync::atomic::Ordering::Relaxed) == u64::MAX {
+                        // notify that the initial scan is done
+                        let _ = node_state
+                            .clone()
+                            .send(NodeState::ChangeState(Event::FinishedFullScan(metadata)))
+                            .await;
+                    }
```
ewienik, line 298: Notification could be moved into `new` function - after `initial_scan` end. Then there will be no need to pass node_state down here.

crates/vector-store/src/node_state.rs
```diff
@@ -0,0 +1,76 @@
...
+
+pub enum NodeState {
+    ChangeState(Event),
+    GetStatus(oneshot::Sender<Status>),
+}
+
+pub(crate) async fn new() -> mpsc::Sender<NodeState> {
+    let (tx, mut rx) = mpsc::channel(10);
+
+    tokio::spawn(async move {
```
knowack1, line 37: Do we need to spawn "task while actions of  this actor  are not time consuming?

crates/vector-store/src/engine.rs
```diff
@@ -26,7 +28,7 @@ type GetIndexIdsR = Vec<IndexId>;
 type AddIndexR = anyhow::Result<()>;
 type GetIndexR = Option<(mpsc::Sender<Index>, mpsc::Sender<DbIndex>)>;
 
-pub(crate) enum Engine {
+pub enum Engine {
```
ewienik, line 31:
```suggestion
pub(crate) enum Engine {
```
Do we need a change in visibility?

crates/vector-store/src/httproutes.rs
```diff
@@ -518,8 +551,20 @@ enum Status {
...
-    (StatusCode::NOT_IMPLEMENTED, "").into_response()
+async fn get_status(State(state): State<RoutesInnerState>) -> Response {
+    let (tx, rx) = tokio::sync::oneshot::channel();
+    state
+        .node_state
+        .send(NodeState::GetStatus(tx))
+        .await
+        .unwrap_or_else(|_| {
+            tracing::warn!("get_status: failed to send GetStatus message to node state");
+        });
```
ewienik, line 562: It should be a part of `NodeStateExt` trait.

crates/vector-store/src/httpserver.rs
```diff
@@ -7,17 +7,19 @@ use crate::HttpServerAddr;
...
+use crate::node_state::NodeState;
 use std::net::SocketAddr;
 use std::sync::Arc;
 use tokio::net::TcpListener;
 use tokio::sync::Notify;
 use tokio::sync::mpsc;
 use tokio::sync::mpsc::Sender;
 
-pub(crate) enum HttpServer {}
+pub enum HttpServer {}
```
ewienik, line 18:
```suggestion
pub(crate) enum HttpServer {}
```
Do we need a change in visibility?

crates/vector-store/src/lib.rs
```diff
@@ -445,13 +449,29 @@ pub async fn run(
...
+        node_state.clone(),
+        engine::new(db_actor, index_factory, node_state).await?,
+        metrics,
+    )
+    .await
 }
 
-pub async fn new_db(uri: ScyllaDbUri) -> anyhow::Result<Sender<Db>> {
-    db::new(uri).await
+pub async fn send_event(node_state: &Sender<NodeState>, event: Event) -> anyhow::Result<()> {
```
knowack1, line 462: why not make it a method of NodeState?

crates/vector-store/src/node_state.rs
```diff
@@ -0,0 +1,76 @@
...
+pub enum Event {
+    StartedConnectingToDb,
+    ConnectedToDb,
+    StartedDiscoveringIndexes,
+    DiscoveredIndexes(HashSet<IndexMetadata>),
+    FinishedFullScan(IndexMetadata),
+}
+
+pub enum NodeState {
+    ChangeState(Event),
```
knowack1, line 30: the name of this operation does not necessarily reflects what it does.

crates/vector-store/src/main.rs
```diff
@@ -52,15 +54,18 @@ async fn main() -> anyhow::Result<()> {
         vector_store::new_index_factory_usearch()?
     };
 
-    let db_actor = vector_store::new_db(scylladb_uri).await?;
+    let db_actor = vector_store::new_db(scylladb_uri, node_state.clone()).await?;
+
     let (_server_actor, addr) = vector_store::run(
         vector_store_addr,
         background_threads,
+        node_state.clone(),
```
ewienik, line 62:
```suggestion
        node_state,
```

crates/vector-store/src/monitor_indexes.rs
```diff
@@ -45,6 +48,12 @@ pub(crate) async fn new(
                         if !schema_version.has_changed(&db).await {
                             continue;
                         }
+                        let _ = send_event(
+                            &node_state,
+                            crate::node_state::Event::StartedDiscoveringIndexes,
```
ewienik, line 53: `use crate::node_state::Event`

crates/vector-store/src/node_state.rs
```diff
@@ -0,0 +1,76 @@
...
+    GetStatus(oneshot::Sender<Status>),
+}
+
+pub(crate) async fn new() -> mpsc::Sender<NodeState> {
```
ewienik, line 34: You could add unit tests for this actor.

crates/vector-store/src/node_state.rs
```diff
@@ -0,0 +1,76 @@
...
+                        idxs = Some(idx.clone());
+                        if idx.is_empty() {
+                            status = Status::Serving;
+                        }
+                    }
+                },
+                NodeState::GetStatus(tx) => {
+                    tracing::info!("Current state: {:?}", status);
+                    tx.send(status).unwrap_or_else(|_| {
+                        tracing::warn!("Failed to send current state");
```
ewienik, line 68:
```suggestion
                        tracing::debug!("Failed to send current state");
```

crates/vector-store/src/node_state.rs
```diff
@@ -0,0 +1,76 @@
...
+    ChangeState(Event),
+    GetStatus(oneshot::Sender<Status>),
+}
+
+pub(crate) async fn new() -> mpsc::Sender<NodeState> {
+    let (tx, mut rx) = mpsc::channel(10);
+
+    tokio::spawn(async move {
+        let mut status = Status::Initializing;
+        let mut idxs = Option::<HashSet<IndexMetadata>>::None;
```
ewienik, line 39: Do we need `Option` here? Empty HashSet could have `None` meaning.

crates/vector-store/src/node_state.rs
```diff
@@ -0,0 +1,76 @@
...
+    FinishedFullScan(IndexMetadata),
+}
+
+pub enum NodeState {
+    ChangeState(Event),
+    GetStatus(oneshot::Sender<Status>),
+}
+
+pub(crate) async fn new() -> mpsc::Sender<NodeState> {
+    let (tx, mut rx) = mpsc::channel(10);
```
ewienik, line 35: `10` could be defined const

crates/vector-store/src/node_state.rs
```diff
@@ -0,0 +1,76 @@
...
+                            idxs = Some(indexes);
+                        }
+                    }
+                    Event::FinishedFullScan(metadata) => {
+                        let mut idx = idxs.unwrap().clone();
+                        idx.remove(&metadata);
+                        idxs = Some(idx.clone());
+                        if idx.is_empty() {
+                            status = Status::Serving;
+                        }
```
ewienik, line 62: Could it be done without cloning?

crates/vector-store/tests/integration/info.rs
```diff
@@ -9,10 +9,12 @@ use std::net::SocketAddr;
 
 #[tokio::test]
 async fn get_applicaiton_info() {
-    let (db_actor, _) = db_basic::new();
+    let node_state = vector_store::new_node_state().await;
+    let (db_actor, _) = db_basic::new(node_state.clone());
     let (_server_actor, addr) = vector_store::run(
         SocketAddr::from(([127, 0, 0, 1], 0)).into(),
         Some(1),
+        node_state.clone(),
```
ewienik, line 17:
```suggestion
        node_state,
```

crates/vector-store/tests/integration/opensearch.rs
```diff
@@ -41,11 +41,13 @@ async fn simple_create_search_delete_index() {
     let (_server_actor, addr) = vector_store::run(
         SocketAddr::from(([127, 0, 0, 1], 0)).into(),
         Some(1),
+        node_state.clone(),
```
ewienik, line 44:
```suggestion
        node_state,
```

crates/vector-store/tests/integration/usearch.rs
```diff
@@ -14,12 +14,21 @@ use reqwest::StatusCode;
...
 use vector_store::Percentage;
 
-async fn setup_store() -> (IndexMetadata, HttpClient, DbBasic, impl Sized) {
-    let (db_actor, db) = db_basic::new();
+async fn setup_store() -> (
+    IndexMetadata,
+    HttpClient,
+    DbBasic,
+    impl Sized,
+    Sender<vector_store::node_state::NodeState>,
```
ewienik, line 27: `use vector_store::node_state::NodeState`

crates/vector-store/tests/integration/usearch.rs
```diff
@@ -64,32 +73,39 @@ async fn setup_store() -> (IndexMetadata, HttpClient, DbBasic, impl Sized) {
...
         Some(1),
+        node_state.clone(),
         db_actor,
         index_factory,
     )
     .await
     .unwrap();
 
-    (index, HttpClient::new(addr), db, server)
+    (index, HttpClient::new(addr), db, server, node_state.clone())
```
ewienik, line 83:
```suggestion
    (index, HttpClient::new(addr), db, server, node_state)
```

crates/vector-store/tests/integration/usearch.rs
```diff
@@ -174,11 +191,13 @@ async fn failed_db_index_create() {
     let (_server_actor, addr) = vector_store::run(
         SocketAddr::from(([127, 0, 0, 1], 0)).into(),
         Some(1),
+        node_state.clone(),
```
ewienik, line 194:
```suggestion
        node_state,
```

ewienik (changes requested):

crates/vector-store/src/node_state.rs
smoczy123: I'm not sure what would be the alternative here

smoczy123:
Changelog:

- Implemented @ewienik's and @knowack1's suggestions

crates/vector-store/src/db.rs
```diff
@@ -197,24 +201,34 @@ impl DbExt for mpsc::Sender<Db> {
...
+    uri: ScyllaDbUri,
+    node_state: Sender<NodeState>,
+) -> anyhow::Result<mpsc::Sender<Db>> {
     let (tx, mut rx) = mpsc::channel(10);
     tokio::spawn(
         async move {
+            let _ = node_state.send_event(Event::ConnectingToDb).await;
+            let mut statements = Statements::new(uri.clone()).await;
+            while statements.is_err() {
+                tracing::error!("Failed to create statements, retrying");
```
ewienik, line 214: add here a sleep. Maybe 1s?

crates/vector-store/src/db.rs
```diff
@@ -197,24 +201,34 @@ impl DbExt for mpsc::Sender<Db> {
...
+                statements = Statements::new(uri.clone()).await;
+            }
+            let _ = node_state.send_event(Event::ConnectedToDb).await;
+            let arc = Arc::new(statements.unwrap());
```
ewienik, line 218:
```suggestion
            let statements = Arc::new(statements.unwrap());
```

ewienik (changes requested):

ewienik (approved):

smoczy123:
Changelog: 

- Fixed weird autoinclude

crates/vector-store/src/node_state.rs
knowack1: ok got it. channel/message queue is required to ensure synchronization between send_event and get_status

PR #188, opened 2025-07-22
Add a counter for vector operations

amnonh:
This patch adds a counter for the number of times an index was modified. The counter is split between "update" (which includes both insert and update operations) and "delete".

The counter reflects the number of modification attempts, not their success.

This is how it looks:
    # HELP index_modified Number of modified items per index
    # TYPE index_modified counter
  index_modified{index_name="myindex",keyspace="mykey",operation="update"} 200

Reference: https://scylladb.atlassian.net/browse/VECTOR-63


amnonh:
When testing, I saw that the delete does propagate. I assume it's either a bug or not implemented yet.

PR #198, opened 2025-07-28
Rename embedding to vector

knowack1:
Rename 'embedding' to 'vector' in public API
    
Rename 'embedding' to 'vector' in public API.
Maintain backward compatibility through serde field aliasing to support existing clients using the "embedding" field name.
Verify backward compatibility with test.

References: VECTOR-148

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1036,6 +1036,7 @@ name = "httpclient"
 version = "0.1.0"
 dependencies = [
  "reqwest",
+ "serde",
  "serde_json",
  "vector-store",
 ]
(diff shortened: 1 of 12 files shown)


ewienik:
> Vector 148 rename embedding to vector

What is the meaning of `148`?
Is there any Jira issue for these changes to reference?

> Refactor httpclient to use httproutes::Status: This patch makes httproutes::Status public and uses it in integration tests httpclient. Support for serialization has been removed from node_status::Status.

Should these changes be in a single PR? IMHO I don't see connection between them. Maybe it is better to separate them.

PR #209, opened 2025-08-06
Test for status endpoint while discovering indexes

knowack1:


crates/vector-store/tests/integration/db_basic.rs
```diff
@@ -265,8 +274,11 @@ fn process_db(db: &DbBasic, msg: Db, node_state: Sender<NodeState>) {
             .map_err(|_| anyhow!("Db::LatestSchemaVersion: unable to send response"))
             .unwrap(),
 
-        Db::GetIndexes { tx } => tx
-            .send(Ok(db
+        Db::GetIndexes { tx } => {
+            if db.0.read().unwrap().simulate_endless_get_indexes_processing {
+                return;
```
ewienik, line 279: In case of return here we drop `tx` and then will be an error in `DbExt::get_indexes` (tx dropped, error during rx.await). Better would be to spawn a task with sleep inside and move tx into this task.

knowack1: Done. 
```Rust
        Db::GetIndexes { tx } => {
            tokio::spawn({
                let db = db.clone();
                async move {
                    if db.0.read().unwrap().simulate_endless_get_indexes_processing {
                        tokio::time::sleep(std::time::Duration::from_secs(3600)).await;
                    }
                    tx.send(Ok(db
                        .0
                        .read()
                        .unwrap()
                        .keyspaces
                        .iter()
                        .flat_map(|(keyspace_name, keyspace)| {
                            keyspace
                                .indexes
                                .iter()
                                .map(|(index_name, index)| DbCustomIndex {
                                    keyspace: keyspace_name.clone(),
                                    index: index_name.clone(),
                                    table: index.index.table_name.clone(),
                                    target_column: index.index.target_column.clone(),
                                })
                        })
                        .collect()))
                        .map_err(|_| anyhow!("Db::GetIndexes: unable to send response"))
                        .unwrap()
                }
            });
        }
```

ewienik: Now it moves every `GetIndex` request into the async task. I don't think there is a reason for this. Could we create this task only when it is needed?

```rust
if db.0.read().unwrap().simulate_endless_get_indexes_processing {
    tokio::spawn(async move {
        tokio::time::sleep(std::time::Duration::from_secs(3600)).await;
        drop(tx); // don't use it but move tx into the task explicitly.
    }
    return;
}
```

ewienik: Looks good. I think we should change the behavior only when flag is enabled.

ewienik (approved):

PR #210, opened 2025-08-06
Temporary use zero UUID as an index version

ewienik:
We're waiting for the VECTOR-142 issue, the issue will bring solution to assigning unique version to each index creation. We've got this version from the Scylla view version, which is not accessible anymore. This patch gives a workaround and creates a zero UUID for each index. We won't discover a fast drop/create index with the same parameters, but it seems not urgent an edge case (we simply didn't recreate index in that situation).

### DIFF
--- a/crates/vector-store/src/db.rs
+++ b/crates/vector-store/src/db.rs
@@ -285,7 +285,6 @@ struct Statements {
     session: Arc<Session>,
     st_latest_schema_version: PreparedStatement,
     st_get_indexes: PreparedStatement,
-    st_get_index_version: PreparedStatement,
     st_get_index_target_type: PreparedStatement,
     st_get_index_options: PreparedStatement,
     re_get_index_target_type: Regex,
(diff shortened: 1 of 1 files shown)


QuerthDP:
One thing I'm concerned about is that the PR changed the Jira Ticket status, although it does not reference it in `Refs:` or `Fixes:`.
I'm curious if it will be marked `Done` after merging this.

QuerthDP:
> > One thing I'm concerned about is that the PR changed the Jira Ticket status, although it does not reference it in `Refs:` or `Fixes:`. I'm curious if it will be marked `Done` after merging this.
> 
> I've put the name [VECTOR-142](https://scylladb.atlassian.net/browse/VECTOR-142) in the description - this links PR with the issue. It shouldn't change the PR to done as there is no information that it fixes the issue :-)

I think you were right, as nothing yet happened.
Although I'm not sure if tagging the issue as `In Progress` does tell the truth.
Maybe yes, as it relates to the issue - but the main task is not yet `In Progress`.

PR #216, opened 2025-08-12
Add integration tests for serialization

smoczy123:
We have found a bug that caused an error when a date/time/timestamp was a primary key.
In this PR we fix it and add integration tests that check for regressions in serializing primary keys.

Fixes: VECTOR-178
Fixes: VECTOR-206

### DIFF
--- a/crates/validator/src/tests/serde.rs
+++ b/crates/validator/src/tests/serde.rs
@@ -1,0 +1,110 @@
+/*
+ * Copyright 2025-present ScyllaDB
+ * SPDX-License-Identifier: LicenseRef-ScyllaDB-Source-Available-1.0
+ */
+
+use crate::{common::wait_for, tests::*};
+use scylla::value::CqlValue;
+
+pub(crate) async fn new() -> TestCase {
+    let timeout = Duration::from_secs(30);
+    TestCase::empty()
+        .with_init(timeout, crate::common::init)
+        .with_cleanup(timeout, crate::common::cleanup)
+        .with_test(
+            "test_serialization_deserialization_all_types",
+            timeout,
+            test_serialization_deserialization_all_types,
+        )
+}
+
+async fn test_serialization_deserialization_all_types(actors: TestActors) {
+    let (session, _client) = crate::common::prepare_connection(actors).await;
+
+    let cases = vec![
+        ("ascii", "'random_text'"),
+        ("bigint", "1234"),
+        ("boolean", "true"),
+        ("date", "'2023-10-01'"),
+        ("double", "3.14159"),
+        ("float", "2.71828"),
+        ("int", "42"),
+        ("smallint", "123"),
+        ("tinyint", "7"),
+        ("uuid", "841685b2-8803-11f0-8de9-0242ac120002"),
+        ("timeuuid", "841685b2-8803-11f0-8de9-0242ac120002"),
+        ("time", "'08:12:54.2137'"),
+        ("timestamp", "'2023-10-01T12:34:56.789Z'"),
+        ("text", "'some_text'"),
+    ];
+
+    session.query_unpaged(
+        "CREATE KEYSPACE ks WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 1}",
+        (),
+    ).await.expect("failed to create a keyspace");
+
+    session
+        .use_keyspace("ks", false)
+        .await
+        .expect("failed to use a keyspace");
+
+    for (typ, data) in &cases {
+        session
+            .query_unpaged(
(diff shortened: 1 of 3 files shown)


crates/vector-store/src/httproutes.rs
```diff
@@ -604,20 +631,48 @@ mod tests {
...
         );
         assert_eq!(
             to_json(CqlValue::Time(now.time().into())),
-            Value::String(now.format(&Iso8601::TIME).unwrap())
+            Value::String(
+                now.format({
+                    const CONFIG: u128 = Config::DEFAULT
+                        .set_formatted_components(FormattedComponents::Time)
+                        .set_time_precision(TimePrecision::Second {
+                            decimal_digits: NonZero::new(3),
```
swasik, line 655: This is a bit tricky. We should make sure that we allow to represent nanoseconds for time. Excerpt from the documentation for C++ driver (https://cpp-driver.docs.scylladb.com/master/topics/basics/date_and_time/):

The date type uses an unsigned 32-bit integer (cass_uint32_t) to represent the number of days with Epoch centered at 2^31. Because it’s centered at Epoch it can be used to represent days before Epoch. The time type uses a signed 64-bit integer (cass_int64_t) to represent the number of nanoseconds since midnight and valid values are in the range 0 to 86399999999999.

Keep in mind that the key is used to merge data back in Scylla so we have to keep the exactly same precision as in Scylla. In this case nanoseconds.

smoczy123:
Changelog:

- Addressed @knowack1 comments

crates/validator/src/tests/serde.rs
```diff
@@ -0,0 +1,111 @@
...
+
+        session
+            .query_unpaged(
+                format!("CREATE INDEX idx ON tbl_{typ}(vec) USING 'vector_index'"),
+                (),
+            )
+            .await
+            .expect("failed to create an index");
+
+        while client.indexes().await.is_empty() {}
```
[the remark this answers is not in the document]
knowack1: What we expect is to have search result in some reasonable time. I would assert this expectation.
Also having indexes returned by this API does not guarantee that data will be returned from ANN OF query.

smoczy123:
Changelog:

- Removed all references to "now" from serialization unit tests

crates/vector-store/src/httproutes.rs
```diff
@@ -489,7 +489,10 @@ fn to_json(value: CqlValue) -> Value {
             TryInto::<Time>::try_into(value)
                 .expect("CqlValue::Time should be correct")
                 .format(&Iso8601::TIME)
-                .expect("Time should be correct"),
+                .expect("Time should be correct")
+                .strip_prefix("T")
+                .unwrap_or("")
```
QuerthDP, line 494: Why do we unwrap to empty string if the strip fails?
Do we allow empty time value?

crates/validator/src/tests/serde.rs
```diff
@@ -0,0 +1,112 @@
...
+    session
+        .use_keyspace("ks", false)
+        .await
+        .expect("failed to use a keyspace");
+
+    for (typ, data) in &cases {
+        session
+            .query_unpaged(
+                format!(
+                    "CREATE TABLE tbl_{typ} (id {typ} PRIMARY KEY, vec vector<float, 3>) WITH cdc = {{'enabled': true }}"
```
QuerthDP, line 55: CDC is now auto-enabled by creating the vector index. No need to specify it here, but that's ok.

ewienik (approved):

QuerthDP (approved):
It would be appreciated to have more descriptive commit messages 😉

crates/vector-store/src/httproutes.rs
```diff
@@ -489,7 +489,10 @@ fn to_json(value: CqlValue) -> Value {
             TryInto::<Time>::try_into(value)
                 .expect("CqlValue::Time should be correct")
                 .format(&Iso8601::TIME)
-                .expect("Time should be correct"),
+                .expect("Time should be correct")
+                .strip_prefix("T")
+                .unwrap() // Safe to unwrap as format always adds 'T' prefix
```
QuerthDP, line 494: Good to add this comment. Nice one.

QuerthDP (approved):
Thanks for adding the commit messages! Looks much better now :)

crates/vector-store/src/httproutes.rs
ewienik, line 494: Better to use expect.

```suggestion
                .expect("format should always add 'T' prefix")
```

By the way - there is a problem with this function, because problems with conversions could crash the service. I'm creating a new issue for this.

PR #218, opened 2025-08-13
Add ScyllaDB authentication support

knowack1:
This commit enables the vector-store service to authenticate with ScyllaDB using a username and password.

Credentials are provided via environment variables:
- `VECTOR_STORE_SCYLLADB_USERNAME`: The database username.
- `VECTOR_STORE_SCYLLADB_PASSWORD_FILE`: The path to a file containing the password.

The password is read from the specified file and trimmed of whitespace. Using a file for the password enhances security by preventing the secret from being exposed in the environment or process list.

Unit tests are included to verify various configuration scenarios, including missing credentials, invalid configurations, and successful loading.

References: VECTOR-67


crates/vector-store/src/main.rs
```diff
@@ -15,6 +16,28 @@ mod info;
 #[clap(version)]
 struct Args {}
 
+async fn credentials<F>(env: F) -> anyhow::Result<Option<vector_store::Credentials>>
+where
+    F: Fn(&'static str) -> Result<String, std::env::VarError>,
+{
+    let Ok(username) = env("VECTOR_STORE_SCYLLADB_USERNAME") else {
+        return Ok(None);
+    };
```
swasik, line 25: nit: I think that it is worth checking if the username is not empty. Depending on how the authentication errors are formatted it is sometimes non-obvious to find out that the variable was defined but not set to any specific value.

crates/vector-store/src/main.rs
```diff
@@ -77,3 +105,93 @@ async fn main() -> anyhow::Result<()> {
...
+        let result = credentials(env).await;
+
+        assert!(result.is_err());
+    }
+
+    #[tokio::test]
+    async fn credentials_success() {
+        let mut password_file = NamedTempFile::new().unwrap();
+        writeln!(password_file, "my_secret_pass").unwrap();
+        let env = mock_env(HashMap::from([
```
swasik, line 166: I am not sure about Rust conventions but I would move the values for test user name and test password to some constants common for those tests. It will also unify them across the test - now you use two different names for the user.

swasik (changes requested):
I added Copilot just to play with it but I think it added interesting comments worth at least commenting.

crates/vector-store/src/main.rs
knowack1: Done

crates/vector-store/src/main.rs
```diff
@@ -15,6 +16,28 @@ mod info;
...
+        return Ok(None);
+    };
+    let Ok(password_file) = env("VECTOR_STORE_SCYLLADB_PASSWORD_FILE") else {
+        bail!(
+            "credentials: VECTOR_STORE_SCYLLADB_PASSWORD_FILE env required when VECTOR_STORE_SCYLLADB_USERNAME is set"
+        );
+    };
+    let password = tokio::fs::read_to_string(&password_file)
+        .await
+        .map_err(|e| anyhow!("credentials: failed to read password file: {}", e))?;
```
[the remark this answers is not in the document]
knowack1: I think that file path is not confidential. Also example code where file is not present does not contain file path in its message. Logging error code seems to be essential to understand what is wrong with the file.

knowack1:
> I added Copilot just to play with it but I think it added interesting comments worth at least commenting.

The comments are very valuable. One is fixed one is answered.

crates/vector-store/src/main.rs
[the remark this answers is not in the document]
swasik: I agree - security by obscurity was never the most efficient security practice.

swasik (approved):

PR #220, opened 2025-08-14
Fix date/time serialization bug

smoczy123:
We have found a bug that caused an error when a timestamp was a primary key.
In this PR we fix it. It was manually tested, integration tests will be added in a followup PR.

### DIFF
--- a/crates/vector-store/src/httproutes.rs
+++ b/crates/vector-store/src/httproutes.rs
@@ -42,11 +42,14 @@ use scylla::value::CqlValue;
 use serde_json::Number;
 use serde_json::Value;
 use std::collections::HashMap;
+use std::num::NonZero;
 use std::sync::Arc;
 use time::Date;
 use time::OffsetDateTime;
 use time::Time;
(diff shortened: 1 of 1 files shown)


crates/vector-store/src/httproutes.rs
```diff
@@ -488,7 +491,14 @@ fn to_json(value: CqlValue) -> Value {
         CqlValue::Timestamp(value) => Value::String(
             TryInto::<OffsetDateTime>::try_into(value)
                 .expect("CqlValue::Timestamp should be correct")
-                .format(&Iso8601::DEFAULT)
+                .format({
+                    const CONFIG: u128 = Config::DEFAULT
+                        .set_time_precision(TimePrecision::Second {
+                            decimal_digits: NonZero::new(3),
```
swasik, line 497: Are you sure that precision of 3 is correct here? Does it map to nanoseconds precision in Scylla?

BTW: Did you checked in your manual test scenario when nanoseconds are different than 0?

PR #225, opened 2025-08-25
Add logging to connection creation

smoczy123:
This PR adds logging to connection creation so we know when did we connect and the version of Scylla running on the node that we connected to.


crates/vector-store/src/db.rs
```diff
@@ -301,6 +303,31 @@ impl Statements {
...
+                .query_unpaged(
+                    "SELECT version FROM system.versions WHERE key = 'local'",
+                    &[],
+                )
+                .await?
+                .into_rows_result()?
+                .single_row()?;
+            info!("Connected to ScyllaDB {} at {}", version.0, uri.0);
+        } else {
+            warn!("Connection to an unknown entity at {}", uri.0);
```
QuerthDP, line 328: Could we add some information to clearly state we did not connect to ScyllaDB?
Something like: "Could not connect to ScyllaDB. Connected to an unknown entity..."

crates/vector-store/src/db.rs
```diff
@@ -301,6 +303,31 @@ impl Statements {
         }
 
         let session = Arc::new(builder.build().await?);
+
+        let cluster_state = session.get_cluster_state();
+
+        let node = &cluster_state.get_nodes_info()[0];
+
+        if !node.is_enabled() {
+            return Err(anyhow::anyhow!("Node is not enabled"));
```
QuerthDP, line 312: ❔What does this exactly tell us? I'm generally curious, I believe it's ok.

crates/vector-store/src/db.rs
QuerthDP: That's a fair point. So maybe we should tell the user to verify if the given ScyllaDB URI parameter is correct when connected to different instance.

crates/vector-store/src/db.rs
smoczy123: If I understand correctly, in our case this doesn't tell us anything, but I think it is a good idea to keep it here as it is more verbose

QuerthDP:
> Reconnecting is a much more difficult issue, as after we connect the connection is being managed internally by the driver/scylla-cdc so we don't know when/if we have reconnected

That's a pity. But I believe the issue is only partially fixed by this PR if we could not support reconnecting.
Maybe we could access some information about connection from the driver?

swasik:
> But I believe the issue is only partially fixed by this PR if we could not support reconnecting.
> Maybe we could access some information about connection from the driver?

Connection is higher priority as it allows us to see if the configuration is correct at all. 

For re-connection I suggest creating a separate issue and adding it to M2. We can always modify the driver to share the connection status changes.

QuerthDP (approved):
> Connection is higher priority as it allows us to see if the configuration is correct at all.

Ok. I do approve the changes then. But please let's create a Jira Ticket about reconnecting for future development.

PR #226, opened 2025-08-26
Make error about full scan progress more descriptive

swasik:
The current error message about full scan being in progress does not explain WHY the service is unavailable. It requires understanding internal architecture - how the index is constructed using full scan to understand this. This should be improved.

Extra improvements:
- Use 3 decimal digits as for large indexes 2 can be too few.
- Change formatting of qualified index names to use “.” instead of “/”.

Fixes VECTOR-190


swasik:
> It also fixes: [VECTOR-160](https://scylladb.atlassian.net/browse/VECTOR-160)

Thanks! I updated the status.

PR #227, opened 2025-08-27
db_index: run `initial_scan` once

QuerthDP:
When received a message during initial scanning, the `initial_scan` future was built every time again.
Due to the bug, the completed_scan_length counter was incremented to many times, leading the counter to overflow, which lead into the infinite loop described in the Jira ticket.

Now it's built outside the select once and is still listened to concurrently with listening for messages.

Fixes: VECTOR-192

### DIFF
--- a/crates/httpclient/src/lib.rs
+++ b/crates/httpclient/src/lib.rs
@@ -12,6 +12,8 @@ use vector_store::ColumnName;
 use vector_store::Distance;
 use vector_store::IndexInfo;
 use vector_store::IndexMetadata;
+use vector_store::IndexName;
+use vector_store::KeyspaceName;
 use vector_store::Limit;
 use vector_store::Vector;
 use vector_store::httproutes::InfoResponse;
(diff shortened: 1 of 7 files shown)


knowack1:
> I wanted to add tests to this PR, but I believe the [VECTOR-114](https://scylladb.atlassian.net/browse/VECTOR-114) is required to accomplish this. Maybe I'm wrong, I will look into it further.

Nice investigation. The problem is really tricky.
Indeed reliable test requires control over range scanning - this driver call:
```
.session
            .execute_iter(self.st_range_scan.clone(), (begin.value(), end.value()))
```

swasik:
> And as the Validator test failed it seems that the test does not work on `scylladb/scylladb-releng:2025.4.0-dev-0.20250718.d36c9b3119b7-x86_64` on which the CI is run. However it does work on the newest build (`scylladb/scylla-nightly` for example). The difference comes from that the error message is more descriptive now since [scylladb/scylladb#24444](https://github.com/scylladb/scylladb/pull/24444) was merged.

We had plenty of changes merged recently - I think it would be worth to bump the version used by validator (in a separate PR of course).

QuerthDP:
Changelog:

- rebased on master
- added descriptive commit messages

ewienik:
> Validator tests should fail due to [VECTOR-197](https://scylladb.atlassian.net/browse/VECTOR-197).

The problem could be with DNS TTL - currently it is set to 60s: https://github.com/scylladb/vector-store/blob/master/crates/validator/src/dns.rs#L119. I think it could be solved with this PR. Check with TTL set to 1, and add a async sleep for TTL in upsert.

ewienik (changes requested):

QuerthDP:
Changelog:

- refactored the httpclient `count` function not to use `IndexMetadata` object
- made the `wait_for` timeout adjustable as a function parameter

QuerthDP:
Changelog:

- rebased on master

ewienik (approved):

PR #228, opened 2025-08-29
workflows: use recent ScyllaDB docker image in CI

QuerthDP:
Use the most recently built docker image from scylladb/master branch.

### DIFF
--- a/.github/workflows/validator.yml
+++ b/.github/workflows/validator.yml
@@ -32,5 +32,5 @@ jobs:
       run: ./run-with-release-toolchain cargo build --release --workspace
 
     - name: Run the Validator tests
-      run: ./run-validator-with-scylla-docker scylladb/scylladb-releng:2025.4.0-dev-0.20250718.d36c9b3119b7-x86_64 target/release/vector-store target/release/vector-search-validator --verbose
+      run: ./run-validator-with-scylla-docker scylladb/scylla-nightly target/release/vector-store target/release/vector-search-validator --verbose
 


.github/workflows/validator.yml
```diff
@@ -32,5 +32,5 @@ jobs:
       run: ./run-with-release-toolchain cargo build --release --workspace
 
     - name: Run the Validator tests
-      run: ./run-validator-with-scylla-docker scylladb/scylladb-releng:2025.4.0-dev-0.20250718.d36c9b3119b7-x86_64 target/release/vector-store target/release/vector-search-validator --verbose
+      run: ./run-validator-with-scylla-docker scylladb/scylla-nightly target/release/vector-store target/release/vector-search-validator --verbose
```
knowack1, line 35: Using the scylla-nightly build in our CI could introduce instability, causing failures even when no code has changed in this repository. To ensure our CI remains stable, better we pin it to a specific version and manage upgrades manually.

QuerthDP: This way or another - it would be beneficial for us to update the docker image as there were some crucial changes introduced since the last update.

swasik: But as for now, if I am correct, we do not have a counterpart in scylladb repo, i.e. integration tests that run vector-store for each PR. If I am up to date this is a plan but not implemented yet. So as for now, this is the only way to validate that we work with the fresh Scylla. In the future we may switch to pinned version again - @QuerthDP it would be great to create an issue for this.

QuerthDP: [VECTOR-199](https://scylladb.atlassian.net/browse/VECTOR-199)

QuerthDP:
After further investigation, we could indeed just update the docker version to the current nightly version: `scylladb/scylla-nightly:2025.4.0-dev-0.20250829.98b1dda753c7-x86_64`. These versions are kept safe and sound [here](https://hub.docker.com/r/scylladb/scylla-nightly/tags).

@swasik @knowack1 what do you think?

.github/workflows/validator.yml
knowack1: I suppose Szymon meant a PR in ScyllaDB. OK, let's stick to scylla-nightly:latest. If this causes our CI to be flaky, we can easily take a step back and stick to a specific Scylla build.

knowack1 (approved):

PR #229, opened 2025-08-29
validator: move test infrastructure to `common.rs`

QuerthDP:
Move initialisation, cleanup and connection preparation to the common file in validator crate.


crates/validator/src/tests/crud.rs
knowack1, line 1: Does the movement make sense when the code is still used in one place?

QuerthDP: It's for future development.
Being used already here: https://github.com/scylladb/vector-store/pull/227

Wanted to split the move into a separate PR.

knowack1 (approved):

PR #230, opened 2025-09-01
validator: fix removing dns records

ewienik:
validator: fix removing dns records

InMemory authority (from hickory-server) does not support direct removing dns entries. Current solution uses zero DNS record to simulate removing, but it doesn't work as it dissallows to insert a new entry with the same name. This patch introduces a new API for removing entry and removes the DNS record from a map of the records directly.

Fixes: VECTOR-197


QuerthDP (approved):
Looks good. Works as intended.

PR #231, opened 2025-09-03
Fix fullscan hang when losing connection

smoczy123:
We have detected an annoying issue that stops the fullscan without a way to resume when the connection is lost, losing hours of index building time. This PR fixes the issue by adding a retry loop to the fullscan.

Fixes: VECTOR-193

### DIFF
--- a/crates/validator/src/scylla_cluster.rs
+++ b/crates/validator/src/scylla_cluster.rs
@@ -4,6 +4,7 @@
  */
 
 use std::net::Ipv4Addr;
+use std::path::Path;
 use std::path::PathBuf;
 use std::process::Stdio;
 use std::time::Duration;
@@ -31,6 +32,11 @@ pub(crate) enum ScyllaCluster {
         tx: oneshot::Sender<bool>,
     },
     Stop,
+    Up {
+        vs_uri: String,
+        conf: Option<Vec<u8>>,
+    },
+    Down,
 }
 
 pub(crate) trait ScyllaClusterExt {
@@ -45,6 +51,12 @@ pub(crate) trait ScyllaClusterExt {
 
     /// Waits for the ScyllaDB cluster to be ready.
     async fn wait_for_ready(&self) -> bool;
+
+    /// Starts a paused instance back again.
+    async fn up(&self, vs_uri: String, conf: Option<Vec<u8>>);
+
+    /// Pauses an instance
+    async fn down(&self);
 }
 
 impl ScyllaClusterExt for mpsc::Sender<ScyllaCluster> {
@@ -80,6 +92,18 @@ impl ScyllaClusterExt for mpsc::Sender<ScyllaCluster> {
             .expect("ScyllaClusterExt::wait_for_ready: internal actor should receive request");
         rx.await
             .expect("ScyllaClusterExt::wait_for_ready: internal actor should send response")
+    }
+
+    async fn up(&self, vs_uri: String, conf: Option<Vec<u8>>) {
+        self.send(ScyllaCluster::Up { vs_uri, conf })
+            .await
+            .expect("ScyllaClusterExt::up: internal actor should receive request")
+    }
+
+    async fn down(&self) {
+        self.send(ScyllaCluster::Down)
+            .await
+            .expect("ScyllaClusterExt::down: internal actor should receive request")
     }
 }
 
@@ -175,13 +199,26 @@ async fn process(msg: ScyllaCluster, state: &mut State) {
             tx.send(wait_for_ready(state).await)
                 .expect("process ScyllaCluster::WaitForReady: failed to send a response");
         }
+
+        ScyllaCluster::Up { vs_uri, conf } => {
+            up(vs_uri, conf, state).await;
+        }
+
+        ScyllaCluster::Down => {
+            down(state).await;
+        }
     }
 }
 
-async fn start(vs_uri: String, db_ip: Ipv4Addr, conf: Option<Vec<u8>>, state: &mut State) {
-    let workdir = TempDir::new().expect("start: failed to create temporary directory for scylladb");
+async fn run_cluster(
+    vs_uri: &String,
+    db_ip: &Ipv4Addr,
+    conf: &Option<Vec<u8>>,
+    path: &Path,
+    state: &mut State,
+) {
     let conf = if let Some(conf) = conf {
-        let conf_path = workdir.path().join("scylla.conf");
+        let conf_path = path.join("scylla.conf");
         fs::write(&conf_path, conf)
             .await
             .expect("start: failed to write scylla config");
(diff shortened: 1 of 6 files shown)


smoczy123:
Changelog:

- fix clippy

crates/vector-store/src/db_index.rs
```diff
@@ -278,25 +278,36 @@ impl Statements {
...
-                    scan_length.fetch_add(
-                        end.value().abs_diff(begin.value() - 1),
-                        std::sync::atomic::Ordering::Relaxed,
-                    );
-                    drop(permit);
+            });
+            loop {
+                if range_scan.is_ok() {
+                    break;
+                }
```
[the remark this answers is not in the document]
ewienik: Some errors comes from driver, some errors comes from inconsistency of data types in db with expectations. In case of no repeatable errors we should have an entry in a logfile.

smoczy123:
> > > Is it possible to create test case for it?
> > 
> > 
> > It would require us to bring down a scylla instance during a fullscan. I'm not sure if it's possible to do something like that in our testing infrastructure without inserting an unwieldy amount of vectors.
> 
> This sounds like a good scenario for SCT - there should be Nemesis bringing down Scylla.

The issue here is not the interrupting (we can stop the cluster in the validator tests), but interrupting exactly during the fullscan. We can try to hit it with some success rate, as we usually hit it during other tests and have to wait for it to end, but I'm not sure if we can guarantee that we interrupt at the correct time.

crates/vector-store/src/db_index.rs
```diff
@@ -278,25 +278,36 @@ impl Statements {
         for (begin, end) in self.fullscan_ranges() {
             let permit = Arc::clone(&semaphore).acquire_owned().await.unwrap();
 
-            if let Ok(embeddings) = self.range_scan_stream(begin, end).await.inspect_err(|err| {
+            let mut range_scan = self.range_scan_stream(begin, end).await.inspect_err(|err| {
```
knowack1, line 281: encapsulate the new repeat logic in dedicated function.

smoczy123:
Changelog:

- Addressed @knowack1's remarks
- Added a validator test for the fix

crates/vector-store/src/db_index.rs
```diff
@@ -265,6 +273,40 @@ impl Statements {
...
+            if range_scan.is_ok() {
+                break;
+            }
+            tokio::time::sleep(retry_timeout).await;
+            range_scan = self.range_scan_stream(begin, end).await.inspect_err(|err| {
+                warn!("unable to do initial scan for range ({begin:?}, {end:?}): {err}")
+            });
+            if retry_timeout.as_secs() < 16 {
+                retry_timeout *= 2;
+            }
```
ewienik, line 305: nit: Could these constant be defined and documented, that you want to increase retry timeout and why?

crates/vector-store/src/db_index.rs
```diff
@@ -279,9 +321,9 @@ impl Statements {
         for (begin, end) in self.fullscan_ranges() {
             let permit = Arc::clone(&semaphore).acquire_owned().await.unwrap();
 
-            if let Ok(embeddings) = self.range_scan_stream(begin, end).await.inspect_err(|err| {
-                warn!("unable to do initial scan for range ({begin:?}, {end:?}): {err}")
-            }) {
+            let range_scan = self.preform_range_scan(begin, end).await;
+            if range_scan.is_ok() {
+                let embeddings = range_scan.unwrap();
```
ewienik, line 326: nit: `if let Ok(...)`

crates/vector-store/src/db_index.rs
```diff
@@ -297,6 +339,8 @@ impl Statements {
                     );
                     drop(permit);
                 });
+            } else {
+                drop(permit);
```
ewienik, line 343: nit: `drop(permit)` could be done once, after `if` block

crates/vector-store/src/db_index.rs
```diff
@@ -265,6 +273,40 @@ impl Statements {
...
+                    );
+                }
+            }
+            if range_scan.is_ok() {
+                break;
+            }
+            tokio::time::sleep(retry_timeout).await;
+            range_scan = self.range_scan_stream(begin, end).await.inspect_err(|err| {
+                warn!("unable to do initial scan for range ({begin:?}, {end:?}): {err}")
+            });
```
ewienik, line 302: nit: Is this warn still needed if there is error with the same err code?

crates/validator/src/scylla_cluster.rs
```diff
@@ -262,3 +280,60 @@ async fn wait_for_ready(state: &State) -> bool {
...
+            .arg(format!("seeds={db_ip}"))
+            .arg("--vector-store-uri")
+            .arg(vs_uri)
+            .arg("--developer-mode")
+            .arg("true")
+            .arg("--smp")
+            .arg("2")
+            .spawn()
+            .expect("start: failed to spawn scylladb"),
+    );
```
ewienik, line 338: This part is the same as for `start`. Could it be encapsulated in a separated abstraction/function?

ewienik (changes requested):

crates/vector-store/src/db_index.rs
```diff
@@ -265,6 +273,40 @@ impl Statements {
...
+                {
+                    error!("Fatal error during scan of the range ({begin:?}, {end:?}): {err}");
+                    break;
+                } else {
+                    warn!(
+                        "Lost connection during scan of the range ({begin:?}, {end:?}), retrying"
+                    );
+                }
+            }
+            if range_scan.is_ok() {
```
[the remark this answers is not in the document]
smoczy123: I'm not sure what you mean

crates/validator/src/tests/reconnect.rs
```diff
@@ -0,0 +1,112 @@
...
+
+    match &result {
+        Err(e) if format!("{e:?}").contains("503 Service Unavailable") => {}
+        _ => panic!("Expected SERVICE_UNAVAILABLE error, got: {result:?}"),
+    }
+
+    actors
+        .db
+        .restart(get_default_vs_url(&actors).await, None)
+        .await;
```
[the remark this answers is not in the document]
smoczy123: If scylla is down, we can't query it for the index, making a check like this impossible. We cannot *guarantee* that we will hit this error as expected, but this exact same behavior is used in other tests (full_scan.rs)

crates/vector-store/src/db_index.rs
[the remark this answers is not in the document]
ewienik: I mean this:
```
if let Err(err) = &range_scan {
...
} else {
...
```

smoczy123:
Changelog:

- rebased on master, 
- addresed most remarks

PR #233, opened 2025-09-03
Log more information about serving status and finishing full scan

QuerthDP:
This patch adds the debug information on the `node_state`, informs the user about SERVING status and when the full scan has been completed.

Fixes the issue that discovering indexes after the initial bootstrapping caused the status to change to BOOTSTRAPING again, although it is properly setup and does respond to the requests on built indexes.

Fixes: VECTOR-195

### DIFF
--- a/crates/vector-store/src/node_state.rs
+++ b/crates/vector-store/src/node_state.rs
@@ -191,4 +191,43 @@ mod tests {
             .await;
         assert_eq!(node_state.get_status().await, Status::Serving);
     }
+
+    #[tokio::test]
+    async fn status_remains_serving_when_discovering_indexes() {
+        let node_state = new().await;
+        // Move to Serving status
+        node_state.send_event(Event::ConnectingToDb).await;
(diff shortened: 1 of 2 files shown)


crates/vector-store/src/node_state.rs
```diff
@@ -57,44 +61,54 @@ pub(crate) async fn new() -> mpsc::Sender<NodeState> {
...
+                        Event::ConnectingToDb => {
+                            status = Status::ConnectingToDb;
                         }
-                        if status == Status::DiscoveringIndexes {
-                            status = Status::IndexingEmbeddings;
-                            idxs = indexes;
+                        Event::ConnectedToDb => {}
+                        Event::DiscoveringIndexes => {
+                            if status != Status::Serving {
+                                status = Status::DiscoveringIndexes;
```
ewienik, line 79: Needs a unit test

ewienik (changes requested):

ewienik (approved):

PR #235, opened 2025-09-04
Update workflow trigger to pull_request_target - fixing fork PR bug

dani-tweig:
The previous version had a problem: Fork PRs didn't manage to pass the Jira credentials to the main code, which updates the Jira key status.

### DIFF
--- a/.github/workflows/call_jira_status_in_progress.yml
+++ b/.github/workflows/call_jira_status_in_progress.yml
@@ -1,11 +1,11 @@
 name: Call Jira Status In Progress
 
 on:
-  pull_request:
+  pull_request_target:
     types: [opened]
 
 jobs:
   call-jira-status-in-progress:
     uses: scylladb/github-automation/.github/workflows/main_update_jira_status_to_in_progress.yml@main
     secrets:
-      jira_auth: ${{ secrets.USER_AND_KEY_FOR_JIRA_AUTOMATION }}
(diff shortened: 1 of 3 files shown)


.github/workflows/call_jira_status_in_progress.yml
```diff
@@ -1,11 +1,11 @@
...
-  pull_request:
+  pull_request_target:
     types: [opened]
 
 jobs:
   call-jira-status-in-progress:
     uses: scylladb/github-automation/.github/workflows/main_update_jira_status_to_in_progress.yml@main
     secrets:
-      jira_auth: ${{ secrets.USER_AND_KEY_FOR_JIRA_AUTOMATION }}
+      caller_jira_auth: ${{ secrets.USER_AND_KEY_FOR_JIRA_AUTOMATION }}
```
QuerthDP, line 11: Does this change anything or is it just a cosmetic rename?

QuerthDP (approved):

.github/workflows/call_jira_status_in_progress.yml
dani-tweig: Cosmetic and separates the credentials passed by the calling workflow, with internal credentials passed in the main logic.
It also tells me which of the workflows for all the repos that have the call workflows were not updated. (they will issue an error on execution)

PR #238, opened 2025-09-08
db: add more information in the connection error

smoczy123:
This patch adds more information to the log message printed whenever a connection to ScyllaDB fails. Now we will log the error message and the expected IP.

Fixes: VECTOR-165


crates/vector-store/src/db.rs
```diff
@@ -219,7 +219,11 @@ pub(crate) async fn new(
             node_state.send_event(Event::ConnectingToDb).await;
             let mut statements = Statements::new(uri.clone(), credentials.clone()).await;
             while statements.is_err() {
-                tracing::error!("Failed to create statements, retrying");
+                tracing::error!(
+                    "Failed to connect to ScyllaDB (error: {}) at {}, retrying in 1s",
```
QuerthDP, line 223: Maybe we could move the constant to the timeout variable.

QuerthDP (approved):

QuerthDP (approved):

PR #241, opened 2025-09-18
Add vector similarity functions integration tests

QuerthDP:
This patch introduce the validator integration tests for the `vector_similarity()` function added in ScyllaDB.

Refs: scylladb/scylladb#25993

### DIFF
--- a/crates/validator/src/tests/crud.rs
+++ b/crates/validator/src/tests/crud.rs
@@ -38,7 +38,7 @@ async fn simple_create_drop_index(actors: TestActors) {
     )
     .await;
 
-    let index = create_index(&session, &client, &table, "embedding").await;
+    let index = create_index(&session, &client, &table, "embedding", None).await;
 
     assert_eq!(index.keyspace.as_ref(), &keyspace);
 
@@ -71,7 +71,7 @@ async fn simple_create_drop_multiple_indexes(actors: TestActors) {
     .await;
 
     // Create index on column v1
-    let index1 = create_index(&session, &client, &table, "v1").await;
+    let index1 = create_index(&session, &client, &table, "v1", None).await;
 
     // Wait for the full scan to complete and check if ANN query succeeds on v1
     wait_for(
@@ -99,7 +99,7 @@ async fn simple_create_drop_multiple_indexes(actors: TestActors) {
         .expect_err("ANN query should fail when index does not exist");
 
     // Create index on column v2
-    let index2 = create_index(&session, &client, &table, "v2").await;
+    let index2 = create_index(&session, &client, &table, "v2", None).await;
 
     // Check if ANN query on v1 still succeeds
     session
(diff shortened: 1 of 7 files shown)


crates/validator/src/tests/vector_similarity.rs
```diff
@@ -0,0 +1,212 @@
...
+        session
+            .query_unpaged(
+                format!("INSERT INTO {table} (pk, ck, v) VALUES (?, ?, ?)"),
+                (123, i as i32, &embedding.as_slice()),
+            )
+            .await
+            .expect("failed to insert data");
+    }
+
+    // Create index with EUCLIDEAN similarity function
```
knowack1, line 132: IMO this comment is not needed - code expresses this clearly.

crates/validator/src/tests/vector_similarity.rs
```diff
@@ -0,0 +1,212 @@
...
+        .with_test(
+            "vector_similarity_function_with_multi_column_partition_key",
+            timeout,
+            vector_similarity_function_with_multi_column_partition_key,
+        )
+}
+
+pub(crate) static EMBEDDINGS: [[f32; 3]; 3] = [[1.0, 2.0, 3.0], [4.0, 5.0, 6.0], [7.0, 8.0, 9.0]];
+
+async fn check_similarity_function_results(session: &Session, table: &str, key_column: &str) {
```
knowack1, line 36: I would call this function `assert_similarity_function_results, to be more specific of what it does.

crates/validator/src/tests/crud.rs
```diff
@@ -38,7 +38,7 @@ async fn simple_create_drop_index(actors: TestActors) {
     )
     .await;
 
-    let index = create_index(&session, &client, &table, "embedding").await;
+    let index = create_index(&session, &client, &table, "embedding", None).await;
```
knowack1, line 41: is it possible to avoid `None` in this call?

QuerthDP: That would require creating a separate function for example `create_index_with_options`.
However, the same approach with the option parameter is done already with `create_table`.

knowack1: Ok, If we have same approach already for create_table then let's keep the convention.

crates/validator/src/tests/vector_similarity.rs
```diff
@@ -0,0 +1,209 @@
...
+        let (key, distance) = row;
+        assert_eq!(
+            (key, distance),
+            expected_distances[i],
+            "Row {i} does not match expected result"
+        );
+    }
+}
+
+async fn vector_similarity_function_returns_expected_results(actors: TestActors) {
```
knowack1, line 60: Consider to rename to `vector_similarity_function_with_single_column_partition_key` to follow the pattern from consecutive  tests.

knowack1 (dismissed):

knowack1 (approved):

PR #242, opened 2025-09-18
Add support for building on both amd64 and arm64

ewienik:
`run-with-release-toolchain` script will now set a new CARGO_TARGET_DIR for
specific target architecture. Default target arch is amd64, you can change it
by setting TARGETARCH environment.

For multi-platform build scripts use docker multi-platform feature. You need to
prepare your environment according to
https://docs.docker.com/build/building/multi-platform/. There is additional
script `prepare-docker-qemu` to prepare using qemu in docker.

`build-release` now depends on docker multi-platform feature. It builds two
tar.gz and one multi-platform docker.

First patch moves all supporting scripts into the one directory

Fixes: VECTOR-236


scripts/prepare-docker-qemu
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+docker run --privileged --rm tonistiigi/binfmt --install all
```
swasik, line 3: Worth adding a comment what this cryptic tonistiigi/binfmt refers to. We can save 1 Google Search query per reader.

scripts/build-release
swasik, line 1: It would be great to add some short .md file with tutorial what to do when we want to release new version of vector store including all the code snippets. Could be in separate PR.

swasik (approved):

scripts/build-release
```diff
@@ -3,11 +3,12 @@
 set -e
 
 repo=$(dirname "${BASH_SOURCE[0]}")
-repo=$(realpath $repo)
+repo=$(realpath $repo/..)
 
 cd $repo
 
 vector_store_version=$(./run-with-release-toolchain cargo info vector-store | grep ^version: | cut -d ' ' -f 2)
```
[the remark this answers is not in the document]
ewienik: It is already done in the current version

scripts/build-release
ewienik: done

scripts/prepare-docker-qemu
ewienik: done

knowack1 (approved):

scripts/build-release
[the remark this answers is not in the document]
QuerthDP: I'm still seeing it as duplicated after the first commit. But let it be.

PR #247, opened 2025-09-23
validator/tests: add `test_` prefix to all tests

QuerthDP:
We need to be able to simply differ the test cases from the utility
functions implemented in a test file.
This patch introduces a conversion that each test should be named
with the `test_` prefix.
The test name should be the same as test function name but without
the `test_` prefix, as this would duplicate the test keywords in logs.

From this commit, let's keep the conversion for all future tests.

### DIFF
--- a/crates/validator/src/common.rs
+++ b/crates/validator/src/common.rs
@@ -151,13 +151,20 @@ pub(crate) async fn create_index(
     client: &HttpClient,
     table: &str,
     column: &str,
+    options: Option<&str>,
 ) -> IndexInfo {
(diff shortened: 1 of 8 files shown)


QuerthDP:
Changelog:
- rebased on #241

knowack1:
@QuerthDP consider closing this draft PR, as is idle for long time.

QuerthDP:
Since the changes to e2e-test framework, the tests are printed with good visibility since there's no need for this change.

The only pros of this change would be to clearly separate tests from utils in test files, but there's no way to enforce that for future tests so let's close this one.

PR #248, opened 2025-09-24
db: fix option parsing

smoczy123:
We allowed lowercase similarity function names in Scylla, however we parsed them wrong in Vector Store. This patch makes it so it is converted to uppercase before parsing.

Fixes: VECTOR-235

### DIFF
--- a/crates/vector-store/src/db.rs
+++ b/crates/vector-store/src/db.rs
@@ -518,7 +518,7 @@ impl Statements {
                 .unwrap_or_default();
             let space_type = options
                 .remove("similarity_function")
-                .and_then(|s| s.parse().ok())
+                .and_then(|s| s.to_ascii_uppercase().parse().ok())
                 .unwrap_or_default();
             (connectivity, expansion_add, expansion_search, space_type)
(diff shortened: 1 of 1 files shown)


crates/vector-store/src/db.rs
```diff
@@ -518,7 +518,7 @@ impl Statements {
                 .unwrap_or_default();
             let space_type = options
                 .remove("similarity_function")
-                .and_then(|s| s.parse().ok())
+                .and_then(|s| s.to_ascii_uppercase().parse().ok())
```
knowack1, line 521: Can we please add test case for this problem?

knowack1 (changes requested):

crates/vector-store/src/db.rs
knowack1: Better would be to validate if proper distance is calculated. Could you please create an issue for adding test case?

smoczy123: Created: https://scylladb.atlassian.net/browse/VECTOR-243.

PR #252, opened 2025-10-02
[RFC] This patch adds a perleminary work for cluster installation, configuration and running

amnonh:
This is preliminary work, don't merge!


ewienik:
Consider moving  directory into `scripts/cluster_test`

swasik:
I am converting it to draft. I think it can still be very useful - I am working on the plan for benchmarking VS and we may reuse big parts of it, but we need requirements and implementation plan first.

PR #372, opened 2026-02-26
vector-store: make index updates faster with additional CDC listener

QuerthDP:
## Motivation

The current single CDC reader is insufficient because new vectors can be delayed up to 40 seconds (30s safety interval + 10s sleep interval). The limits are set based on the default CDC driver intervals. This conflicts with our goal of real-time AI support, where vectors should be available to the Vector Store service immediately after writing.

To address this, we propose a trade-off: add a CDC reader with shorter intervals (e.g., sub-second safety and sleep), reducing end-to-end latency but increasing network and system load.

As this may cause data loss in case of cluster overload, we retain the existing CDC reader with longer intervals as a correctness backup to ensure no vectors are missed.

## Implementation

Refactor CDC reader into a new `db_cdc` actor.
This allows us to manage the CDC reader(s) with different configuration of safety and sleep intervals.
Having it in separate module makes a logical split, which clarifies the design.

## CDC readers default parameters

**Wide-framed CDC reader:**
- Safety interval: 30 s 
- Sleep interval: 10 s 
- Purpose: Provides consistent, reliable updates

**Fine-grained CDC reader:**
- Safety interval: 100 ms
- Sleep interval: 500 ms 
- Purpose: Provides low-latency updates

---

References: [Design Document](https://scylladb.atlassian.net/wiki/spaces/RND/pages/256344351/Vector+Store+Real-Time+CDC+Reader+Design+Document)

Fixes: VECTOR-485

### DIFF
--- a/crates/vector-store/src/db_cdc.rs
+++ b/crates/vector-store/src/db_cdc.rs
@@ -1,0 +1,435 @@
+/*
+ * Copyright 2026-present ScyllaDB
+ * SPDX-License-Identifier: LicenseRef-ScyllaDB-Source-Available-1.0
+ */
+
+use crate::AsyncInProgress;
+use crate::ColumnName;
+use crate::Config;
+use crate::DbEmbedding;
+use crate::IndexMetadata;
+use crate::internals::Internals;
+use crate::internals::InternalsExt;
+use ::time::Date;
+use ::time::Month;
+use ::time::OffsetDateTime;
+use ::time::PrimitiveDateTime;
+use ::time::Time;
+use anyhow::Context;
+use anyhow::anyhow;
+use anyhow::bail;
+use async_trait::async_trait;
+use scylla::client::session::Session;
+use scylla::value::CqlValue;
+use scylla_cdc::consumer::CDCRow;
+use scylla_cdc::consumer::Consumer;
+use scylla_cdc::consumer::ConsumerFactory;
+use scylla_cdc::log_reader::CDCLogReaderBuilder;
+use std::pin::pin;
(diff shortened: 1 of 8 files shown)


QuerthDP:
Changelog:
- update core commits commit messages
- update PR cover letter

crates/vector-store/src/db_cdc.rs
```diff
@@ -61,11 +103,10 @@ pub(crate) fn new(
...
 
-            let cdc_now = || SystemTime::now().duration_since(SystemTime::UNIX_EPOCH).unwrap();
-            let mut cdc_start = cdc_now();
-            let mut cdc_reader: Option<scylla_cdc::log_reader::CDCLogReader> = None;
-            let mut cdc_handler_task: Option<tokio::task::JoinHandle<Duration>> = None;
-            let shutdown_notify = Arc::new(Notify::new());
+            let mut readers = vec![
+                CdcReaderState::new("wide", CdcReaderParams::wide),
+                CdcReaderState::new("fine", CdcReaderParams::fine),
+            ];
```
ewienik: After refactoring db_cdc into an actor creating two cdc actors should be done in the caller of the db_cdc::new, not inside. db_cdc actor should be responsible for only one cdc process.

ewienik (changes requested):
The general advise is to create a real actor from db_cdc, remove dcd_error_notify (use actor channel instead) and use a db_cdc actor for single cdc process.

ewienik (changes requested):

ewienik (approved):

PR #406, opened 2026-04-02
test: add LWT CDC visibility tests to validator

m-szymon:
Verify that LWT writes (INSERT IF NOT EXISTS, UPDATE IF EXISTS, DELETE IF EXISTS) on vector-indexed tables are picked up by the fine-grained CDC reader and reflected in the index within the expected latency budget.

Reproduces SCYLLADB-1342

### DIFF
--- a/crates/validator-vector-store/src/cdc.rs
+++ b/crates/validator-vector-store/src/cdc.rs
@@ -10,6 +10,7 @@ use vector_search_validator_tests::common::*;
 use vector_search_validator_tests::*;
 
 const FINE_GRAINED_CDC_MAX_LATENCY: Duration = Duration::from_secs(2);
+const CDC_MAX_LATENCY: Duration = Duration::from_secs(60);
 
 #[framed]
 pub(crate) async fn new() -> TestCase {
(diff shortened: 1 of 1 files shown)


m-szymon:
These tests fail - they require fix in scylla: https://github.com/scylladb/scylladb/pull/29300

I added separate tests, but maybe one LWT test would be better? Or merging these conditional cases into existing tests?

I can't request review. @ewienik @swasik @knowack1 @QuerthDP @smoczy123

crates/validator-vector-store/src/cdc.rs
```diff
@@ -223,3 +238,193 @@ async fn cdc_delete_visible_immediately(actors: TestActors) {
...
+    // LWT insert - should be picked up by the fine-grained CDC reader
+    session
+        .query_unpaged(
+            format!("INSERT INTO {table} (pk, v) VALUES (1, [1.0, 2.0, 3.0]) IF NOT EXISTS"),
+            (),
+        )
+        .await
+        .expect("failed to insert data");
+
+    // Should timeout only when using wide-framed CDC reader
```
QuerthDP: I think those tests should not care whether the data is read from fine or wide reader.
We could timeout on the wide reader read I think.

QuerthDP (commented):
The tests itself look good.
My only remark is that we should not care which CDC is used to read the LWT.
Those tests should be transparent to that, as they were added to test the LWT transactions, not the fine grained CDC reader with them.

m-szymon:
I removed "fine-grained CDC reader" from LWT tests and rebased.

QuerthDP (approved):

PR #408, opened 2026-04-08
validator: add exact match test filer

QuerthDP:
Recognize double-quoted file and test segments in Validator filters as exact matches while keeping the existing partial matching for unquoted filters.

Preserve quoted filters in the Validator wrapper scripts so the new syntax survives shell forwarding in both docker and unshare runs.

Update the Validator workflows to quote matrix testcase names before appending :: so CI selects a single Validator test case instead of every case whose name merely contains the same substring.

Fixes: VECTOR-607

### DIFF
--- a/crates/validator-engine/src/lib.rs
+++ b/crates/validator-engine/src/lib.rs
@@ -95,6 +95,9 @@ enum Command {
         /// Filters to select specific tests to run.
         /// The syntax is as follows:
         ///     `<partially_matching_test_file_name>::<partially_matching_test_case_name>`
+        /// Wrap either side in double quotes to require an exact match, for example:
+        ///     `"crud"::`
+        ///     `::"simple_create"`
         /// Without specifying `::`, the filter will try to match both the file and test names.
         #[arg(value_name = "FILTER")]
         filters: Vec<String>,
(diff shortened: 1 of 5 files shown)


crates/validator-engine/src/lib.rs
```diff
@@ -95,6 +95,9 @@ enum Command {
         /// Filters to select specific tests to run.
         /// The syntax is as follows:
         ///     `<partially_matching_test_file_name>::<partially_matching_test_case_name>`
+        /// Wrap either side in double quotes to require an exact match, for example:
+        ///     `"crud"::`
+        ///     `::"simple_create"`
         /// Without specifying `::`, the filter will try to match both the file and test names.
```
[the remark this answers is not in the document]
QuerthDP: It does render properly

scripts/run-validator-with-scylla-unshare
```diff
@@ -18,22 +18,24 @@ validator=$(realpath $3)
...
-mount --bind $tmp_resolv_conf /etc/resolv.conf
+mount --bind "$tmp_resolv_conf" /etc/resolv.conf
 ip link set lo up
 ip addr add $dns_ip/32 dev lo
 for i in {1..10}; do
     ip addr add $base_ip.\$i/32 dev lo
 done
 cat /etc/resolv.conf
-$validator run --dns-ip $dns_ip --base-ip $base_ip.1 $scylla $vector_store $@
+$validator run --dns-ip $dns_ip --base-ip $base_ip.1 $scylla $vector_store$validator_args
```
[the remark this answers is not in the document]
QuerthDP: Changed

QuerthDP:
Changelog:
- refactor the shell script to properly quote the arguments

scripts/run-validator-with-scylla-docker
```diff
@@ -31,5 +31,5 @@ docker run --rm \
     --volume="$validator:/vector-search-validator" \
     --network=none \
     --entrypoint=/vector-search-validator \
-    $scylla_tag \
-    run --dns-ip $dns_ip --base-ip $base_ip --scylla-default-conf /etc/scylla/scylla.yaml /usr/bin/scylla /vector-store $@
+    "$scylla_tag" \
+    run --dns-ip "$dns_ip" --base-ip "$base_ip" --scylla-default-conf /etc/scylla/scylla.yaml /usr/bin/scylla /vector-store "$@"
```
ewienik: Is this refactoring for scylla_tag, dns_ip and base_ip needed?

scripts/run-validator-with-scylla-unshare
```diff
@@ -22,18 +22,33 @@ dns_ip=127.0.1.1
...
+    "$scylla" \
+    "$vector_store" \
+    "$@" <<'EOF'
+tmp_resolv_conf=$1
+dns_ip=$2
+base_ip=$3
+validator=$4
+scylla=$5
+vector_store=$6
+shift 6
```
ewienik: Is this refactoring of passing arguments really needed?

ewienik (changes requested):

scripts/run-validator-with-scylla-unshare
QuerthDP: Not really, if we expect that the parameters will be passed properly and does not need quoting then we shouldn't mind

scripts/run-validator-with-scylla-docker
QuerthDP: If we suppose those won't need quoting then don't. And I think we can suppose that

QuerthDP:
Changelog:

- reverted the refactoring

ewienik (approved):

PR #409, opened 2026-04-08
vector-store: fix blob primary key round-trip in CQL path

m-szymon:
Blob is a valid CQL primary key type. Previously, try_to_json hit unimplemented!() for CqlValue::Blob (crashing the server when a blob key appeared in an ANN result), and try_from_json had no NativeType::Blob arm (failing any ANN filter restriction on a blob-keyed column).

- try_to_json: serialise CqlValue::Blob as a '0x'-prefixed lowercase hex string, consistent with how CQL JSON represents blob values.
- try_from_json: parse '0x'-prefixed hex strings back into CqlValue::Blob so blob keys round-trip through the VS HTTP layer.
- Add unit tests covering both directions including empty blobs, single bytes, multi-byte values, and error cases (missing prefix, invalid hex, odd-length hex).
- Add blob to the serde validator test matrix, exercising the full CQL blob-key lifecycle: table creation, full-scan ingestion into InvariantKey, and ANN query results serialized via try_to_json.

Fixes: VECTOR-608

### DIFF
--- a/crates/vector-store/src/httproutes.rs
+++ b/crates/vector-store/src/httproutes.rs
@@ -867,6 +867,8 @@ fn try_to_json(value: CqlValue) -> anyhow::Result<Value> {
             })?,
         )),
 
+        CqlValue::Blob(value) => Ok(Value::String(const_hex::encode_prefixed(&value))),
+
         _ => unimplemented!(),
     }
 }
@@ -919,6 +921,14 @@ fn try_from_json(value: Value, cql_type: &NativeType) -> anyhow::Result<CqlValue
                 })
                 .map_err(|err| anyhow!("Failed to parse Timestamp from string '{value}': {err}"))?;
                 Ok(CqlValue::Timestamp(datetime.into()))
+            }
+            NativeType::Blob => {
+                if !value.starts_with("0x") {
+                    bail!("Blob value must be a '0x'-prefixed hex string");
+                }
+                let bytes = const_hex::decode(&value)
+                    .map_err(|err| anyhow!("Invalid hex in blob value: {err}"))?;
+                Ok(CqlValue::Blob(bytes))
             }
             _ => bail!("Cannot convert string to CqlValue::{cql_type:?}, unsupported type"),
         },
@@ -1547,6 +1557,26 @@ mod tests {
             .unwrap(),
             CqlValue::Timestamp(OffsetDateTime::from_unix_timestamp(64).unwrap().into())
         );
+
+        assert_eq!(
+            try_from_json(Value::String("0xdeadbeef".to_string()), &NativeType::Blob).unwrap(),
+            CqlValue::Blob(vec![0xde, 0xad, 0xbe, 0xef])
+        );
+        assert_eq!(
+            try_from_json(Value::String("0x".to_string()), &NativeType::Blob).unwrap(),
+            CqlValue::Blob(vec![])
+        );
+        assert_eq!(
+            try_from_json(Value::String("0x00".to_string()), &NativeType::Blob).unwrap(),
+            CqlValue::Blob(vec![0x00])
+        );
+
+        // missing 0x prefix
+        assert!(try_from_json(Value::String("deadbeef".to_string()), &NativeType::Blob).is_err());
+        // invalid hex characters
+        assert!(try_from_json(Value::String("0xgg".to_string()), &NativeType::Blob).is_err());
+        // odd-length hex digits (after stripping prefix)
+        assert!(try_from_json(Value::String("0xabc".to_string()), &NativeType::Blob).is_err());
     }
 
     #[test]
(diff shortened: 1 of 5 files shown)


m-szymon:
@QuerthDP please review.

crates/vector-store/src/httproutes.rs
```diff
@@ -867,6 +867,16 @@ fn try_to_json(value: CqlValue) -> anyhow::Result<Value> {
             })?,
         )),
 
+        CqlValue::Blob(value) => {
+            use std::fmt::Write;
```
QuerthDP: I'd import it in the top of the file

crates/vector-store/src/httproutes.rs
```diff
@@ -920,6 +930,23 @@ fn try_from_json(value: Value, cql_type: &NativeType) -> anyhow::Result<CqlValue
                 .map_err(|err| anyhow!("Failed to parse Timestamp from string '{value}': {err}"))?;
                 Ok(CqlValue::Timestamp(datetime.into()))
             }
+            NativeType::Blob => {
+                let hex = value
+                    .strip_prefix("0x")
+                    .ok_or_else(|| anyhow!("Blob value must be a '0x'-prefixed hex string"))?;
```
QuerthDP: Could use `.ok_or(anyhow!(...)` instead

crates/vector-store/src/httproutes.rs
```diff
@@ -920,6 +930,23 @@ fn try_from_json(value: Value, cql_type: &NativeType) -> anyhow::Result<CqlValue
...
+                let bytes = (0..hex.len())
+                    .step_by(2)
+                    .map(|i| {
+                        u8::from_str_radix(
+                            hex.get(i..i + 2)
+                                .ok_or_else(|| anyhow!("Blob hex string has odd length"))?,
```
QuerthDP: ditto

crates/vector-store/src/httproutes.rs
```diff
@@ -1639,6 +1666,42 @@ mod tests {
...
         assert!(try_to_json(CqlValue::Float(f32::NAN)).is_err());
         assert!(try_to_json(CqlValue::Double(f64::NAN)).is_err());
+
+        assert_eq!(
+            try_to_json(CqlValue::Blob(vec![0xde, 0xad, 0xbe, 0xef])).unwrap(),
+            Value::String("0xdeadbeef".to_string())
+        );
+        assert_eq!(
+            try_to_json(CqlValue::Blob(vec![])).unwrap(),
+            Value::String("0x".to_string())
```
QuerthDP, line 1679: Should it be 0x or just empty? Why?

QuerthDP (dismissed):
Looks fine. Leaving some nitpicks and a question

QuerthDP:
This could be opened right? Why keep it in draft?

crates/vector-store/src/httproutes.rs
m-szymon: Scylla validates the 0x prefix (throws "Blob JSON strings must start with 0x" if missing)

QuerthDP (dismissed):

QuerthDP:
Needs rebase for cargo deny to pass

crates/vector-store/src/httproutes.rs
```diff
@@ -867,6 +868,15 @@ fn try_to_json(value: CqlValue) -> anyhow::Result<Value> {
             })?,
         )),
 
+        CqlValue::Blob(value) => {
+            let mut hex = String::with_capacity(2 + value.len() * 2);
+            hex.push_str("0x");
+            for b in &value {
+                write!(hex, "{b:02x}").expect("writing to String never fails");
+            }
```
ewienik: consider using https://docs.rs/const-hex/latest/const_hex/fn.encode_prefixed.html

crates/vector-store/src/httproutes.rs
```diff
@@ -920,6 +930,23 @@ fn try_from_json(value: Value, cql_type: &NativeType) -> anyhow::Result<CqlValue
...
+                    .step_by(2)
+                    .map(|i| {
+                        u8::from_str_radix(
+                            hex.get(i..i + 2)
+                                .ok_or(anyhow!("Blob hex string has odd length"))?,
+                            16,
+                        )
+                        .map_err(|err| anyhow!("Invalid hex byte in blob value: {err}"))
+                    })
+                    .collect::<anyhow::Result<Vec<u8>>>()?;
```
ewienik: consider using https://docs.rs/const-hex/latest/const_hex/fn.decode.html

ewienik (changes requested):

crates/vector-store/src/httproutes.rs
m-szymon: ok.
scylla-cdc internally uses hex crate - it is already in dependency tree.
I went with const-hex suggestion as it reportedly has much better performance. However sending large blobs is not a recommended scenario anyway (and probably not very useful), so maybe it would be cleaner to use hex crate?

crates/vector-store/src/httproutes.rs
m-szymon: ok.
BTW, I am not sure if we need to be so strict about accepting encoded blobs. When Scylla reads it, it requires 0x prefix. Theoretically we don't have to check it (and const_hex::decode handles it.

m-szymon (commented):
Updated and rebased.

Could you check what is going on with failed (unrelated to this patch) test?

crates/vector-store/src/httproutes.rs
ewienik: In the maintenance period we could reach different versions of the same crates, so as scylla-cdc is in different repository and project I don't think we should try to use the same dependency.

crates/vector-store/src/httproutes.rs
ewienik: Let's be strict with checking for now.

ewienik (approved):
There are failed tests

PR #413, opened 2026-04-13
vector-store: add index routing

QuerthDP:
# Motivation

This PR implements index routing in Vector Store, following the changes in ScyllaDB (https://github.com/scylladb/scylladb/pull/29407) allowing for multiple indexes on the same column.

# Solution

When a replacement index is created for the same keyspace, table, and target column, ANN requests can hit the new index while it is still bootstrapping. Before this change, those requests returned `503 Service Unavailable` even if another compatible index in the same routing group was already fully serving.

This change adds routing so ANN requests are served by the best matching index that has completed its full scan. That keeps queries available during index replacement while preserving the existing `503` behavior when no safe fallback exists.

## Index selection algorithm

When an ANN request arrives, the system finds all indexes that cover the same keyspace, table, and target column and picks the best one:

1. **Eliminate ineligible indexes.** A candidate is dropped if any of the query's restriction columns (equality or range) are not present in the index's filtering columns (which include the table's primary key columns). A local index is additionally dropped if its partition key columns are not all present in the query's equality restrictions.

2. **Score the remaining candidates.** Each index is scored by how many restriction columns still need server-side filtering. For local indexes, partition-key columns covered by equality restrictions are not counted. An index that covers every restriction column scores `NeedsFiltering::No`; otherwise, `NeedsFiltering::Yes(n)` where fewer uncovered columns is better.

3. **Require the index to be serving.** Only indexes whose `IndexStatus` is `Serving` are considered.

4. **Break ties by version.** Among candidates with the same score, the one with the newest `IndexVersion` (created most recently) wins.

If the selected index needs filtering (`NeedsFiltering::Yes`) and the query does not include `ALLOW FILTERING`, the request returns `400 Bad Request`.

If no serving candidate exists and the originally requested index is still building, the request returns `503 Service Unavailable` with a progress indicator. If the full scan already finished but the index is not serving, it returns `500 Internal Server Error`. If the requested index does not exist at all, the request returns `404 Not Found`.

## What changed
- add `indexes` module with `Indexes`, `RoutingMap`, `IndexEntry`, `score_index`, and `route_index`
- group indexes by keyspace, table, and target column via `RoutingGroupKey`
- add `Engine::GetBestIndex` message and `EngineExt::get_best_index` to route ANN queries through the engine actor
- score serving candidates using the ANN query's equality and range restriction columns
- prefer local indexes when partition key restrictions match, prefer indexes with fewer uncovered restriction columns, and break ties by newest `IndexVersion`
- use the routed index key for ANN execution (`ann` and `filtered_ann` calls)
- status and metrics endpoints continue using `get_index` (direct lookup, no routing)
- return `400 Bad Request` when a routed index needs filtering but the query omits `ALLOW FILTERING`
- make `IndexVersion` orderable so tie-breaking is deterministic
- add `slow-test-hooks` feature flag with an `ann-served-request` counter to verify which index actually served an ANN request
- tighten integration test waits to block on `Serving` status rather than simple index discovery
- update full scan progress simulation in `DbBasic` to report `InProgress(0%)` until the scan finishes
- add `ALLOW FILTERING` to all validator filtering test queries that include restriction columns

## Testing
- add integration tests for index scoring (`score_index`), local-vs-global routing, partition key precision, and filtering column match ranking
- add validator end-to-end tests that ANN requests to a bootstrapping replacement index are served by the newest best matching serving index
- add validator coverage that routing does not cross target columns and still returns `503` until the requested index is serving
- add validator coverage for not-found, unavailable-while-bootstrapping, and dropped-index edge cases

Fixes: VECTOR-640

### DIFF
--- a/crates/vector-store/src/engine.rs
+++ b/crates/vector-store/src/engine.rs
@@ -3,6 +3,7 @@
  * SPDX-License-Identifier: LicenseRef-ScyllaDB-Source-Available-1.0
  */
 
+use crate::ColumnName;
 use crate::Config;
 use crate::DbIndexType;
 use crate::IndexKey;
@@ -16,14 +17,17 @@ use crate::db_index::DbIndexExt;
 use crate::factory::IndexFactory;
 use crate::index::Index;
 use crate::index::factory::IndexConfiguration;
+use crate::indexes;
+use crate::indexes::BestIndexState;
+use crate::indexes::IndexEntry;
+use crate::indexes::Indexes;
+use crate::indexes::RoutingMap;
 use crate::memory;
 use crate::memory::Memory;
 use crate::monitor_indexes;
 use crate::monitor_items;
-use crate::monitor_items::MonitorItems;
 use crate::node_state::NodeState;
 use crate::table::Table;
-use std::collections::HashMap;
 use std::sync::Arc;
 use std::sync::RwLock;
 use tokio::sync::mpsc;
@@ -39,6 +43,7 @@ use tracing::trace;
 type GetIndexKeysR = Vec<(IndexKey, Quantization)>;
 type AddIndexR = anyhow::Result<()>;
 type GetIndexR = Option<(mpsc::Sender<Index>, mpsc::Sender<DbIndex>)>;
+type GetBestIndexR = BestIndexState;
 
 pub(crate) enum Engine {
     GetIndexIds {
@@ -55,6 +60,12 @@ pub(crate) enum Engine {
         key: IndexKey,
         tx: oneshot::Sender<GetIndexR>,
     },
+    GetBestIndex {
+        key: IndexKey,
+        equality_columns: Vec<ColumnName>,
+        range_columns: Vec<ColumnName>,
+        tx: oneshot::Sender<GetBestIndexR>,
+    },
 }
 
 pub(crate) trait EngineExt {
@@ -62,6 +73,12 @@ pub(crate) trait EngineExt {
     async fn add_index(&self, metadata: IndexMetadata) -> AddIndexR;
     async fn del_index(&self, key: IndexKey);
     async fn get_index(&self, key: IndexKey) -> GetIndexR;
+    async fn get_best_index(
+        &self,
+        key: IndexKey,
+        equality_columns: Vec<ColumnName>,
+        range_columns: Vec<ColumnName>,
+    ) -> GetBestIndexR;
 }
 
 impl EngineExt for mpsc::Sender<Engine> {
@@ -97,17 +114,26 @@ impl EngineExt for mpsc::Sender<Engine> {
         rx.await
             .expect("EngineExt::get_index: internal actor should send response")
     }
-}
 
-type IndexesT = HashMap<
-    IndexKey,
-    (
-        mpsc::Sender<Index>,
-        mpsc::Sender<MonitorItems>,
-        mpsc::Sender<DbIndex>,
-        Quantization,
-    ),
->;
+    async fn get_best_index(
+        &self,
+        key: IndexKey,
+        equality_columns: Vec<ColumnName>,
+        range_columns: Vec<ColumnName>,
+    ) -> GetBestIndexR {
+        let (tx, rx) = oneshot::channel();
+        self.send(Engine::GetBestIndex {
+            key,
+            equality_columns,
+            range_columns,
+            tx,
+        })
+        .await
+        .expect("EngineExt::get_best_index: internal actor should receive request");
+        rx.await
+            .expect("EngineExt::get_best_index: internal actor should send response")
+    }
+}
 
 pub(crate) async fn new(
     db: mpsc::Sender<Db>,
@@ -118,15 +144,21 @@ pub(crate) async fn new(
 ) -> anyhow::Result<mpsc::Sender<Engine>> {
     let (tx, mut rx) = mpsc::channel(10);
 
-    let monitor_actor =
-        monitor_indexes::new(db.clone(), tx.downgrade(), node_state, config_rx.clone()).await?;
+    let monitor_actor = monitor_indexes::new(
+        db.clone(),
+        tx.downgrade(),
+        node_state.clone(),
+        config_rx.clone(),
+    )
+    .await?;
     let memory_actor = memory::new(config_rx);
 
     tokio::spawn(
         async move {
             debug!("starting");
 
-            let mut indexes: IndexesT = HashMap::new();
+            let mut indexes = Indexes::new();
+            let mut routing_map = RoutingMap::new();
             while let Some(msg) = rx.recv().await {
                 match msg {
                     Engine::GetIndexIds { tx } => get_index_keys(tx, &indexes).await,
@@ -138,15 +170,36 @@ pub(crate) async fn new(
                             &db,
                             index_factory.as_ref(),
                             &mut indexes,
+                            &mut routing_map,
                             metrics.clone(),
                             memory_actor.clone(),
                         )
                         .await
                     }
 
-                    Engine::DelIndex { key } => del_index(key, &mut indexes).await,
+                    Engine::DelIndex { key } => {
+                        del_index(key, &mut indexes, &mut routing_map).await
+                    }
 
                     Engine::GetIndex { key, tx } => get_index(key, tx, &indexes).await,
+
+                    Engine::GetBestIndex {
+                        key,
+                        equality_columns,
+                        range_columns,
+                        tx,
+                    } => {
+                        get_best_index(
+                            key,
+                            equality_columns,
+                            range_columns,
+                            tx,
+                            &indexes,
+                            &routing_map,
+                            &node_state,
+                        )
+                        .await
+                    }
                 }
             }
             drop(monitor_actor);
@@ -159,22 +212,24 @@ pub(crate) async fn new(
     Ok(tx)
 }
 
-async fn get_index_keys(tx: oneshot::Sender<GetIndexKeysR>, indexes: &IndexesT) {
+async fn get_index_keys(tx: oneshot::Sender<GetIndexKeysR>, indexes: &Indexes) {
     tx.send(
         indexes
             .iter()
-            .map(|(key, (_, _, _, quantization))| (key.clone(), *quantization))
+            .map(|(key, entry)| (key.clone(), entry.quantization))
             .collect(),
     )
     .unwrap_or_else(|_| trace!("Engine::GetIndexIds: unable to send response"));
 }
 
+#[allow(clippy::too_many_arguments)]
 async fn add_index(
     metadata: IndexMetadata,
     tx: oneshot::Sender<AddIndexR>,
     db: &mpsc::Sender<Db>,
     index_factory: &(dyn IndexFactory + Send + Sync),
-    indexes: &mut IndexesT,
+    indexes: &mut Indexes,
+    routing_map: &mut RoutingMap,
     metrics: Arc<Metrics>,
     memory: Sender<Memory>,
 ) {
@@ -198,13 +253,13 @@ async fn add_index(
 
     let primary_key_columns = db_index.get_primary_key_columns().await;
     let table_columns = db_index.get_table_columns().await;
-    let partition_key_columns = match metadata.index_type {
-        DbIndexType::Local(partition_key_columns) => Some(partition_key_columns),
+    let partition_key_columns = match &metadata.index_type {
+        DbIndexType::Local(partition_key_columns) => Some(Arc::clone(partition_key_columns)),
         DbIndexType::Global => None,
     };
     let table = match Table::new(
         key.clone(),
-        primary_key_columns,
+        primary_key_columns.clone(),
         partition_key_columns,
         &metadata.filtering_columns,
         table_columns,
@@ -260,27 +315,56 @@ async fn add_index(
         }
     };
 
-    indexes.insert(
-        key.clone(),
-        (index_actor, monitor_actor, db_index, metadata.quantization),
+    let index_entry = IndexEntry::new(
+        index_actor,
+        monitor_actor,
+        db_index,
+        primary_key_columns,
+        metadata,
     );
+
+    routing_map.add(index_entry.routing_group.clone(), key.clone());
+
+    indexes.insert(key.clone(), index_entry);
+
     info!("creating the index {key}");
     tx.send(Ok(()))
         .unwrap_or_else(|_| trace!("add_index: unable to send response"));
 }
 
-async fn del_index(key: IndexKey, indexes: &mut IndexesT) {
-    indexes.remove(&key);
+async fn del_index(key: IndexKey, indexes: &mut Indexes, routing_map: &mut RoutingMap) {
+    if let Some(entry) = indexes.remove(&key) {
+        routing_map.remove(entry.routing_group, &key);
+    }
     info!("removed the index {key}");
 }
 
-async fn get_index(key: IndexKey, tx: oneshot::Sender<GetIndexR>, indexes: &IndexesT) {
-    tx.send(
-        indexes
-            .get(&key)
-            .map(|(index, _, db_index, _)| (index.clone(), db_index.clone())),
+async fn get_index(key: IndexKey, tx: oneshot::Sender<GetIndexR>, indexes: &Indexes) {
+    let result = indexes::get_index(&key, indexes);
+    tx.send(result)
+        .unwrap_or_else(|_| trace!("get_index: unable to send response"));
+}
+
+async fn get_best_index(
+    key: IndexKey,
+    equality_columns: Vec<ColumnName>,
+    range_columns: Vec<ColumnName>,
+    tx: oneshot::Sender<GetBestIndexR>,
+    indexes: &Indexes,
+    routing_map: &RoutingMap,
+    node_state: &mpsc::Sender<NodeState>,
+) {
+    let result = indexes::route_index(
+        &key,
+        &equality_columns,
+        &range_columns,
+        indexes,
+        routing_map,
+        node_state,
     )
-    .unwrap_or_else(|_| trace!("get_index: unable to send response"));
+    .await;
+    tx.send(result)
+        .unwrap_or_else(|_| trace!("get_best_index: unable to send response"));
 }
 
 #[cfg(test)]
@@ -308,6 +392,14 @@ pub(crate) mod tests {
             key: IndexKey,
             tx: oneshot::Sender<GetIndexR>,
         ) -> impl Future<Output = ()> + Send + 'static;
+
+        fn get_best_index(
+            &self,
+            key: IndexKey,
+            equality_columns: Vec<ColumnName>,
+            range_columns: Vec<ColumnName>,
+            tx: oneshot::Sender<GetBestIndexR>,
+        ) -> impl Future<Output = ()> + Send + 'static;
     }
 
     pub(crate) fn new(sim: impl SimEngine + Send + 'static) -> mpsc::Sender<Engine> {
@@ -330,6 +422,15 @@ pub(crate) mod tests {
                         Engine::AddIndex { metadata, tx } => sim.add_index(metadata, tx).await,
                         Engine::DelIndex { key } => sim.del_index(key).await,
                         Engine::GetIndex { key, tx } => sim.get_index(key, tx).await,
+                        Engine::GetBestIndex {
+                            key,
+                            equality_columns,
+                            range_columns,
+                            tx,
+                        } => {
+                            sim.get_best_index(key, equality_columns, range_columns, tx)
+                                .await
+                        }
                     }
                 }
 
--- a/crates/vector-store/src/httproutes.rs
+++ b/crates/vector-store/src/httproutes.rs
@@ -15,6 +15,7 @@ use crate::engine::Engine;
 use crate::engine::EngineExt;
 use crate::index::IndexExt;
 use crate::index::validator;
+use crate::indexes;
 use crate::info::Info;
 use crate::internals::Internals;
 use crate::internals::InternalsExt;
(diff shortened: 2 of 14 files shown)


QuerthDP:
https://github.com/scylladb/scylladb/pull/29407 needs to be merged to enable this patch

QuerthDP:
Changelog:
- rebase on master

crates/validator-vector-store/src/routing.rs
```diff
@@ -0,0 +1,374 @@
+/*
+ * Copyright 2026-present ScyllaDB
+ * SPDX-License-Identifier: LicenseRef-ScyllaDB-Source-Available-1.0
+ */
+
+use async_backtrace::framed;
+use httpclient::HttpClient;
+use reqwest::StatusCode;
+use scylla_proxy::Condition;
+use scylla_proxy::Reaction;
```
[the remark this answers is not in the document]
QuerthDP: The tests does not compile without it. As you said `RequestReaction::delay` does need it

QuerthDP:
Changelog:
- move all routing to Vector Store side, meaning that both local and global indexes are checked to be the best match for ANN query based on the restrictions on the columns
- add comprehensive tests for best matching index scoring

crates/vector-store/src/engine.rs
```diff
@@ -259,27 +326,87 @@ async fn add_index(
...
+    }))
+    .await
+    .into_iter()
+    .flatten()
+    .max_by_key(|(_, version)| *version)
+    .map_or(key, |(k, _)| k);
+
+    let routed_entry = indexes.get(routed_key)?;
+    if routed_key != key {
+        info!("routing ANN query from {key} to {}", routed_key);
```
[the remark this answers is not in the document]
QuerthDP: Done

crates/vector-store/src/engine.rs
```diff
@@ -259,27 +326,87 @@ async fn add_index(
...
+        }
+        Some(async move {
+            matches!(entry.db_index.full_scan_progress().await, Progress::Done)
+                .then_some((k, &entry.version))
+        })
+    }))
+    .await
+    .into_iter()
+    .flatten()
+    .max_by_key(|(_, version)| *version)
```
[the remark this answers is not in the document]
QuerthDP: Changed

crates/vector-store/src/engine.rs
```diff
@@ -259,27 +326,87 @@ async fn add_index(
...
+        let entry = indexes.get(k)?;
+        if !is_same_index_type(&requested_entry.index_type, &entry.index_type) {
+            return None;
+        }
+        Some(async move {
+            matches!(entry.db_index.full_scan_progress().await, Progress::Done)
+                .then_some((k, &entry.version))
+        })
+    }))
+    .await
```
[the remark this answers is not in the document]
QuerthDP: This is more serious problem, out of scope for this PR. When `full_scan_progress` will be optimized, we will change it here as well.

QuerthDP:
Validator failed: VECTOR-651
Rerunning

crates/vector-store/src/lib.rs
```diff
@@ -588,6 +588,27 @@ pub struct Filter {
...
+
+impl PartialOrd for IndexVersion {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for IndexVersion {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        self.gregorian_ticks().cmp(&other.gregorian_ticks())
```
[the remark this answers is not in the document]
QuerthDP: We only care about the timestamp here. Additionally, the index version should always contain a timestamp as that's the way it's implemented on Scylla side. I'll leave it as is.

crates/validator-vector-store/src/routing.rs
```diff
@@ -0,0 +1,798 @@
...
+    .await;
+
+    wait_for_bootstrapping(client, &index).await;
+
+    let response = post_ann(client, &index).await;
+    assert_eq!(response.status(), StatusCode::SERVICE_UNAVAILABLE);
+
+    actors.db_proxy.turn_off_rules().await;
+    let index_status = common::wait_for_index(client, &index).await;
+    assert_eq!(index_status.count, 1);
```
knowack1: The same logic is already tested in `ann_does_not_route_between_columns_while_requested_index_is_bootstrapping`. I would finish the test at `assert_eq!(response.status(), StatusCode::SERVICE_UNAVAILABLE);`, because this is what we want to verify.

crates/validator-vector-store/src/routing.rs
```diff
@@ -0,0 +1,798 @@
...
+        &clients,
+        &table,
+        "embedding",
+    ))
+    .await;
+
+    wait_for_bootstrapping(client, &replacement).await;
+
+    let response =
+        assert_ann_query_served_by(client, &oldest, post_ann(client, &replacement)).await;
```
knowack1: Since these are integration tests, the validator tests should use ANN via CQL whenever possible.

knowack1 (commented):
Looks good. I expected the first commit to also contain some comprehensive engine unit tests for the non-trivial logic of `route_index` and `score_index`. However, the next commit already covers all use cases in the validator tests - this suite is very comprehensive and high quality. I like this very much. Please check my comments.

crates/validator-vector-store/src/routing.rs
QuerthDP: When using CQL SELECT we cannot directly choose which index are we sending the query directly to. Since I've added the `ann-served-request` counter which directly tells us which index was really queried, maybe we don't need to choose by hand. I'll change that where I think we can

QuerthDP: The only thing we lose by querying via CQL is the confidence that the query was routed properly, not just by blindly picking the correct index. I think it's quite interesting part of this tests, so I would prefer it to stay, but if you think and can convince me otherwise I may change all `post_ann`s to SELECTs. Maybe we should do that and add unit tests for engine?

knowack1: If I understand correctly we assert which index was targeted by means of internal counters - this must stay as is. If possible change only ANN HTTP request to CQL select.

QuerthDP: Consider the `ann_routes_to_newest_local_index_with_same_score` test case. In this scenario, both indexes are equally valid, but the expected behavior is to prioritize the most recent one.

Without `post_ann`, Scylla might naturally select the newest index by default. If the request were to land on that endpoint directly (`/ann/newer_local`), the test would pass even if the routing logic wasn't functioning correctly, as no actual re-routing would be required.

By utilizing `post_ann`, we can intentionally direct the request to the older index (`/ann/older_local`). This allows us to verify that the routing mechanism properly intercepts the request and redirects it to the newer index as intended

ewienik: As validator tests are e2e tests we should check behavior visible by the user, so mostly CQL. HTTP API we are using to check the index status (we might use cql for checking index status if this is possible). Tests for checking internal routing are also necessary. What about this approach: let's check routing by HTTP API in integration tests and leave CQL tests in validator?

crates/vector-store/src/engine.rs
```diff
@@ -89,25 +97,115 @@ impl EngineExt for mpsc::Sender<Engine> {
...
+/// Computes a routing score for an index given the query's restriction columns.
+///
+/// The score determines index preference:
+/// - A local index with at least one matching partition key column gets the
+///   highest priority (2). Among such indexes, more matching partition key
+///   columns are preferred, then more matching filtering columns.
+/// - A global index gets medium priority (1), scored by matching filtering
+///   columns.
+/// - A local index with no matching partition key columns gets the lowest
+///   priority (0).
```
ewienik: Local index with no matching partition is always not valid, should be returned at all.

crates/vector-store/src/engine.rs
```diff
@@ -89,25 +97,115 @@ impl EngineExt for mpsc::Sender<Engine> {
...
+            filtering_columns: metadata.filtering_columns,
+            version: metadata.version,
+            quantization: metadata.quantization,
+        }
+    }
+}
+
+type IndexesT = HashMap<IndexKey, IndexEntry>;
+
+type RoutingMapT = HashMap<RoutingGroupKey, Vec<IndexKey>>;
```
ewienik: Could these types be new structs and go into separate modules, ie. `indexes.rs`? This would separate functionality from engine.rs actor and for the future it facilitates sharing with httproutes directly.

crates/vector-store/src/engine.rs
```diff
@@ -89,25 +97,115 @@ impl EngineExt for mpsc::Sender<Engine> {
...
+///   priority (0).
+///
+/// This ensures queries with partition key restrictions prefer local indexes,
+/// queries without restrictions prefer global indexes, and richer matching
+/// indexes win ties when they can satisfy more request restrictions.
+fn score_index(
+    index_type: &DbIndexType,
+    filtering_columns: &[ColumnName],
+    restriction_columns: &[ColumnName],
+) -> (u8, usize, usize) {
```
ewienik: What about returning `Option<NeedsFiltering>`? Optional means the index is valid (could be used to search because all restriction columns are accessible by the index), and
```
enum NeedsFiltering{
No, // restriction columns are equal to index partition key columns
Yes(usize), // this number of restriction columns is not covered in partition key columns
}
```

crates/vector-store/src/engine.rs
```diff
@@ -259,27 +367,99 @@ async fn add_index(
...
+///
+/// 1. Returns `None` if the requested index key does not exist.
+/// 2. Identifies all candidate indexes within the same routing group
+///    (i.e., sharing the same keyspace, table, and target column).
+/// 3. Narrows down the candidates to those that are actively serving.
+/// 4. Scores each serving candidate.
+/// 5. Picks the candidate with the highest score, breaking ties by
+///    the newest `IndexVersion`.
+/// 6. If no candidate meets the criteria, safely falls back to the
+///    originally requested index.
```
ewienik: If there are no candidates function should return None

crates/vector-store/src/engine.rs
```diff
@@ -259,27 +367,99 @@ async fn add_index(
...
+            &entry.filtering_columns,
+            restriction_columns,
+        );
+        Some(async move {
+            matches!(entry.db_index.full_scan_progress().await, Progress::Done).then_some((
+                k,
+                score,
+                &entry.version,
+            ))
+        })
```
ewienik: We shouldn't check scan progress if score_index returns none.

crates/vector-store/src/engine.rs
```diff
@@ -259,27 +367,99 @@ async fn add_index(
...
+    }))
+    .await
+    .into_iter()
+    .flatten()
+    .max_by(|(_, score_a, version_a), (_, score_b, version_b)| {
+        score_a.cmp(score_b).then_with(|| version_a.cmp(version_b))
+    })
+    .map_or(key, |(k, _, _)| k);
```
ewienik: It should be none in case of not found.

crates/vector-store/src/httproutes.rs
```diff
@@ -435,6 +436,30 @@ pub enum PostIndexAnnRestriction {
...
+            | PostIndexAnnRestriction::Lt { lhs, .. }
+            | PostIndexAnnRestriction::Lte { lhs, .. }
+            | PostIndexAnnRestriction::Gt { lhs, .. }
+            | PostIndexAnnRestriction::Gte { lhs, .. } => vec![lhs.clone()],
+            PostIndexAnnRestriction::EqTuple { lhs, .. }
+            | PostIndexAnnRestriction::InTuple { lhs, .. }
+            | PostIndexAnnRestriction::LtTuple { lhs, .. }
+            | PostIndexAnnRestriction::LteTuple { lhs, .. }
+            | PostIndexAnnRestriction::GtTuple { lhs, .. }
+            | PostIndexAnnRestriction::GteTuple { lhs, .. } => lhs.clone(),
```
ewienik: We should distinguish between equality and range restrictions. Only equality restrictions should be checked for index partition key match, range restrictions must be checked with available filtering columns. Functionality for calculating score should be updated with this information.

crates/vector-store/src/httproutes.rs
ewienik: As full scan progress is checked during get_index, we don't need to repeat this check here.

ewienik (changes requested):

QuerthDP:
Changelog:
- move routing logic tests to integration

crates/validator-vector-store/src/routing.rs
QuerthDP: Changed

crates/validator-vector-store/src/routing.rs
QuerthDP: Moved the routing logic tests to integration leaving the user facing tests in Validator

swasik:
> prefer local indexes when partition key restrictions match, prefer more precise partition key matches and filtering column matches, and otherwise prefer global indexes
select the highest IndexVersion among equally good serving candidates

@QuerthDP could you describe the selection algorithm in a way that will not require knowledge of internal implementation to understand? It's a key change in this PR and should be clear how it works from the cover letter.

crates/vector-store/tests/integration/quantization.rs
```diff
@@ -292,6 +303,17 @@ async fn binary_quantization_with_non_divisible_by_8_dimensions() {
 
     let (client, _server, _config_tx) = run.await;
 
+    wait_for(
```
knowack1, line 316: Why do we need this wait for? We are polling on ann search, so this should still work without changes.

QuerthDP:
```
thread 'quantization::binary_quantization_with_non_divisible_by_8_dimensions' (1093204) panicked at crates/httpclient/src/lib.rs:76:14:
called `Result::unwrap()` on an `Err` value: reqwest::Error { kind: Decode, source: Error("expected value", line: 1, column: 1) }
```

QuerthDP: We need this since `ann()` does `post_ann().unwrap()` which fails with nonserving index

crates/vector-store/src/httproutes.rs
QuerthDP: `route_index()` does check full scan progress, but only while picking a routable serving candidate. Its return type is just Option, so once `httproutes.rs` gets None, it no longer knows whether that meant:
- the index does not exist,
- no candidate matched the query shape because `score_index` rejected them,
- the exact index exists but all candidates are still bootstrapping.

The extra direct lookup in `httproutes` is what preserves the 503 response for a still-building exact index.

QuerthDP:
Changelog:
- moved routing logic to `indexes.rs` file
- adjusted algorithm to always reject the local indexes with not matching partition keys

QuerthDP:
Changelog:
- satisfy cargo fmt

QuerthDP:
> > prefer local indexes when partition key restrictions match, prefer more precise partition key matches and filtering column matches, and otherwise prefer global indexes
> > select the highest IndexVersion among equally good serving candidates
> 
> @QuerthDP could you describe the selection algorithm in a way that will not require knowledge of internal implementation to understand? It's a key change in this PR and should be clear how it works from the cover letter.

Added the algorithm description

QuerthDP:
Changelog:
- update commit message to reflect the index is not routed in case of `/status` and `/metrics` endpoints

crates/vector-store/src/engine.rs
```diff
@@ -52,6 +55,12 @@ pub(crate) enum Engine {
         key: IndexKey,
     },
     GetIndex {
+        key: IndexKey,
+        equality_columns: Vec<ColumnName>,
+        range_columns: Vec<ColumnName>,
+        tx: oneshot::Sender<GetIndexR>,
+    },
+    GetIndexDirect {
```
ewienik: Consider leaving GetIndex as it is and add GetBestIndex as a message

crates/vector-store/src/engine.rs
```diff
@@ -38,7 +41,7 @@ use tracing::trace;
 
 type GetIndexKeysR = Vec<(IndexKey, Quantization)>;
 type AddIndexR = anyhow::Result<()>;
-type GetIndexR = Option<(mpsc::Sender<Index>, mpsc::Sender<DbIndex>)>;
+type GetIndexR = Option<(IndexKey, mpsc::Sender<Index>, mpsc::Sender<DbIndex>)>;
```
ewienik: Consider leaving GetIndexR as it it and add GetBestIndexR as a new alias.

crates/vector-store/src/httproutes.rs
```diff
@@ -572,24 +599,31 @@ async fn post_index_ann(
...
-    let Some((index, db_index)) = state.engine.get_index(index_key.clone()).await else {
+    let (equality_cols, range_cols) = restriction_columns(&request.filter);
+    let Some((routed_key, index, db_index)) = state
+        .engine
+        .get_index(index_key.clone(), equality_cols, range_cols)
+        .await
+    else {
         timer.observe_duration();
+
+        if let Some((_, _, db_index)) = state.engine.get_index_direct(index_key).await {
```
ewienik: engine.get_index (and maybe get_best_index) ought to know if index is serving. Consider refactoring return value from `get_index` as
```
enum BestIndexState {
None,
NotServing(db_index),
Serving(...),
}
```
In this case there is no need to next call to get_index_direct

crates/vector-store/src/indexes.rs
```diff
@@ -0,0 +1,290 @@
...
+    /// This many restriction columns are not covered by the partition key
+    /// or filtering columns and require server-side filtering.
+    Yes(usize),
+}
+
+impl PartialOrd for NeedsFiltering {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
```
ewienik, line 46: nit: could it be `derive(PartialOrd)`?

crates/vector-store/src/indexes.rs
```diff
@@ -0,0 +1,290 @@
...
+
+impl RoutingGroupKey {
+    pub(crate) fn from_metadata(metadata: &IndexMetadata) -> Self {
+        Self {
+            keyspace: metadata.keyspace_name.clone(),
+            table: metadata.table_name.clone(),
+            column: metadata.target_column.clone(),
+        }
+    }
+}
```
ewienik: nit: consider `impl From<&IndexMetadata>`

crates/vector-store/src/indexes.rs
```diff
@@ -0,0 +1,290 @@
...
+pub(crate) fn score_index(
+    index_type: &DbIndexType,
+    filtering_columns: &[ColumnName],
+    equality_columns: &[ColumnName],
+    range_columns: &[ColumnName],
+) -> Option<NeedsFiltering> {
+    let range_uncovered = range_columns
+        .iter()
+        .filter(|col| !filtering_columns.contains(col))
+        .count();
```
ewienik: If there is **any** uncovered range_columns or equality_columns it means index cannot serve the search.

crates/vector-store/src/indexes.rs
```diff
@@ -0,0 +1,290 @@
...
+
+    match index_type {
+        DbIndexType::Global => {
+            let eq_uncovered = equality_columns
+                .iter()
+                .filter(|col| !filtering_columns.contains(col))
+                .count();
+            let uncovered = eq_uncovered + range_uncovered;
```
ewienik: For global index all eq_columns and range_columns are uncovered. They can be calculated by `len()``.

crates/vector-store/src/indexes.rs
```diff
@@ -0,0 +1,290 @@
...
+                NeedsFiltering::No
+            } else {
+                NeedsFiltering::Yes(uncovered)
+            })
+        }
+        DbIndexType::Local(pk_columns) => {
+            let pk_match_count = pk_columns
+                .iter()
+                .filter(|col| equality_columns.contains(col))
+                .count();
```
ewienik: Consider check using `all()`

crates/vector-store/src/indexes.rs
```diff
@@ -0,0 +1,290 @@
...
+            if pk_match_count == 0 {
+                return None;
+            }
+            let eq_uncovered = equality_columns
+                .iter()
+                .filter(|col| !pk_columns.contains(col) && !filtering_columns.contains(col))
+                .count();
+            let uncovered = eq_uncovered + range_uncovered;
```
ewienik: consider `uncovered = equality_columns.len() - pk_columns.len() + range_columns.len()`

crates/vector-store/src/indexes.rs
```diff
@@ -0,0 +1,290 @@
...
+            equality_columns,
+            range_columns,
+        )?;
+        Some(async move {
+            matches!(entry.db_index.full_scan_progress().await, Progress::Done).then_some((
+                k,
+                score,
+                &entry.version,
+            ))
+        })
```
ewienik, line 280: it should check state with node_state, shouldn't it?

crates/vector-store/tests/integration/db_basic.rs
```diff
@@ -353,15 +356,19 @@ pub(crate) fn new_db_index(
...
         async move {
             let fullscan_fn = fullscan(&mut db, &metadata);
             let fullscan = {
                 let tx_embeddings = tx_embeddings.clone();
+                let fullscan_finished = fullscan_finished.clone();
                 async move {
                     if let Some(fullscan_fn) = fullscan_fn {
                         fullscan_fn(tx_embeddings).await;
                     }
+                    fullscan_finished.store(true, Ordering::Release);
```
ewienik, line 377: This seems like a fix for another problem. Consider putting this in a separate commit

crates/vector-store/tests/integration/quantization.rs
ewienik, line 1: Fixes in these tests seems like a separate issue. Consider creating additional commit

crates/vector-store/tests/integration/usearch.rs
ewienik, line 1: fix in this tests seems like different issue. Consider adding additional commit

ewienik (changes requested):

crates/vector-store/src/indexes.rs
QuerthDP: It is not possible to use default `PartialOrd` when there is custom `Ord` implemented

crates/vector-store/src/indexes.rs
QuerthDP: Changed

QuerthDP:
Changelog:
- removed `get_index_direct`, added `get_best_index` instead. `get_index` serves as `get_index_direct` now
- verified and tightened the index scoring and eligibility rules
- split the routing implementation and integration test fixes into 2 commits
- added `BestIndexState` to correctly inform why the best serving index was not found to use
- added integration test on routing to global indexes with filtering columns

QuerthDP:
Changelog:
- require ALLOW FILTERING for queries that need filtering

QuerthDP:
Changelog:
- rebased on master

QuerthDP:
Changelog:
- adjust filtering tested by adding required ALLOW FILTERING
- treat table primary key columns always like filtering columns (allow to filter on those)

QuerthDP:
Changelog:
- rebased on master (required due to last Validator infrastructure changes)

crates/vector-store/tests/integration/routing.rs
QuerthDP, line 1: Ah this should be fixuped in the previous commit

QuerthDP:
Changelog:
- move `routing.rs` integration tests fixup to correct commit

crates/vector-store/src/httproutes.rs
```diff
@@ -654,6 +654,15 @@ async fn post_index_ann(
...
 
+    state
+        .internals
+        .increment_counter(format!(
+            "ann-served-request--{}--{}",
+            routed_key.keyspace(),
+            routed_key.index(),
+        ))
+        .await;
+
```
ewienik, line 583: This is additional async call in hotpath. We ought to avoid this - the benchmarks shows that each async call adds latency. Try to find another way to check correctness of routing.

ewienik (changes requested):

crates/vector-store/src/httproutes.rs
ewienik: Possible solution could be to prepare unit tests for routing and post_index_ann to check possible paths. You can use axum-test crate to check endpoint handler without starting http server.

QuerthDP:
Changelog:
- added `slow-test-hooks` feature flag for hotpath internal counters enablement required in tests

crates/vector-store/src/httproutes.rs
QuerthDP: Added `slow-test-hooks` feature flag which should be only enabled in tests

QuerthDP:
Changelog:
- rebased on master

QuerthDP:
Changelog:
- rebased on master

QuerthDP:
Changelog:
- rebased on master

ewienik (approved):

knowack1 (approved):

PR #416, opened 2026-04-17
vector-store: add varint and decimal key types

m-szymon:
Add support for CQL `varint` and `decimal` types in primary keys and filtering.

Varint is handled via bigint encoding in the invariant key representation.

Decimal requires special handling: semantically equal values (e.g. `3.14` and
`3.140`) have different byte representations in Scylla's CQL encoding. To support
correct BTreeMap lookup, decimal clustering keys are normalized at insertion time
into a canonical form. The original byte representation from CDC/fullscan is
preserved in ANN responses. A `partition_key_count` field on `Table` distinguishes
partition keys (byte-identity only) from clustering keys (where normalization
applies). The normalization is skipped entirely when no decimal clustering columns
exist.

Semantic comparison via BigDecimal is used for CQL filtering on both global and
local indexes.

Also upgrades hickory-server 0.25 → 0.26 to fix RUSTSEC-2026-0118.

Fixes VECTOR-596


m-szymon:
The new tests will fail until fix in Scylla: https://github.com/scylladb/scylladb/pull/29505
Question - do we want to backport that fix?

@QuerthDP please review

m-szymon:
I filled some more places where Decimal and Varint types were missing, especially in unit tests.
I commented out new filtering tests, although the scylla fix https://github.com/scylladb/scylladb/pull/29505 is approved and should be merged soon. When that happens I will re-enable them, but in the meantime it already can be reviewed @QuerthDP, @knowack1, @ewienik

crates/validator/src/serde.rs
```diff
@@ -188,3 +195,243 @@ async fn test_varint_filter(actors: TestActors) {
         .await
         .expect("failed to drop keyspace");
 }
+
+/// Test the fundamental PK vs CK asymmetry for decimal types:
+///
+/// - PK is byte-identity: `1.0` and `1.00` are different partitions.
+/// - CK is semantic: `3.14` and `3.140` are the same row (overwrite).
```
ewienik, line 203: There is another fundamental problem - partition key columns can be used as restrictions and a part of partition keys could be used as partition keys for local indexes. In the future also clustering or non-primary key columns will be used as partition keys for local indexes. So the same column could be used as a partition key or as a filtering column - as I understand for the partition keys of indexes we need to preserve original decimal value and for filtering columns we need to normalize them.
Do we really need to store original decimal value? If so, we should store two versions of them (bigger memory consumption) or normalize decimals on the fly (slower filtering). I would go with normalizing on the fly.

ewienik (changes requested):

crates/validator/src/serde.rs
m-szymon: I changed it - now the ANN requests return original representation.
Normalization happens only in the BTreeMap - to update rows same as they are handled by the base table. BTreeMap doesn't have custom comparator - so two versions of key are stored (but only if they are different) - normalized BTreeMap key and original row data.

If the decimal column is used as a local index partition key - it is compared by bytes. If the column is used as filtering column it is compared by value - and this is consistent with MV behavior.
However in scylla original table is still requested and filtering happens. So e.g. in global index decimal partition key of base table will be filtered by value and several representations can be returned, but in scylla probably it will be filtered out.

ewienik: This solution is going to increase memory footprint, but no other one appeared.

ewienik (dismissed):

ewienik:
@m-szymon There are branch conflicts

m-szymon:
> @m-szymon There are branch conflicts

Fixed

ewienik (approved):

PR #392, opened 2026-03-17
validator: Alternator API test

m-szymon:
Add Alternator validator coverage for the main request paths exercised by
the Vector Store integration.

The new tests cover table creation, table updates, and table deletion, as
well as item operations including put, update, delete, and batch writes.
They also verify query behavior, including vector search requests and
result limiting.

Necessary AWS SDK dependencies are added to enable Alternator client.

Fixes: https://scylladb.atlassian.net/browse/VECTOR-581

PR #420, opened 2026-04-20
Support optimized Alternator Vector type

QuerthDP:
## Motivation

Alternator can store vectors using an optimized binary encoding, but the vector extraction path only supported the JSON list form. That made `FLOAT32VECTOR`-encoded vectors unusable even though they represent the same logical data.

Supporting both encodings makes the vector handling path consistent with Alternator behavior and ensures queries work regardless of which representation was used to write the item.

## Summary

Add support for Alternator's optimized `FLOAT32VECTOR` type encoding.

Until now, vector extraction only handled the existing JSON-based DynamoDB list representation. This change adds support for the optimized binary encoding used by Alternator.

## Changes

- add parsing for Alternator `FLOAT32VECTOR` type payloads encoded as sequential big-endian `f32` values
- keep support for the existing JSON `L` representation
- validate malformed binary payloads before decoding
- add unit coverage for the new parsing path in `vector-store`

## Test Coverage

- unit tests for Alternator vector blob parsing

Fixes: VECTOR-650

### DIFF
--- a/crates/vector-store/src/vector.rs
+++ b/crates/vector-store/src/vector.rs
@@ -50,7 +50,7 @@ impl TryFrom<CqlValue> for Vector {
                     Ok(f)
                 })
                 .collect(),
-            CqlValue::Blob(bytes) => parse_dynamodb_vector_json(&bytes),
+            CqlValue::Blob(bytes) => parse_alternator_vector(&bytes),
             other => Err(anyhow!(
                 "unsupported CQL type for embedding column: {other:?}"
             )),
@@ -59,23 +59,52 @@ impl TryFrom<CqlValue> for Vector {
     }
 }
 
-/// Alternator type tag for the DynamoDB List type (`L`), which is how vector embeddings are serialised.
-/// Alternator prefixes each attribute value in the `:attrs` map column with a 1-byte type discriminator.
-/// The List type uses the tag value `0x04` (named `NOT_SUPPORTED_YET`)
-const ALTERNATOR_TYPE_NOT_SUPPORTED_YET: u8 = 4;
+/// Alternator type tag for unoptimized JSON encoding.
+/// Type `0x04` (`NOT_SUPPORTED_YET`) is used for any type that does not have an optimized encoding.
+/// The payload is an unoptimized JSON value.
+const ALTERNATOR_TYPE_JSON: u8 = 4;
 
-/// Parses a DynamoDB-style JSON vector stored as raw bytes.
+/// Alternator type tag for the optimized `FLOAT32VECTOR` type.
+/// The value is serialized as this 1-byte tag followed by sequential 32-bit big-endian floats,
+/// matching the CQL `VECTOR<float, N>` on-wire encoding.
+const ALTERNATOR_TYPE_FLOAT32VECTOR: u8 = 5;
+
+/// Parses an Alternator-encoded vector stored as raw bytes.
 ///
-/// Handles two representations:
-/// - Plain JSON: `{"L": [{"N": "123.4"}, {"N": "234.5"}, ...]}`
-/// - Alternator-prefixed: a 1-byte type tag (`0x04`) followed by the JSON above.
-///   This prefix is used by Alternator for attribute values in the `:attrs` map.
-fn parse_dynamodb_vector_json(bytes: &[u8]) -> anyhow::Result<Vec<f32>> {
-    let bytes = match bytes.first() {
-        Some(&ALTERNATOR_TYPE_NOT_SUPPORTED_YET) => &bytes[1..],
-        _ => bytes,
-    };
+/// Alternator prefixes each attribute value in the `:attrs` map column with a 1-byte type discriminator.
+/// Handles two representations based on the discriminator:
+/// - [`ALTERNATOR_TYPE_FLOAT32VECTOR`]: optimized sequential 32-bit big-endian floats.
+/// - [`ALTERNATOR_TYPE_JSON`]: unoptimized JSON representing List values.
+fn parse_alternator_vector(bytes: &[u8]) -> anyhow::Result<Vec<f32>> {
+    match bytes.first() {
+        Some(&ALTERNATOR_TYPE_FLOAT32VECTOR) => parse_alternator_vector_binary(&bytes[1..]),
+        Some(&ALTERNATOR_TYPE_JSON) => parse_alternator_list_json(&bytes[1..]),
+        Some(tag) => bail!("unsupported Alternator type tag: {tag:#04x}"),
+        None => bail!("empty blob for Alternator attribute value"),
+    }
+}
 
+/// Parses the optimized Alternator vector encoding: sequential 32-bit big-endian floats.
+fn parse_alternator_vector_binary(bytes: &[u8]) -> anyhow::Result<Vec<f32>> {
+    let chunks = bytes.chunks_exact(4);
+
+    if !chunks.remainder().is_empty() {
+        bail!(
+            "invalid Alternator vector encoding: byte length {} is not a multiple of 4",
+            bytes.len()
+        );
+    }
+
+    Ok(chunks
+        .map(|chunk| {
+            let arr: [u8; 4] = chunk.try_into().expect("chunks_exact guarantees 4 bytes");
+            f32::from_be_bytes(arr)
+        })
+        .collect())
+}
+
+/// Parses an Alternator JSON list of numbers: `{"L": [{"N": "..."}, ...]}`.
+fn parse_alternator_list_json(bytes: &[u8]) -> anyhow::Result<Vec<f32>> {
     #[derive(serde::Deserialize)]
     struct DynamoDbList {
         #[serde(rename = "L")]
@@ -94,7 +123,7 @@ fn parse_dynamodb_vector_json(bytes: &[u8]) -> anyhow::Result<Vec<f32>> {
         .map(|item| {
             item.n
                 .parse::<f32>()
-                .map_err(|e| anyhow!("invalid value in DynamoDB vector element: {e}"))
+                .map_err(|e| anyhow!("invalid value in Alternator list element: {e}"))
         })
         .collect()
 }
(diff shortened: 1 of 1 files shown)


nyh:
Thanks! This PR has a bunch of other, unrelated commits - is it planned to merge them separately?

crates/vector-store/src/vector.rs
```diff
@@ -71,20 +71,46 @@ impl TryFrom<CqlValue> for Vector {
 /// Alternator type tag for the DynamoDB List type (`L`), which is how vector embeddings are serialised.
 /// Alternator prefixes each attribute value in the `:attrs` map column with a 1-byte type discriminator.
 /// The List type uses the tag value `0x04` (named `NOT_SUPPORTED_YET`)
-const ALTERNATOR_TYPE_NOT_SUPPORTED_YET: u8 = 4;
+const ALTERNATOR_TYPE_LIST: u8 = 4;
```
nyh, line 74: Note that type 4 isn't **just** a list, it is the byte used for any type we don't have an optimized encoding, so it is just this byte 4 (alternator_type::NOT_SUPPORTED_YET) followed by an unoptimized JSON value.

Perhaps the better name for it would have been ALTERNATOR_TYPE_JSON. 
The name "ALTERNATOR_TYPE_LIST" is a bit of misname. Yes, for the vector search's purpose, List is the only one that interests it, but other stuff - e.g., a map, would also look like this type 4.

QuerthDP:
> Thanks! This PR has a bunch of other, unrelated commits - is it planned to merge them separately?

Yes https://github.com/scylladb/vector-store/pull/392

crates/vector-store/src/vector.rs
```diff
@@ -71,20 +71,46 @@ impl TryFrom<CqlValue> for Vector {
...
-///   This prefix is used by Alternator for attribute values in the `:attrs` map.
+/// Handles two representations based on the first byte (the alternator type tag):
+/// - `0x05` (vector): optimized binary encoding — sequential 32-bit big-endian floats.
+/// - `0x04` (list): JSON `{"L": [{"N": "..."}, ...]}`.
 fn parse_dynamodb_vector_json(bytes: &[u8]) -> anyhow::Result<Vec<f32>> {
-    let bytes = match bytes.first() {
-        Some(&ALTERNATOR_TYPE_NOT_SUPPORTED_YET) => &bytes[1..],
-        _ => bytes,
-    };
+    match bytes.first() {
```
nyh, line 87: The suffix "_json" in this function's name, and in parse_alternator_vector_json, is not accurate, because the optimized vector type isn't JSON at all.

Also it's not clear why function is called "_dynamodb" and the second "_alternator".

I would consider renaming these functions.

crates/vector-store/src/vector.rs
```diff
@@ -71,20 +71,46 @@ impl TryFrom<CqlValue> for Vector {
...
+    if bytes.len() % 4 != 0 {
+        bail!(
+            "invalid alternator vector encoding: byte length {} is not a multiple of 4",
+            bytes.len()
+        );
+    }
+    Ok(bytes
+        .chunks_exact(4)
+        .map(|chunk| {
+            let arr: [u8; 4] = chunk.try_into().expect("chunks_exact guarantees 4 bytes");
```
nyh, line 106: It looks like you have 3 (!?) separate checks here that the vector is a multiple of 4 bytes. Rust is a weird language ;-)

nyh (approved):
I'll let the vector-store developers review the full contents of this PR (which seems to have some unrelated patches in it). But I'm "approve"ing because the good news is that this patch does exactly what Alternator needs to support the optimized vector type. I checked, and all the tests I wrote for Alternator with this type pass with this vector-store patch.

QuerthDP:
Changelog:
- added commit messages
- renamed parsing functions
- adjusted comments and error messages

crates/vector-store/src/vector.rs
QuerthDP: Actually there was only one effective check :)
But it's true that Rust is really explicit in that case with mentioning the 4 many times.

crates/vector-store/src/vector.rs
QuerthDP: Renamed

crates/vector-store/src/vector.rs
QuerthDP: Fixed

QuerthDP:
@ewienik you can take a look at the [af1a7ed](https://github.com/scylladb/vector-store/pull/420/commits/af1a7edef8a94ec6e2fbaeb7c60ecfe405429ecd) patch. The previous commits are coming from https://github.com/scylladb/vector-store/pull/392 and the last one is based on that, so let's not review that yet.

QuerthDP:
BTW we're still waiting for https://github.com/scylladb/scylladb/pull/29554 to decide how would we name the type. It is possible that the Vector won't be encoded with  `V` discriminator.

crates/vector-store/src/vector.rs
```diff
@@ -68,23 +68,52 @@ impl TryFrom<CqlValue> for Vector {
...
-///   This prefix is used by Alternator for attribute values in the `:attrs` map.
-fn parse_dynamodb_vector_json(bytes: &[u8]) -> anyhow::Result<Vec<f32>> {
-    let bytes = match bytes.first() {
-        Some(&ALTERNATOR_TYPE_NOT_SUPPORTED_YET) => &bytes[1..],
-        _ => bytes,
-    };
+/// Alternator prefixes each attribute value in the `:attrs` map column with a 1-byte type discriminator.
+/// Handles two representations based on the discriminator:
+/// - `0x05` (vector): optimized sequential 32-bit big-endian floats.
+/// - `0x04` (list): unoptimized JSON representing List values.
```
ewienik: nit: comment doesn't align with the function body - function uses named values.

crates/vector-store/src/vector.rs
```diff
@@ -150,6 +179,24 @@ impl TryFrom<AlternatorAttrs<'_>> for Option<Vector> {
 mod tests {
     use super::*;
 
+    /// Prepend the Alternator JSON type tag (0x04) to a DynamoDB JSON string,
```
ewienik: nit: body uses named value instead of actual binary tag

crates/vector-store/src/vector.rs
```diff
@@ -150,6 +179,24 @@ impl TryFrom<AlternatorAttrs<'_>> for Option<Vector> {
...
+    /// mirroring how Alternator serialises List values.
+    fn alternator_list_blob(json: &str) -> Vec<u8> {
+        let mut v = vec![ALTERNATOR_TYPE_JSON];
+        v.extend_from_slice(json.as_bytes());
+        v
+    }
+
+    /// Prepend the Alternator Vector type tag (0x05) to a sequence of big-endian floats,
```
ewienik: nit: body uses named value instead of actual binary tag

ewienik:
> @ewienik you can take a look at the [af1a7ed](https://github.com/scylladb/vector-store/pull/420/commits/af1a7edef8a94ec6e2fbaeb7c60ecfe405429ecd) patch. The previous commits are coming from #392 and the last one is based on that, so let's not review that yet.

The commit seems ok

nyh:
> BTW we're still waiting for [scylladb/scylladb#29554](https://github.com/scylladb/scylladb/pull/29554) to decide how would we name the type. It is possible that the Vector won't be encoded with `V` discriminator.

This discussion is relevant for the tests that need to send Alternator commands, but not the actual code. The actual code that reads the data will continue, like today, to have a single byte "5" followed by 32-bit floats. This shouldn't change, because the whole point of this feature was to make this thing efficient - and it can't be 64-bit floats or anything else.

nyh:
@QuerthDP can you please rebase this PR? As I noted above, the name "V" changed to the new name "FLOAT32VECTOR" in the latest version of https://github.com/scylladb/scylladb/pull/29554, but I don't think this makes any difference for your patch.

Thanks. I would love to see this patch merged even before the everlasting merge freeze on scylla.git is lifted, so when we want to merge https://github.com/scylladb/scylladb/pull/29554, the vector store would be ready.

QuerthDP:
> Thanks. I would love to see this patch merged even before the everlasting merge freeze on scylla.git is lifted, so when we want to merge [scylladb/scylladb#29554](https://github.com/scylladb/scylladb/pull/29554), the vector store would be ready.

I was waiting with this patch as it's blocked by both https://github.com/scylladb/scylladb/pull/29554 and https://github.com/scylladb/vector-store/pull/392.
Without the Scylla one, the tests won't pass as our CI only runs on some scylla-nightly build, so that we need the Scylla patch first.
Without the Alternator test API (on which this PR is based) we cannot write tests at all!

If you really want this patch to be merged before. I may split it into separate feature and test PRs, but we would need to have it mind, that the feature will not be CI tested in that case (as we already did for the previous patch, so probably not a big deal).

nyh:
I see. I just updated the vector store (and forgot to use your branch) and was surprised to see this feature disappeared :-) I forgot it was an unmerged PR.

I guess we can wait.  The splitting of feature and test is also good I think, but it's not urgent and will cause you more work.

QuerthDP:
> I see. I just updated the vector store (and forgot to use your branch) and was surprised to see this feature disappeared :-) I forgot it was an unmerged PR.
> 
> I guess we can wait. The splitting of feature and test is also good I think, but it's not urgent and will cause you more work.

TBH with so many merge conflicts after rebase on master, it would be less work to split the patch and wait for Alternator API to get in first :)

crates/vector-store/src/vector.rs
QuerthDP: fixed

crates/vector-store/src/vector.rs
QuerthDP: fixed

crates/vector-store/src/vector.rs
QuerthDP: fixed

QuerthDP:
Changelog:
- rebased on master
- removed the Validator tests, those will be included in [separate PR](https://github.com/scylladb/vector-store/pull/437)

QuerthDP:
`cargo test` failed due to known issue: VECTOR-552
Rerunning.

ewienik (approved):

PR #421, opened 2026-04-20
fix(db_index): prevent orphaned CDC readers across db_index replacement

swasik:
### Problem

On a single vector-store instance, after a cluster topology change, logs showed repeated creation of the fine CDC reader for the same index. Old CDC reader generations were staying alive after `db_index` replacement.

### Root cause

`db_index::new` spawned a small monitor task that owned `cdc_wide` and `cdc_fine` and only awaited their `.closed()` signals to forward `cdc_error_notify`. When the engine replaced a `db_index` (e.g. after schema/topology change), that monitor kept the old CDC senders alive even though the corresponding main db_index task was gone. The old CDC actors then continued running concurrently with the new generation.

### Fix

Add a `Notify` (`db_index_stopped`) from the main db_index task to the CDC monitor task:

- The main task signals `db_index_stopped` when it exits (e.g. when `tx_index` is dropped by the engine).
- The CDC monitor adds a third branch to its `select!` that exits on `db_index_stopped.notified()`.
- On monitor exit, CDC senders are dropped, terminating old CDC actors.
- Behavior on CDC `.closed()` path (notifying `cdc_error_notify`) is unchanged.

The change is intentionally minimal: ~20 lines in `db_index.rs`.

### Regression test

A second commit adds a deterministic validator-level regression test (`cdc::recreating_index_terminates_old_cdc_actors`) that exercises the invariant directly. The test:

1. Creates an index with CDC enabled.
2. Waits until both CDC actors (wide + fine) report `-cdc-actor-started` via the internals counters endpoint.
3. `DROP INDEX` + `CREATE INDEX` with the same name, forcing a `db_index` replacement.
4. Asserts that both old CDC actors terminate (both `-cdc-actor-stopped` counters reach 1) within 10 seconds.

Supporting instrumentation is added in `db_cdc.rs`: two counter increments per CDC actor lifecycle. The counters are no-op unless registered via `internals_start_counter`, so there is no runtime cost in production.

Verified locally against `scylladb/scylla:2026.1` in Docker:

- With the fix reverted: **10/10 runs fail** on the `wait_for` timeout for the stopped counters.
- With the fix applied: **10/10 runs pass**.

Fixes VECTOR-653

### DIFF
--- a/crates/validator/src/cdc.rs
+++ b/crates/validator/src/cdc.rs
@@ -12,6 +12,7 @@ use vector_search_validator_tests::TestCase;
 
 const FINE_GRAINED_CDC_MAX_LATENCY: Duration = Duration::from_secs(2);
 const CDC_MAX_LATENCY: Duration = Duration::from_secs(60);
+const CDC_ACTOR_STOP_TIMEOUT: Duration = Duration::from_secs(10);
 
 #[framed]
 pub(crate) async fn new() -> TestCase<TestActors> {
@@ -37,6 +38,11 @@ pub(crate) async fn new() -> TestCase<TestActors> {
         .with_test("cdc_lwt_insert_visible", timeout, cdc_lwt_insert_visible)
         .with_test("cdc_lwt_update_visible", timeout, cdc_lwt_update_visible)
         .with_test("cdc_lwt_delete_visible", timeout, cdc_lwt_delete_visible)
+        .with_test(
+            "recreating_index_terminates_old_cdc_actors",
+            timeout,
+            recreating_index_terminates_old_cdc_actors,
+        )
 }
 
 #[framed]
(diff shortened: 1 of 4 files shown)


swasik:
The failing `validator-tests (cdc)` job is a pre-existing flaky test on `master`, not caused by this change.

The most recent master run (after #418 merged) — run [24675936648](https://github.com/scylladb/vector-store/actions/runs/24675936648) — fails with the same test and the same error signature:

```
test{name="cdc_lwt_update_visible"}: test failed: task 322 panicked with message "failed to run query:
LastAttemptError(DbError(Invalid, "Vector Store error: HTTP status 404 Not Found,
message: missing index: ksp_4.idx_4"))"
failed_tests: ["cdc::cdc_lwt_update_visible"]
```

The failures on this PR (404 "missing index" / 503 "still being constructed") all occur right after `Waiting for index to be SERVING` begins — the vector store never reports the index as SERVING within the wait window. This is independent of the CDC reader lifecycle changed in this PR. I've retriggered the failed jobs.

ewienik (changes requested):
> Fix
> ---
> Move ownership of `cdc_wide` and `cdc_fine` into the main `db_index`
> task and observe their `.closed()` signals inside the existing
> `select!` loops. When the main task exits -- either because
> `tx_index` was dropped by the engine, or because a CDC actor closed
> -- the senders go out of scope and the CDC actors terminate cleanly.
> The error notification to `cdc_error_notify` is preserved for the
> CDC-close case

It could be done easier, without moving ownership. Send notify event from main loop to the monitor loop using `Notify` type.

swasik:
> It could be done easier, without moving ownership. Send notify event from main loop to the monitor loop using `Notify` type.

Done.

crates/validator/src/db_index_replace.rs
ewienik, line 1: as this is test for cdc functionality, consider moving it into cdc.rs module.

crates/validator/src/db_index_replace.rs
```diff
@@ -0,0 +1,169 @@
...
+) {
+    let cql = format!(
+        "CREATE CUSTOM INDEX {index_name} ON {table}({vector_column}) USING 'vector_index'"
+    );
+    info!("Create index: '{cql}'");
+    session
+        .query_unpaged(cql, ())
+        .await
+        .expect("failed to create an index");
+}
```
ewienik: Consider adding method `index_name()` to `CreateIndexQuery` instead of creating a new function for it.

crates/validator/src/db_index_replace.rs
```diff
@@ -0,0 +1,169 @@
...
+    .await;
+
+    let index = client
+        .indexes()
+        .await
+        .into_iter()
+        .find(|idx| idx.index.as_ref() == index_name)
+        .expect("index not found");
+    wait_for_index(client, &index).await;
+}
```
ewienik: consider using a common wait_for_index function

ewienik (changes requested):

QuerthDP (commented):
Looks fine to me.
Please consider applying @ewienik's comments. I agree with them, it should be changed.
Apart from that it seems all right.

crates/validator/src/db_index_replace.rs
swasik: Fixed.

crates/validator/src/db_index_replace.rs
swasik: Fixed.

crates/validator/src/db_index_replace.rs
swasik: Done.

swasik:
Fixed @ewienik comments.

ewienik (approved):

PR #424, opened 2026-04-23
docs: add manual runbook for replacing a dead Vector Store node

mykaul:
## Summary

- Adds a step-by-step manual procedure for replacing a dead Vector Store node, aimed at self-managed ScyllaDB Enterprise users (non-Cloud)
- Modeled after the [ScyllaDB "Replace a Dead Node" doc](https://docs.scylladb.com/manual/stable/operating-scylla/procedures/cluster-management/replace-dead-node.html)
- Derived from the automated replacement logic in [scylladb/siren](https://github.com/scylladb/siren) (`cluster/vectorstore/vectorstore.go`, `cluster/vectorstore/configurator.go`, `cluster/server/vector_store.go`)

## What's covered

- Prerequisites (healthy cluster, collecting config from existing nodes)
- Full procedure: provision instance, firewall rules, configure `.env` + tmpfs password + TLS certs, start service, wait for SERVING status, update `scylla.yaml` on ScyllaDB nodes, decommission old node, update monitoring
- Verification steps
- Operational notes: hot reload via SIGHUP, single-ScyllaDB-node connection model, port/file/env-var reference tables

### DIFF
--- a/docs/replace-dead-node.md
+++ b/docs/replace-dead-node.md
@@ -1,0 +1,296 @@
+# Replace a Dead Vector Store Node
+
+The replace operation provisions a new Vector Store node in the same availability zone as the dead node, waits for it to fetch all data from ScyllaDB and rebuild its indexes, then decommissions the old node. This is an **add-then-remove** strategy that ensures zero downtime for vector search queries (assuming other VS nodes in the cluster remain healthy).
+
+The duration of this procedure depends on the data volume and index complexity. Index rebuilding can take anywhere from a few minutes to over an hour.
+
+## Prerequisites
+
+### Healthy ScyllaDB Cluster
+
+The ScyllaDB cluster must be healthy and reachable. The new Vector Store node needs to connect to a ScyllaDB node via CQL to fetch data and build indexes.
+
+### Identify the Dead Node
+
+Collect the following information from the dead Vector Store node (or from your infrastructure records):
+
+- **Private IP address** of the dead node
+- **Availability zone / rack** the node was running in
+- **Instance type** (e.g., `i3.xlarge`)
+- **Vector Store version** (the version of the `vector-store` binary or AMI)
+
+### Collect Configuration from a Healthy Vector Store Node
+
+If you have other healthy Vector Store nodes in the cluster, SSH into one and collect:
+
+```bash
+# View the current configuration
+cat /home/ubuntu/.env
+```
+
+This will show the environment variables used by the Vector Store service, including the CQL username, password file path, ScyllaDB connection URI, and TLS settings. You will replicate this configuration (with appropriate IP changes) on the replacement node.
+
+### Collect CQL Credentials
+
+You need the CQL username and password that Vector Store uses to connect to ScyllaDB. On a healthy VS node:
+
+```bash
+# Username is in the .env file
+grep VECTOR_STORE_SCYLLADB_USERNAME /home/ubuntu/.env
+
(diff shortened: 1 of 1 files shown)


QuerthDP:
> aimed at self-managed ScyllaDB Enterprise users (non-Cloud)

Is not VS only Cloud feature?

swasik:
> > aimed at self-managed ScyllaDB Enterprise users (non-Cloud)
> 
> Is not VS only Cloud feature?

Currently but we plan to release as part of Enterprise as well. We are just not sure when.

swasik:
@mykaul I think it should be generalized - this procedure is also used for scaling nodes and upgrading to new vector store version.

docs/replace-dead-node.md
```diff
@@ -0,0 +1,296 @@
...
+```bash
+# View the current configuration
+cat /home/ubuntu/.env
+```
+
+This will show the environment variables used by the Vector Store service, including the CQL username, password file path, ScyllaDB connection URI, and TLS settings. You will replicate this configuration (with appropriate IP changes) on the replacement node.
+
+### Collect CQL Credentials
+
+You need the CQL username and password that Vector Store uses to connect to ScyllaDB. On a healthy VS node:
```
swasik, line 35: This partially duplicates the previous paragraph.

docs/replace-dead-node.md
```diff
@@ -0,0 +1,296 @@
...
+> **Note:** The instance name convention used by ScyllaDB Cloud is
+> `Scylla-Cloud-<CLUSTER_ID>-<DC_NAME>-VectorSearch-<NODE_ID>` on AWS, and
+> `scylla-cloud-<cluster_id>-vectorsearch-<node_id>-<checksum>` on GCP.
+> You can use any naming convention that fits your environment.
+
+### Step 2: Configure Firewall Rules
+
+The following network access is required:
+
+| Direction | From | To | Port | Purpose |
```
swasik, line 65: +9100 for node_exporter

docs/replace-dead-node.md
```diff
@@ -0,0 +1,296 @@
...
+sudo journalctl -u vector-store -f
+```
+
+### Step 5: Wait for the Node to Become Ready
+
+After starting, the Vector Store node fetches data from ScyllaDB and builds its indexes. Poll the status endpoint until it returns `SERVING`:
+
+```bash
+# Without TLS:
+watch -n 30 'curl -s http://localhost:6080/api/v1/status'
```
swasik, line 186: nit: split it into two listings - it will be easier to copy the command to the clipboard.

docs/replace-dead-node.md
```diff
@@ -0,0 +1,296 @@
...
+
+The Vector Store service supports hot configuration reload via `SIGHUP`. If you need to change the `.env` file after the service is running (e.g., to update the ScyllaDB IP or credentials), edit the file and then:
+
+```bash
+sudo systemctl kill -s SIGHUP vector-store
+```
+
+### ScyllaDB Node Failure
+
+Each Vector Store node is configured to connect to a **single ScyllaDB node** in the same rack. If that ScyllaDB node fails, the Vector Store node's configuration must be updated to point to a different ScyllaDB node in the same rack:
```
swasik, line 263: It is not completely true. As long as the driver has connection to at least one node in the cluster the topology will refresh automatically. And as we are using rack aware policy we will automatically switch to the correct node (or fail over to different rack if none of them are left).

PR #428, opened 2026-04-24
e2etest: remove dependency on vector-store and httpclient

ewienik:
e2etest framework should not depend on any crate from vector-store repository, as no crate from that repo shouldn't be published in crates.io. For this reason the commit copy a part for httpclient to retrieve a node status. It is a short duplicate, but e2etest crate eventually will be in different repo and published, so this seems a reasonable move.

Fixes: VECTOR-657

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1221,13 +1221,14 @@ dependencies = [
 name = "e2etest-vector-store-cluster"
 version = "0.0.0-dev"
 dependencies = [
+ "anyhow",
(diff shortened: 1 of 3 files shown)


ewienik:
Changelog for 62cf41f3
- rebase to master

<details>
<summary>range diff</summary>

```
1:  03e22bfb = 1:  62cf41f3 e2etest: remove dependency on vector-store and httpclient
```

swasik (approved):

PR #430, opened 2026-04-27
validator: implement test for CQL per-row TTL

smoczy123:
Verify that rows inserted with USING TTL are filtered out from ANN query results after they expire. The test inserts rows with and without TTL, creates an index, waits for expiry, and asserts that only non-TTL rows are returned.

Fixes: VECTOR-598

### DIFF
--- a/crates/e2etest-scylla-cluster/src/lib.rs
+++ b/crates/e2etest-scylla-cluster/src/lib.rs
@@ -51,6 +51,7 @@ static DEFAULT_SCYLLA_ARGS: LazyLock<RwLock<Vec<String>>> = LazyLock::new(|| {
             "--logger-log-level=migration_manager=warn",
             "--shutdown-announce-in-ms=0",
             "--tablets-mode-for-new-keyspaces=enabled",
+            "--alternator-ttl-period-in-seconds=0.5",
         ]
         .into_iter()
         .map(String::from)
--- a/crates/validator/src/cdc.rs
+++ b/crates/validator/src/cdc.rs
@@ -8,11 +8,14 @@ use crate::common::*;
 use async_backtrace::framed;
 use e2etest::TestCase;
 use std::time::Duration;
+use std::time::SystemTime;
+use std::time::UNIX_EPOCH;
 use tracing::info;
 
 const FINE_GRAINED_CDC_MAX_LATENCY: Duration = Duration::from_secs(2);
 const CDC_MAX_LATENCY: Duration = Duration::from_secs(60);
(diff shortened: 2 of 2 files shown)


crates/validator/src/ttl.rs
```diff
@@ -0,0 +1,133 @@
...
+            result.filter(|r| r.rows_num() == 5)
+        },
+        "Waiting for ANN query to return all 5 rows",
+        DEFAULT_OPERATION_TIMEOUT,
+    )
+    .await;
+    assert_eq!(result.rows_num(), 5, "Expected 5 rows before TTL expiry");
+
+    info!("Wait for TTL to expire");
+    time::sleep(Duration::from_secs(ROW_TTL_SECONDS as u64 + 2)).await;
```
ewienik: Consider waiting for index size to decrement - in that case maybe this sleep is not needed.

crates/validator/src/ttl.rs
ewienik, line 1: Is this is a test for CDC? If so, should it go into cdc.rs?

ewienik (changes requested):

crates/validator/src/ttl.rs
smoczy123: We have to specify a launch option for scylla, so I'd prefer a different file

crates/validator/src/ttl.rs
smoczy123: changed

smoczy123:
Changelog:

- changed to actually use the new ttl functionality

crates/validator/src/ttl.rs
```diff
@@ -0,0 +1,182 @@
...
+/// Custom init that starts ScyllaDB with a low TTL scan period so that the
+/// expiration service picks up expired rows quickly.
+#[framed]
+async fn init_ttl(actors: TestActors) {
+    info!("started");
+
+    let mut scylla_configs = get_default_scylla_node_configs(&actors).await;
+    for cfg in &mut scylla_configs {
+        cfg.args
+            .push("--alternator-ttl-period-in-seconds=0.5".to_string());
```
ewienik: What about add this config as a default?

crates/validator/src/ttl.rs
ewienik: We should specify this option as a default, what do you think?

crates/validator/src/ttl.rs
smoczy123: Added

smoczy123:
Changelog:

-  moved the test to cdc.rs
-  added new default launch option
-  rebased onto master

ewienik (approved):

PR #431, opened 2026-04-29
vector-store: implement ALTER INDEX simulator

ewienik:
Currently we do not support ALTER INDEX, so to be able to test different rescoring index options we need a possibility to lock the index in the memory - only for tests.

It changes processing of monitor_indexes actor. When simulator is enabled, indexes aren't deleted after DELETE INDEX, they stay in the memory. When a new CREATE INDEX is build with the same keyspace and index names the actors compares all fields without version - if only version is different then no new index is created (the existing one in memory is used). It simulates ALTER INDEX behavior.                                                                      

It introduces env variable VECTOR_STORE_ALTER_INDEX_SIMULATOR with true/false value. When env variable is true vector-store will change behavior of monitor indexes actor.

Fixes: VECTOR-659


crates/vector-store/src/monitor_indexes.rs
```diff
@@ -705,4 +705,147 @@ mod tests {
...
+            },
+            &[idx.clone()].into_iter().collect()
+        ));
+        assert!(should_delete_simulator(
+            &IndexMetadata {
+                expansion_add: 1.into(),
+                ..idx.clone()
+            },
+            &[idx.clone()].into_iter().collect()
+        ));
```
QuerthDP, line 788: What case is this in CREATE/DELETE words? Is it possible we have the same index (the same version) but different parameters?

QuerthDP: If it's the case with ALTER INDEX then we would recreate the index anyway, so I don't understand why we need those changes. The only difference I see is that DROP->CREATE of the exactly same index does nothing.

crates/vector-store/src/monitor_indexes.rs
```diff
@@ -705,4 +705,147 @@ mod tests {
...
+    }
+
+    #[test]
+    fn validate_should_delete_simulator() {
```
QuerthDP, line 751: I'd like to see a test with different `expansion_add` and `version`. New or possibly modified the last case

crates/vector-store/src/monitor_indexes.rs
ewienik: You're right - we remove index only when other parameters than version differs. The test usecase is to drop index in scylla (vector-store has index in memory), create index with the same name and options beside rescoring, oversampling (vector-store uses previous index - no parameters used during creation changed). We can test different oversampling and rescoring parameters.

crates/vector-store/src/monitor_indexes.rs
ewienik: I've provided test with changed expansion_add or version in validate_should_delete_simulator. What kind of other usecase should I add?

QuerthDP: I'd see the test changing both expansion_add and version. I think it can be the last one but modified with the addition of version

ewienik:
Changelog for 759c7a13
- add requested assert check

<details>
<summary>range diff</summary>

```
1:  77d29101 ! 1:  759c7a13 vector-store: implement unit tests for alter index simulator
    @@ crates/vector-store/src/monitor_indexes.rs: mod tests {
     +            },
     +            &[idx.clone()].into_iter().collect()
     +        ));
    ++        assert!(should_delete_simulator(
    ++            &IndexMetadata {
    ++                expansion_add: 1.into(),
    ++                version: Uuid::new_v4().into(),
    ++                ..idx.clone()
    ++            },
    ++            &[idx.clone()].into_iter().collect()
    ++        ));
     +    }
     +
     +    #[test]
```

crates/vector-store/src/monitor_indexes.rs
ewienik: fixed in 759c7a13

QuerthDP (approved):

PR #435, opened 2026-05-04
validator: add filtering tests that were removed from test.py

smoczy123:
We add two test cases from test.py that were removed when filtering support was added. We change the ck test, as now we support it.

Fixes: VECTOR-479

### DIFF
--- a/crates/validator/src/filtering.rs
+++ b/crates/validator/src/filtering.rs
@@ -7,7 +7,10 @@ use crate::TestActors;
 use crate::common::*;
 use async_backtrace::framed;
 use e2etest::TestCase;
+use httpapi::KeyspaceName;
+use scylla::client::session::Session;
 use std::collections::HashSet;
+use std::sync::Arc;
 use tracing::info;
 
 #[framed]
@@ -56,6 +59,21 @@ pub(crate) async fn new() -> TestCase<TestActors> {
             "ann_filter_by_non_indexed_column_fails",
             timeout,
             ann_filter_by_non_indexed_column_fails,
+        )
+        .with_test(
+            "ann_filter_by_clustering_key_only_requires_allow_filtering",
+            timeout,
+            ann_filter_by_clustering_key_only_requires_allow_filtering,
+        )
+        .with_test(
+            "ann_filter_by_non_pk_column_rejected_without_allow_filtering",
+            timeout,
+            ann_filter_by_non_pk_column_rejected_without_allow_filtering,
+        )
+        .with_test(
+            "ann_filter_by_non_pk_column_rejected_with_allow_filtering",
+            timeout,
+            ann_filter_by_non_pk_column_rejected_with_allow_filtering,
         )
         .with_test(
             "local_index_filter_by_partition_key_eq",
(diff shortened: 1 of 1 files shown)


crates/validator/src/filtering.rs
```diff
@@ -1024,3 +1034,183 @@ async fn local_ann_with_timestamp_gte_filter(actors: TestActors) {
...
+            )
+            .await
+            .is_some()
+        },
+        "Waiting for filtered ANN query (ck=1 only) to be operational",
+        DEFAULT_OPERATION_TIMEOUT,
+    )
+    .await;
+    let result = get_query_results(
+        format!("SELECT p, ck FROM {table} WHERE ck = 1 ORDER BY v ANN OF [0.1, 0.2, 0.3] LIMIT 5"),
```
ewienik: This query should failed without `ALLOW FILTERING` as it is filtering without partition, shouldn't it?

ewienik (changes requested):

crates/validator/src/filtering.rs
```diff
@@ -1024,3 +1034,183 @@ async fn local_ann_with_timestamp_gte_filter(actors: TestActors) {
 
     info!("finished");
 }
+
+#[framed]
+async fn ann_filter_by_clustering_key_only(actors: TestActors) {
```
knowack1: Function is very long. Just complaining but I think we accept this in tests.

crates/validator/src/filtering.rs
```diff
@@ -1024,3 +1034,183 @@ async fn local_ann_with_timestamp_gte_filter(actors: TestActors) {
...
+        ),
+        &session,
+    )
+    .await;
+    let rows: Vec<(i32, i32)> = result
+        .rows::<(i32, i32)>()
+        .expect("failed to get rows")
+        .map(|row| row.expect("failed to get row"))
+        .collect();
+    assert_eq!(
```
knowack1: Mixing actions and assertions are generally bad batter in test cases. Consider splitting into two test cases.

crates/validator/src/filtering.rs
```diff
@@ -1024,3 +1034,183 @@ async fn local_ann_with_timestamp_gte_filter(actors: TestActors) {
...
+    info!("Create index on non-PK column c");
+    session
+        .query_unpaged(format!("CREATE INDEX ON {table}(c)"), ())
+        .await
+        .expect("failed to create index on c");
+
+    info!("Test ANN query with indexed non-PK column filtering");
+    session
+        .query_unpaged(
+            format!("SELECT * FROM {table} WHERE c = 1 ORDER BY v ANN OF [0.1, 0.2, 0.3] LIMIT 5"),
```
knowack1: Wondering if this will pass in Cassandra.

knowack1 (dismissed):

crates/validator/src/filtering.rs
smoczy123: I think we allow filtering with ck only without ALLOW FILTERING

crates/validator/src/filtering.rs
smoczy123: split into smaller

crates/validator/src/filtering.rs
smoczy123: split

crates/validator/src/filtering.rs
smoczy123: I think it should? Not sure how filtering works in Cassandra.

crates/validator/src/filtering.rs
smoczy123: or we don't? It seemed to work locally.

smoczy123: Fixed

crates/validator/src/filtering.rs
```diff
@@ -1025,3 +1040,193 @@ async fn local_ann_with_timestamp_gte_filter(actors: TestActors) {
...
+        .await
+        .expect("failed to drop a keyspace");
+
+    info!("finished");
+}
+
+#[framed]
+async fn ann_filter_by_non_pk_column_rejected_without_allow_filtering(actors: TestActors) {
+    info!("started");
+    assert_non_pk_column_filter_rejected(actors, false).await;
```
knowack1: Maybe now it has gone too far regarding the number of lines in the test case. I would either extract some repeated pieces of code for the test setup or ensure the test function remains explicit about what is being tested and asserted. It is always a dilemma of how much to repeat versus how explicit to be in a test case - The answer is to find a good balance.

crates/validator/src/filtering.rs
```diff
@@ -1025,3 +1040,193 @@ async fn local_ann_with_timestamp_gte_filter(actors: TestActors) {
...
+    expected_len_message: &str,
+) {
+    assert_eq!(rows.len(), expected_len, "{expected_len_message}");
+    assert!(
+        rows.iter().all(|(_, ck)| *ck == expected_ck),
+        "Expected only rows with ck={expected_ck}"
+    );
+}
+
+async fn assert_non_pk_column_filter_rejected(actors: TestActors, with_allow_filtering: bool) {
```
knowack1: The function is named `assert_` which suggest that this is a tiny wrapper for multiline assertion. While this in fact is entire test case.

nyh:
Hi guys, I was adding pre-filtering (filtering on the vector store side) and projection (returning filtering columns together with the result) to Alternator, and I discovered that filtering was only partially implemented in the vector store, and fixed it and expanded it in the following PR: https://github.com/scylladb/vector-store/pull/444

Now I'm worried that maybe I've been colliding with some unmerged work that you guys are doing in parallel :-(

Can you please take a look?

knowack1:
> Hi guys, I was adding pre-filtering (filtering on the vector store side) and projection (returning filtering columns together with the result) to Alternator, and I discovered that filtering was only partially implemented in the vector store, and fixed it and expanded it in the following PR: #444
> 
> Now I'm worried that maybe I've been colliding with some unmerged work that you guys are doing in parallel :-(
> 
> Can you please take a look?

I left some comments on #444. I think this falls under the scope of VECTOR-561, which is assigned to @ewienik , but I don't see any associated development branch.

smoczy123:
Changelog:

- adressed @knowack1 comments

knowack1 (approved):

ewienik (approved):

PR #436, opened 2026-05-05
Improve release scripts reliability and UX

swasik:
## Changes

This PR covers release scripts reliability and UX improvements only. Cross-compilation changes have been moved to #448.

- Add unified `scripts/release` orchestrator: validates deps, QEMU setup, tag create/push, build, upload, summary
- Semver validation regex for version argument
- Colored step-counter output in all scripts (disabled in non-TTY)
- `GH_REPO` env var for `gh` CLI scoping (avoids `.gh-resolution` file side effects from `gh repo set-default`)
- SBOM files organized into `SBOM/` directory
- Temp file cleanup via `EXIT` trap + `INT`/`TERM` trap in `run-with-release-toolchain`
- `--init` flag on `docker run` for proper signal forwarding
- `-t` flag on `docker run` when stdout is a terminal
- Quote `$repo` / `$repo_dir` in `realpath` and `cd` to handle paths with spaces
- Docker push skipped for forks (only pushes from `scylladb/vector-store`)
- Release notes Copilot prompt printed after completion
- Updated `docs/releasing.md` documentation

## Tests

- Tested by creating a test release in the fork: https://github.com/swasik/vector-store/releases/tag/1.6.1-dev
- Part of the script was also tested while creating the last official release: https://github.com/scylladb/vector-store/releases/tag/1.6.0

## Known issues

The `build-cloud-images` script that is triggered when the version is created fails due to lack of authentication in GCP (RELENG-77).

Fixes: VECTOR-665


scripts/_build-release
```diff
@@ -7,48 +7,88 @@ repo=$(realpath $repo/..)
...
+    echo ""
+}
+
+step_header "Resolving version and verifying repository state"
+
+if [[ -n $1 ]]; then
+    vector_store_version=$1
+    echo "  Version (override): $vector_store_version"
+else
+    vector_store_version=$(./scripts/run-with-release-toolchain cargo info vector-store | grep ^version: | cut -d ' ' -f 2)
```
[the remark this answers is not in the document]
swasik: You are correct. I implemented fix myself.

scripts/upload-release
```diff
@@ -7,42 +7,104 @@ repo_dir=$(realpath $repo_dir/..)
...
 
-# verify the tag exists
-git rev-parse "$tag" > /dev/null 2>&1 || { echo "error: tag $tag does not exist"; exit 1; }
+git rev-parse "$tag" > /dev/null 2>&1 || { echo "${red}error: tag $tag does not exist${reset}"; exit 1; }
+echo "  ${green}Tag $tag exists.${reset}"
 
-# verify gh cli is available
-command -v gh > /dev/null 2>&1 || { echo "error: gh cli is not installed (https://cli.github.com/)"; exit 1; }
+command -v gh > /dev/null 2>&1 || { echo "${red}error: gh cli is not installed (https://cli.github.com/)${reset}"; exit 1; }
+echo "  ${green}gh CLI available.${reset}"
```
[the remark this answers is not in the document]
swasik: Fixed.

docs/releasing.md
```diff
@@ -1,5 +1,17 @@
...
+
+```bash
+./scripts/release X.Y.Z
+```
+
+This single command validates dependencies, sets up Docker QEMU, creates and
+pushes the annotated tag, builds all artifacts (binaries, docker images, SBOMs),
+and uploads everything to GitHub and Docker Hub. Pass `--version-override` if
+you need the version passed explicitly to build/upload scripts instead of
+relying on git tag detection.
```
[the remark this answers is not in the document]
swasik: Fixed.

docs/releasing.md
```diff
@@ -28,34 +42,41 @@ version tag. Then from the root of the repository run:
...
 
 ```bash
-docker push scylladb/vector-store:X.Y.Z-amd64
-docker push scylladb/vector-store:X.Y.Z-arm64
-docker manifest push scylladb/vector-store:X.Y.Z
+./scripts/build-release 1.6.0
 ```
 
+This will create a tar.gz release in the `releases/{amd64/arm64}` directory,
+per-architecture docker images, and SBOM files.
```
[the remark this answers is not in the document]
swasik: Fixed.

scripts/upload-release
```diff
@@ -7,42 +7,104 @@ repo_dir=$(realpath $repo_dir/..)
...
 done
 
-gh release upload "$tag" "${assets[@]}" --clobber
+step_header "Creating and pushing multi-platform Docker manifest"
 
-echo "done"
+docker_manifest=scylladb/vector-store:$vector_store_version
+docker buildx imagetools create -t $docker_manifest \
+    scylladb/vector-store:$vector_store_version-amd64 \
+    scylladb/vector-store:$vector_store_version-arm64
```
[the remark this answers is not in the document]
swasik: Fixed. I blocked docker pushes from forks.

scripts/release
```diff
@@ -0,0 +1,130 @@
...
+
+step_header "Building release"
+
+./scripts/build-release "$version"
+
+step_header "Uploading release"
+
+gh repo set-default "$github_repo"
+
+./scripts/upload-release "$version"
```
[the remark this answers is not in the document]
swasik: This is just the tag that we will have to create in the end anyway. What is most visible to the user is release which is created at the very end. It is not worth complicating the code.

scripts/release
```diff
@@ -0,0 +1,130 @@
...
+}
+
+step_header "Validating dependencies"
+
+missing=()
+command -v docker > /dev/null 2>&1 || missing+=("docker")
+command -v cargo > /dev/null 2>&1 || missing+=("cargo")
+command -v cargo-cyclonedx > /dev/null 2>&1 || missing+=("cargo-cyclonedx (cargo install cargo-cyclonedx)")
+command -v syft > /dev/null 2>&1 || missing+=("syft (https://github.com/anchore/syft)")
+command -v gh > /dev/null 2>&1 || missing+=("gh (https://cli.github.com/)")
```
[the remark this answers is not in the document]
swasik: Added.

scripts/release
```diff
@@ -0,0 +1,130 @@
...
+echo ""
+echo "--------"
+echo "Generate release notes based on the PRs linked here:"
+echo "https://github.com/scylladb/vector-store/releases/tag/$version"
+echo "- prepare them in Markdown format. For each PR analyze its"
+echo "description, not just the title. Assume that everything related"
+echo "to validator should be gathered in the QA related section. Make"
+echo "them consistent in style with release notes for the previous"
+echo "version available at"
+echo "https://github.com/scylladb/vector-store/releases/."
```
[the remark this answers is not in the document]
swasik: Fixed.

scripts/release
```diff
@@ -0,0 +1,136 @@
...
+step_header "Building release"
+
+./scripts/build-release "$version"
+
+step_header "Uploading release"
+
+gh repo set-default "$github_repo"
+
+./scripts/upload-release "$version"
+
```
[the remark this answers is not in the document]
swasik: Fixed.

scripts/build-release
```diff
@@ -57,19 +97,33 @@ RUN mkdir /opt/vector-store
...
+echo "  - vector-store-$vector_store_version.cdx.json"
 for arch in $archs ; do
-    docker_arch_tags+=("scylladb/vector-store:$vector_store_version-$arch")
+    echo "  - vector-store-docker-$vector_store_version-$arch.cdx.json"
+    echo "  - scylladb/vector-store:$vector_store_version-$arch (local docker image)"
 done
-docker manifest create $docker_manifest "${docker_arch_tags[@]}"
-echo "Docker manifest: $docker_manifest"
+echo ""
+echo "Next step: run ./scripts/upload-release to publish."
```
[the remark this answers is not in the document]
swasik: Fixed.

docs/releasing.md
```diff
@@ -18,8 +28,10 @@ GitHub doesn't support creating annotated tags, so you must do them from local
 environment. You can add additional metadata to the tag, for example
 `X.Y.Z-dev` for development releases.
 
-Script `scripts/run-with-release-toolchain` can create a correct version for
-the `Cargo.toml` file based on annotated tags using `git describe --dirty`.
+Script `scripts/run-with-release-toolchain` is a Docker wrapper that derives the
+version from annotated tags using `git describe --dirty` and patches `Cargo.toml`
+accordingly. You do not need to run it manually — it is called automatically by
+`scripts/build-release`.
```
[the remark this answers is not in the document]
swasik: I validated the whole file in the other thread.

scripts/upload-release
```diff
@@ -7,42 +7,121 @@ repo_dir=$(realpath $repo_dir/..)
...
+    echo "    - $asset"
 done
 
-gh release upload "$tag" "${assets[@]}" --clobber
+if gh release view "$tag" > /dev/null 2>&1; then
+    echo "  Release $tag already exists, uploading assets."
+    gh release upload "$tag" "${assets[@]}" --clobber
+else
+    echo "  Creating release $tag with auto-generated notes."
+    gh release create "$tag" --generate-notes "${assets[@]}"
```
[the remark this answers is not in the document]
swasik: This is passed from release script using env variable not to mutate the default repo.

scripts/run-with-release-toolchain
```diff
@@ -3,38 +3,102 @@
...
+if [[ -n $user_cargo_home ]]; then
+    docker_args+=(-v "$user_cargo_home/git:$docker_cargo_home/git")
+    docker_args+=(-v "$user_cargo_home/registry:$docker_cargo_home/registry")
+fi
+
+[[ -t 1 ]] && docker_args+=(-t)
+
+# Cross-compilation: install cross-toolchain, add Rust target, set CC/CXX/linker.
+# Runs as root so apt-get can install packages.
+if [[ "$is_cross" == "true" ]]; then
```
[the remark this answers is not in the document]
swasik: It is a single use scenario, we do not care about polution.

scripts/_build-release
```diff
@@ -7,69 +7,164 @@ repo=$(realpath $repo/..)
...
+    echo "  Version: $vector_store_version"
+fi
```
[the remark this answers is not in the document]
swasik: It is logged to output. If it fails, which has low probability, it will be easy to understand what happened.

scripts/release
```diff
@@ -0,0 +1,134 @@
+#!/bin/bash
+
+set -e
+
+repo=$(dirname "${BASH_SOURCE[0]}")
+repo=$(realpath $repo/..)
+
+cd $repo
```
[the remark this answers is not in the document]
swasik: Fixed.

scripts/release
```diff
@@ -0,0 +1,134 @@
...
+echo "  ${green}Pushed tag $version to origin ($github_repo).${reset}"
+
+step_header "Building release"
+
+./scripts/build-release "$version"
+
+step_header "Uploading release"
+
+GH_REPO="$github_repo" ./scripts/upload-release "$version"
+
```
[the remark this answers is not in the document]
swasik: PR description updated.

scripts/build-release
```diff
@@ -7,69 +7,164 @@ repo=$(realpath $repo/..)
 
 cd $repo
```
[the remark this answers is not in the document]
swasik: Fixed.

docs/releasing.md
```diff
@@ -1,11 +1,23 @@
...
+## TL;DR
+
+```bash
+./scripts/release X.Y.Z
+```
+
+This single command validates dependencies, sets up Docker QEMU (for image
+packaging), creates and pushes the annotated tag, builds all artifacts
+(binaries, docker images, SBOMs), and uploads everything to GitHub and Docker
+Hub.
```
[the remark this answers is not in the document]
swasik: Fixed.

scripts/run-with-release-toolchain
```diff
@@ -3,38 +3,102 @@
...
+    rust_target_upper=$(echo "$rust_target" | tr '[:lower:]-' '[:upper:]_')
+    docker_args+=(
+        -e "CARGO_TARGET_${rust_target_upper}_LINKER=${cross_prefix}gcc"
+        -e "CC_${rust_target}=${cross_prefix}gcc"
+        -e "CXX_${rust_target}=${cross_prefix}g++"
+        -e "AR_${rust_target}=${cross_prefix}ar"
```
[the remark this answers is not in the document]
swasik: You are correct. Fixed.

scripts/run-with-release-toolchain
```diff
@@ -3,38 +3,102 @@
...
+    amd64)
+        rust_target="x86_64-unknown-linux-gnu"
+        cross_pkg=""
+        cross_prefix=""
+        ;;
+    arm64)
+        rust_target="aarch64-unknown-linux-gnu"
+        cross_pkg="gcc-aarch64-linux-gnu g++-aarch64-linux-gnu"
+        cross_prefix="aarch64-linux-gnu-"
+        ;;
```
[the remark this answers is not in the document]
swasik: We will not build releases on ARM.

scripts/build-release
```diff
@@ -7,69 +7,164 @@ repo=$(realpath $repo/..)
...
+    step=$((step + 1))
+    echo ""
+    echo "  ${bold}${cyan}--> [build $step/$total_steps] $1${reset}"
+    echo ""
+}
+
+step_header "Validating build dependencies"
+
+missing=()
+command -v docker > /dev/null 2>&1 || missing+=("docker")
```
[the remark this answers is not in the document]
swasik: Those dependency checks here as not needed as this script will be called from release. I moved them there.

scripts/upload-release
```diff
@@ -7,42 +7,131 @@ repo_dir=$(realpath $repo_dir/..)
...
+
+step_header "Verifying prerequisites"
+
+if [[ -n $1 ]]; then
+    vector_store_version=$1
+    echo "  Version (override): $vector_store_version"
+else
+    vector_store_version=$(./scripts/run-with-release-toolchain cargo info vector-store | grep ^version: | cut -d ' ' -f 2)
+    echo "  Version: $vector_store_version"
+fi
```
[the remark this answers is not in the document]
swasik: It is logged to output. If it fails, which has low probability, it will be easy to understand what happened.

scripts/upload-release
```diff
@@ -7,42 +7,131 @@ repo_dir=$(realpath $repo_dir/..)
 
 cd $repo_dir
```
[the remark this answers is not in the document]
swasik: Fixed.

scripts/build-release
```diff
@@ -57,22 +57,45 @@ trap 'echo ""; echo "  Interrupted."; exit 130' INT TERM
...
+        arm64) echo "aarch64-unknown-linux-gnu" ;;
+    esac
+}
+
+# Resolve the release binary path (cross builds add a target-triple subdir).
+release_dir_for() {
+    local arch=$1
+    local target
+    target=$(rust_target_for "$arch")
+    if [[ -f "target/$arch/$target/release/vector-store" ]]; then
```
ewienik: It is a dependency on dynamic behavior. What about use always target/triple and update .cargo/config.toml `target="triple"` in the docker configuration?

scripts/run-with-release-toolchain
```diff
@@ -48,6 +75,31 @@ fi
...
+        -e "RUST_CROSS_TARGET=$rust_target"
+        -e "FIX_UID=$(id -u)"
+        -e "FIX_GID=$(id -g)"
+        -e "TARGET_DIR=./target/$TARGETARCH"
+    )
+    docker_args+=(-w "$repo" "$image" sh -c "
+        apt-get update -qq && apt-get install -y -qq \$CROSS_PACKAGES > /dev/null 2>&1 &&
+        rustup target add \$RUST_CROSS_TARGET &&
+        \"\$@\" && chown -R \$FIX_UID:\$FIX_GID \$TARGET_DIR
+    " -- "$@")
```
ewienik: It seems that preparing specific docker image would be a better solution. Now, everytime docker is run needs to prepare environment - it should be done in Dockerfile once. This docker could be uploaded to docker hub.
For this reason I would recommend prepare another PP only for this cross compiling.

scripts/upload-release
```diff
@@ -56,8 +56,18 @@ step_header "Collecting and uploading GitHub release assets"
 assets=()
 
 for arch in $archs ; do
-    tarball=target/$arch/release/vector-store-$vector_store_version-$arch.tar.gz
-    [[ -f $tarball ]] || { echo "${red}error: $tarball not found, run scripts/build-release first${reset}"; exit 1; }
+    # Cross-compilation may place artifacts in a target-triple subdir.
```
ewienik: Again dynamic behavior. Consider prepare deterministic target/triple dir

scripts/run-with-release-toolchain
```diff
@@ -48,6 +75,31 @@ fi
...
-docker_args+=(-w "$repo" "$image" "$@")
+# Cross-compilation: install cross-toolchain, add Rust target, set CC/CXX/linker.
+# Runs as root so apt-get can install packages.
+if [[ "$is_cross" == "true" ]]; then
+    rust_target_upper=$(echo "$rust_target" | tr '[:lower:]-' '[:upper:]_')
+    rust_target_underscored=$(echo "$rust_target" | tr '-' '_')
+    docker_args+=(
+        -e "CARGO_TARGET_${rust_target_upper}_LINKER=${cross_prefix}gcc"
+        -e "CC_${rust_target_underscored}=${cross_prefix}gcc"
+        -e "CXX_${rust_target_underscored}=${cross_prefix}g++"
```
ewienik: Is qemu docker with arm also uses gcc or clang? What kind of compiler does previous $image use? We should prepare a frozen toolchain docker to control this.

scripts/build-release
```diff
@@ -3,52 +3,93 @@
...
+# Colors (disabled if not a terminal).
+if [[ -t 1 ]]; then
+    bold=$'\033[1m'
+    red=$'\033[0;31m'
+    green=$'\033[0;32m'
+    cyan=$'\033[0;36m'
+    reset=$'\033[0m'
+else
+    bold="" red="" green="" cyan="" reset=""
+fi
```
ewienik: This part appears in several scripts. Consider using a single script for its definition.
Also consider preparing functions `text_emphasize(txt)` for bold, `text_error()` for red, etc... - to show the semantic rather than colors directlly

Do we really need these colors at all?

scripts/build-release
```diff
@@ -3,52 +3,93 @@
...
+
+step=0
+total_steps=5
+
+step_header() {
+    step=$((step + 1))
+    echo ""
+    echo "  ${bold}${cyan}--> [build $step/$total_steps] $1${reset}"
+    echo ""
+}
```
ewienik: This function is also reused in other scripts. Consider place it in a common one for reuse.

scripts/run-with-release-toolchain
```diff
@@ -3,38 +3,51 @@
 set -e
 
 repo=$(dirname "${BASH_SOURCE[0]}")
-repo=$(realpath $repo/..)
+repo=$(realpath "$repo/..")
```
ewienik, line 6: this is a refactoring. Consider creating additional commit for refactoring for better review.

scripts/run-with-release-toolchain
```diff
@@ -3,38 +3,51 @@
...
 
 TARGETARCH="${TARGETARCH:-amd64}"
 
-rust_version=$(grep channel $repo/rust-toolchain.toml | cut -d '"' -f 2)
-image=rust:$rust_version-bookworm
+rust_version=$(grep channel "$repo/rust-toolchain.toml" | cut -d '"' -f 2)
+image="rust:$rust_version-bookworm"
```
ewienik: we don't need these `"`

scripts/run-with-release-toolchain
```diff
@@ -3,38 +3,51 @@
...
 
 user_cargo_home=$CARGO_HOME
 [[ -z $user_cargo_home ]] && [[ -d "$HOME/.cargo" ]] && user_cargo_home="$HOME/.cargo"
 
-docker_cargo_home=$(docker run --rm $image sh -c 'echo $CARGO_HOME')
+docker_cargo_home=$(docker run --rm "$image" sh -c 'echo $CARGO_HOME')
```
ewienik: not needed `"`

docs/releasing.md
```diff
@@ -18,8 +31,12 @@ GitHub doesn't support creating annotated tags, so you must do them from local
...
 `X.Y.Z-dev` for development releases.
 
-Script `scripts/run-with-release-toolchain` can create a correct version for
-the `Cargo.toml` file based on annotated tags using `git describe --dirty`.
+Script `scripts/run-with-release-toolchain` is a Docker wrapper that derives the
+version from annotated tags using `git describe --dirty` and patches `Cargo.toml`
+accordingly. It handles cross-compilation setup (installing cross-toolchains,
+setting `CARGO_BUILD_TARGET`, linker, and C/C++ compiler environment variables).
+You do not need to run it manually — it is called automatically by
+`scripts/build-release`.
```
ewienik: cross compilation is another commit - but I think cross compilation should be different PR

docs/releasing.md
```diff
@@ -28,38 +45,45 @@ version tag. Then from the root of the repository run:
 ./scripts/build-release
 ```
 
-This will create a tar.gz release in the `releases/{amd64/arm64}` directory,
-per-architecture docker images, and a multi-platform docker manifest. You can
-push the docker images and manifest to docker hub:
+Optionally, you can pass a version explicitly to override automatic detection
+from git tags:
```
ewienik, line 51: Why do we need this override? We shouldn't allow creating same versions from different commits or build setup. Providing annotate tag before is a way to provide this. We should rather check if scylladb/vector-store has the same annotated tag as locally created.

scripts/_build-release
```diff
@@ -57,19 +98,33 @@ RUN mkdir /opt/vector-store
...
 done
 
-echo create multi-platform docker manifest
-docker_manifest=scylladb/vector-store:$vector_store_version
-docker_arch_tags=()
+echo ""
+echo "${bold}${green}========================================"
+echo " BUILD COMPLETE: vector-store $vector_store_version"
+echo "========================================${reset}"
+echo ""
```
ewienik, line 97: Do we really need this fancy stuff in release script?

scripts/_build-release
```diff
@@ -3,52 +3,93 @@
...
     exit 1
 fi
+echo "  ${green}Repository is clean.${reset}"
+
+# Restore Cargo.toml and Cargo.lock from git if any build step fails.
+restore_cargo_files() {
+    git checkout -- Cargo.toml Cargo.lock 2>/dev/null && echo "  Restored Cargo.toml and Cargo.lock from git."
+}
+trap restore_cargo_files EXIT
+trap 'echo ""; echo "  Interrupted."; exit 130' INT TERM
```
ewienik, line 39: This is needed for sbom. Can it be moved after run-with-release-toolchain to not change local git repository for that script?

scripts/release
```diff
@@ -0,0 +1,134 @@
...
+if git rev-parse "$version" > /dev/null 2>&1; then
+    echo "  Tag $version already exists, skipping creation."
+else
+    git tag -a "$version" -m "Release version $version"
+    echo "  ${green}Created annotated tag: $version${reset}"
+fi
+
+git push origin "$version"
+echo "  ${green}Pushed tag $version to origin ($github_repo).${reset}"
```
ewienik: I wouldn't allow pushing anything to repository from release script. I prefer for developer to create a tag on specific commit, push manually this specific annotated tag and release script should read it only without writing anything to repo. It is less error prone in the future if someone run this script on the wrong commit. Doing this manually we can check if tag is on the correct commit without pushing it to the repository.

ewienik (changes requested):

docs/releasing.md
swasik: Done, moved to #448.

scripts/build-release
swasik: I will fix it in #448.

scripts/run-with-release-toolchain
swasik: I will add in #448 the new dedicated Dockerfile that will pin both the Rust toolchain version and the cross-compiler packages, giving us a frozen and reproducible build environment.

scripts/upload-release
swasik: Will fix in #448.

scripts/build-release
swasik: Agreed — extracted shared color definitions and `step_header()` into `scripts/_common.sh` which is sourced by all release scripts. Added semantic helpers (`info()`, `error()`, `success()`). Keeping colors for interactive terminal UX — they're disabled automatically in non-TTY contexts.

> Do we really need these colors at all?

Of course we do not need them. But the output of the script is really long and IMO they make much easier to analyze its output.

scripts/build-release
swasik: Done — moved to `scripts/_common.sh`, see above.

scripts/run-with-release-toolchain
swasik: This PR now contains only the release scripts UX commit. Cross-compilation was split into a separate PR (#448).

scripts/run-with-release-toolchain
swasik: Fixed.

scripts/run-with-release-toolchain
swasik: Fixed.

docs/releasing.md
swasik: The override exists because `scripts/release` orchestrator resolves the version from the annotated tag and passes it down to `_build-release` and `_upload-release`. This avoids each sub-script re-resolving it independently. The `build-release` script is not intended to be called manually in normal workflows — the entry point is `scripts/release`. Clarified this in `docs/releasing.md` and added underscores as the prefix of internal scripts to make it easier to spot.

scripts/_build-release
swasik: As above - we do not need them but the summary provides immediate visibility of all produced artifacts and their paths, which is useful when debugging or verifying a release build.

scripts/_build-release
swasik: This runs in step 4 (SBOM generation), which is after all builds complete in step 2. The `EXIT` trap ensures `Cargo.toml` and `Cargo.lock` are restored if anything fails, so git state is never left dirty. The `sed` is needed because `cargo-cyclonedx` reads the version from `Cargo.toml` to embed in the SBOM.

scripts/release
swasik: I'd prefer to keep tag creation and push in the script for these reasons:

1. The script validates semver format before creating the tag, preventing typos.
2. It skips creation if the tag already exists (so manual pre-creation works too).
3. Having it in one place reduces the chance of pushing a tag from the wrong commit — the script runs on the current HEAD which you can verify with `git log`.
4. The push is to `origin` (fork) — pushing to `upstream` for the official release is a separate manual step documented in `releasing.md`.

If the tag already exists on remote, the push is a no-op. If you still feel strongly, I could add a `--no-tag` flag to skip tag creation/push for workflows where tags are pre-created or the direct acknowledgement question.

scripts/run-with-release-toolchain
swasik: This will be added in #448.

docs/releasing.md
ewienik: > The override exists because `scripts/release` orchestrator resolves the version from the annotated tag and passes it down to `_build-release` and `_upload-release`.

It seems not true. `scripts/release` has mandatory version arguments. It creates annotated tag, if it doesn't exist. If the tag exists it skip creating annotated tag, but it doesn't check if annotated tag is created in current commit. It seems error prone to me.

ewienik: I think that annotated tag should be created by hand, manually. At first release script could resolve current version, resolve it, show to the user. Then version can be read by every script. It could be double check for correctness that the user want to build that version. In the future creating tag in CI should trigger all releasing process - so the version taken from annotated-tag should be de-factor a parameter to the scripts.

scripts/release
ewienik: I don't agree with providing version by hand. The clue with annotated tag is that it provides a more robust way to label version - it is in the git version history so it will stay. We shouldn't have easy way to build version not connected to the specific git commit.

> I'd prefer to keep tag creation and push in the script for these reasons:
> 
>     1. The script validates semver format before creating the tag, preventing typos.

You can still validate semver format of version taken from annotated tag - you can check if the tag in remote repository points to the same commit as in local repository, you can check if tag points to the HEAD, etc...

>     2. It skips creation if the tag already exists (so manual pre-creation works too).

The manual pre-creation should be the only way - it will restrict creation of release, but it will remove all maintenance problem with version provided by hand in the script. I think we should at first point the version in git history, then build release for this point - in other case we could end up with few releases with the same version (ie. concurrent release of the version by two developers).
> 
>     3. Having it in one place reduces the chance of pushing a tag from the wrong commit — the script runs on the current HEAD which you can verify with `git log`.

And I think this could be error prone. Do you now that your HEAD points to the correct git commit? Without pre-creating annotated-tag you can mixed commits on local repo. The sync point is pushing annotated-tag to the git repo - after this operation you are sure, that only one tag exists and you can check if it points to the correct commit.

> 
>     4. The push is to `origin` (fork) — pushing to `upstream` for the official release is a separate manual step documented in `releasing.md`.

It depends how you setup your local git repositry. I can have different remotes, origin could be scylldb, my fork could be named differently. Our scripts shouldn't restrict local git repository layout.
 
> If the tag already exists on remote, the push is a no-op. If you still feel strongly, I could add a `--no-tag` flag to skip tag creation/push for workflows where tags are pre-created or the direct acknowledgement question.

There are problems with concurrently running release, creating tag, etc. We should limit possible combinations - one way is to restrict that we are building version taken from git history and annotated-tag ought to be created before releasing process, manually.

scripts/release
```diff
@@ -0,0 +1,124 @@
...
+cd "$repo"
+source scripts/_common.sh
+
+# Override step_header for top-level release orchestrator style.
+step_header() {
+    _step=$((_step + 1))
+    echo ""
+    echo "${bold}${cyan}==> [${_step}/${total_steps}] $1${reset}"
+    echo ""
+}
```
ewienik, line 17: Do we really need additional override?

scripts/run-with-release-toolchain
```diff
@@ -3,38 +3,51 @@
...
+    -v "$tmp_cargo:$repo/Cargo.toml"
+    -v "$tmp_rust_toolchain:$repo/rust-toolchain.toml"
+)
+
+if [[ -n $user_cargo_home ]]; then
+    docker_args+=(-v "$user_cargo_home/git:$docker_cargo_home/git")
+    docker_args+=(-v "$user_cargo_home/registry:$docker_cargo_home/registry")
+fi
+
+[[ -t 1 ]] && docker_args+=(-t)
```
ewienik, line 49: do we need this `-t`?

scripts/_build-release
ewienik: The ideal would be to prepare frozen toolchain and add cargo-cylonedx into it. Then we would have call to this toolchain docker without need to change local git.

swasik:
This was already moved forward by releng team.

PR #437, opened 2026-05-06
validator: add Alternator FLOAT32VECTOR type end-to-end tests

QuerthDP:
Add end-to-end coverage for the Alternator optimized vector type using requests with the new encoding injected into the JSON body.

Verify that queries work for items stored with the optimized vector representation and remain compatible with the existing L-based vector representation in the same index.

Fixes: VECTOR-666

### DIFF
--- a/crates/validator/src/alternator/mod.rs
+++ b/crates/validator/src/alternator/mod.rs
@@ -24,6 +24,7 @@ use aws_sdk_dynamodb::error::SdkError;
 use aws_sdk_dynamodb::operation::create_table::CreateTableError;
 use aws_sdk_dynamodb::operation::create_table::CreateTableOutput;
 use aws_sdk_dynamodb::operation::delete_table::DeleteTableError;
+use aws_sdk_dynamodb::primitives::Blob;
 use aws_sdk_dynamodb::types::AttributeDefinition;
 use aws_sdk_dynamodb::types::AttributeValue;
 use aws_sdk_dynamodb::types::BillingMode;
@@ -34,6 +35,7 @@ use aws_smithy_runtime_api::box_error::BoxError;
 use aws_smithy_runtime_api::client::interceptors::Intercept;
 use aws_smithy_runtime_api::client::interceptors::context::BeforeTransmitInterceptorContextMut;
 use aws_smithy_runtime_api::client::runtime_components::RuntimeComponents;
+use aws_smithy_types::base64;
 use aws_smithy_types::body::SdkBody;
 use aws_smithy_types::config_bag::ConfigBag;
 use e2etest::TestCase;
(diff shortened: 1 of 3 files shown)


crates/validator/src/alternator/mod.rs
```diff
@@ -97,6 +97,7 @@ pub(crate) async fn test_cases() -> Vec<(String, TestCase<TestActors>)> {
         ("alternator_create_table".into(), create_table::new().await),
         ("alternator_update_table".into(), update_table::new().await),
         ("alternator_put_item".into(), put_item::new().await),
+        ("alternator_query".into(), query::new().await),
```
m-szymon: Unnecessary.

crates/validator/src/alternator/query.rs
```diff
@@ -733,6 +757,154 @@ async fn query_with_filter_expression(actors: TestActors) {
...
+    );
+
+    ctx.done().await;
+
+    info!("finished");
+}
+
+/// Verifies query works end-to-end with optimized `V` vector encoding.
+#[framed]
+async fn query_with_vector_v_type(actors: TestActors) {
```
m-szymon: I don't think that this test much different from previous.

crates/validator/src/alternator/query.rs
```diff
@@ -733,6 +757,154 @@ async fn query_with_filter_expression(actors: TestActors) {
...
+    put_item_with_optimized_vector(
+        &ctx.client,
+        &ctx.table_name,
+        pk_name,
+        pk_v,
+        vec_name,
+        [1.0, 0.0, 0.0],
+    )
+    .await;
+    ctx.client
```
m-szymon: ctx has 'put' method.

crates/validator/src/alternator/query.rs
```diff
@@ -733,6 +757,154 @@ async fn query_with_filter_expression(actors: TestActors) {
...
+    let shape = TableShape {
+        table_prefix: None,
+        index_prefix: None,
+        pk_name: "p".into(),
+        sk_name: None,
+        vec_name: Some("v".into()),
+        pk_type: ScalarAttributeType::S,
+    };
+
+    let ctx = TableContext::create(&actors, &shape).await;
```
m-szymon: These tests don't check initial scan path

crates/validator/src/alternator/query.rs
```diff
@@ -733,6 +757,154 @@ async fn query_with_filter_expression(actors: TestActors) {
...
+
+    ctx.wait_for_count(2).await;
+
+    let items = ctx
+        .client
+        .query()
+        .table_name(&ctx.table_name)
+        .index_name(ctx.index.index.as_ref())
+        .limit(2)
+        .vector_search([1.0_f32, 0.0, 0.0])
```
m-szymon: We should probably also test queries with optimized vector.

m-szymon (changes requested):
1. I was considering adding such tests to types.rs tests (added here https://github.com/scylladb/vector-store/pull/467). I deliberately named it types, not key_types.
2. Adding optimized data to index and querying are slightly separate issues.
3. I was considering defining test datasets using "binary" type and only extending the json injector to convert them to FLOAT32VECTOR when expected - it could better reuse existing utils.

But it works, so consider if you would like to improve anything and we can merge it. I can later refactor.

crates/validator/src/alternator/mod.rs
QuerthDP: Oh indeed, missed that

crates/validator/src/alternator/query.rs
QuerthDP: Merged

crates/validator/src/alternator/query.rs
QuerthDP: Changed so that it does check both paths

crates/validator/src/alternator/mod.rs
QuerthDP: Removed

QuerthDP:
Changelog:
- rebased on master
- moved tests from `query.rs` to `types.rs`
- merged the tests into one, which checks both insert and query

crates/validator/src/alternator/query.rs
QuerthDP: Used

m-szymon:
It is easier to ask AI to do it, than explain it: https://github.com/scylladb/vector-store/commit/3459c12379f9d878d9486dd0cdeeeba937028f15

The problem is the `pk_v_scan` is not inserted at scan phase.
I was thinking about adding same trick as @nyh used for Java driver: https://github.com/scylladb/alternator-client-java/pull/87

I also noticed that other test in that file don't actually check returned values. With that fixed, we could extend and use the `query_with_key_type` function.

ewienik:
> It is easier to ask AI to do it, than explain it: [3459c12](https://github.com/scylladb/vector-store/commit/3459c12379f9d878d9486dd0cdeeeba937028f15)
> 
> The problem is the `pk_v_scan` is not inserted at scan phase. I was thinking about adding same trick as @nyh used for Java driver: [scylladb/alternator-client-java#87](https://github.com/scylladb/alternator-client-java/pull/87)
> 
> I also noticed that other test in that file don't actually check returned values. With that fixed, we could extend and use the `query_with_key_type` function.

@m-szymon Does it mean that something needs to be done in this PR?

m-szymon:
> > It is easier to ask AI to do it, than explain it: [3459c12](https://github.com/scylladb/vector-store/commit/3459c12379f9d878d9486dd0cdeeeba937028f15)
> > The problem is the `pk_v_scan` is not inserted at scan phase. I was thinking about adding same trick as @nyh used for Java driver: [scylladb/alternator-client-java#87](https://github.com/scylladb/alternator-client-java/pull/87)
> > I also noticed that other test in that file don't actually check returned values. With that fixed, we could extend and use the `query_with_key_type` function.
> 
> @m-szymon Does it mean that something needs to be done in this PR?

Yes - one small inconsistency that `pk_v_scan` is not a scan test.
We can either remove/ignore it and enhance tests separately or fix from the beginning.

QuerthDP:
Changelog:
- fixed up with @m-szymon's commit https://github.com/scylladb/vector-store/commit/3459c12379f9d878d9486dd0cdeeeba937028f15 

It looks fine to me. Added you as a co-author.

m-szymon (approved):
> It looks fine to me. Added you as a co-author.

As a co-author my approve is not enough :)

@ewienik

ewienik (approved):

PR #438, opened 2026-05-06
vector-store: fix ALTER INDEX simulator

ewienik:
There is an issue with simulating ALTER INDEX by DROP/CREATE similar index, that the index after this operation is sometimes in INITIALIZING state. The root cause is that node_state receives information about the all indexes read from db without fixing the list by simulation. This commit fix this issue by preparing a correct list of discovered indexes for node_state.

Fixes: VECTOR-669

### DIFF
--- a/crates/vector-store/src/monitor_indexes.rs
+++ b/crates/vector-store/src/monitor_indexes.rs
@@ -79,9 +79,27 @@ pub(crate) async fn new(
                             schema_version.reset();
                             continue;
                         };
-                        node_state.send_event(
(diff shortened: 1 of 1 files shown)


knowack1 (approved):
Is the reason for this code that Scylla does not implement ALTER INDEX? Is my understanding correct?

ewienik:
> Is the reason for this code that Scylla does not implement ALTER INDEX? Is my understanding correct?

Yes, this is a reason for this change - it is needed to test several scylla-only parameters without rebuilding the index.

PR #439, opened 2026-05-06
vector-store: add Tantivy dependency and runtime FTS configuration

knowack1:
Add Tantivy 0.22 as a workspace dependency for full-text search support.
Introduce `VECTOR_STORE_FTS_ENABLED` env var (`true`/`false`, default: `false`) to
control FTS at runtime, parsed in `config_manager` alongside existing env vars.
Add `fts_enabled` field to `Config` and a placeholder `fts` module.

Fixes: VECTOR-620

### DIFF
--- a/crates/vector-store/src/config_manager.rs
+++ b/crates/vector-store/src/config_manager.rs
@@ -305,6 +305,18 @@ pub async fn load_config(env: impl Fn(&str) -> anyhow::Result<String>) -> anyhow
         config.disable_colors = disable_colors;
     }
 
+    if let Some(fts_enabled) = env("VECTOR_STORE_FTS_ENABLED")
+        .ok()
+        .map(|v| {
+            v.trim().parse().or(Err(anyhow!(
+                "Unable to parse VECTOR_STORE_FTS_ENABLED env (true/false)"
+            )))
+        })
+        .transpose()?
+    {
+        config.fts_enabled = fts_enabled;
+    }
+
     if let Some(vector_store_addr) = env("VECTOR_STORE_URI")
         .ok()
         .map(|v| {
--- a/crates/vector-store/src/fts.rs
+++ b/crates/vector-store/src/fts.rs
@@ -1,0 +1,6 @@
+/*
+ * Copyright 2026-present ScyllaDB
+ * SPDX-License-Identifier: LicenseRef-ScyllaDB-Source-Available-1.1
+ */
+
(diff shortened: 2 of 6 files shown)


ewienik:
> Add tantivy 0.22 as an optional dependency gated behind the 'fts' cargo feature flag. When compiled without --features fts (the default), the dependency is excluded from the build.

As I recall we were talking about dynamic flag, not static one. Maybe better would be to add configuration env?

knowack1:
> > Add tantivy 0.22 as an optional dependency gated behind the 'fts' cargo feature flag. When compiled without --features fts (the default), the dependency is excluded from the build.
> 
> As I recall we were talking about dynamic flag, not static one. Maybe better would be to add configuration env?

Unfortunately we did not make notes in jira ticket afterward :) We can switch to dynamic config.

crates/vector-store/src/fts.rs
```diff
@@ -0,0 +1,4 @@
+/*
+ * Copyright 2025-present ScyllaDB
```
QuerthDP: 2026

crates/vector-store/src/config_manager.rs
QuerthDP, line 1: > vector-store: add Tantivy dependency and fts feature flag
Add tantivy 0.22 as an optional dependency gated behind the 'fts'
cargo feature flag. When compiled without --features fts (the default),
the dependency is excluded from the build.

Update the commit message

QuerthDP (dismissed):
Looks good.
Although I think the `cargo machete` will fail since we're not using the `tantivy` anywhere.

crates/vector-store/src/fts.rs
knowack1: Fixed

knowack1:
> > > Add tantivy 0.22 as an optional dependency gated behind the 'fts' cargo feature flag. When compiled without --features fts (the default), the dependency is excluded from the build.
> > 
> > 
> > As I recall we were talking about dynamic flag, not static one. Maybe better would be to add configuration env?
> 
> Unfortunately we did not make notes in jira ticket afterward :) We can switch to dynamic config.

Done

knowack1:
> Looks good. Although I think the `cargo machete` will fail since we're not using the `tantivy` anywhere.

Done. Added use in fts.rs.

knowack1:
Delivered in different PR

PR #440, opened 2026-05-07
httproutes: fix similarity_scores returning raw distances instead of scores

nyh:
The refactoring in commit ecb77d5 ("httpapi: refactor vector-store httproutes components") broke the similarity_scores field in the ANN response. The original code converted Distance -> SimilarityScore via the From<Distance> impl, which applies the correct formula per space type (e.g. for Euclidean: 1/(1+d), for Cosine: (2.0-d)/2.0).

The refactored code instead did:

    .map(|distance| (*distance).into())               // Distance -> raw f32
    .map(|distance: f32| SimilarityScore::from(distance)) // wraps f32 as-is
    .map(httpapi::SimilarityScore::from)

The intermediate `.into()` converts Distance to its raw f32 value, and SimilarityScore::from(f32) (derived via derive_more::From) just wraps it without applying any formula. This caused the response to contain raw distances (lower = closer, ascending order for nearest-first results) instead of similarity scores (higher = closer, descending order).

Fix by routing through From<Distance> for SimilarityScore directly:

    .map(|&distance| httpapi::SimilarityScore::from(
                         f32::from(SimilarityScore::from(distance))))

Add regression test similarity_scores_are_decreasing_and_correctly_converted that inserts 3 vectors at known Euclidean distances from the query, then verifies:
  1. similarity_scores are in strictly decreasing order (nearest first).
  2. The Euclidean formula 1/(1+d) is applied: distance 0 -> 1.0, squared distance 1 -> 0.5, squared distance 9 -> 0.1.

Without the fix the test fails with scores [0.0, 1.0, 9.0] (raw squared distances, ascending).

Fixes VECTOR-670.

### DIFF
--- a/crates/vector-store/src/httproutes.rs
+++ b/crates/vector-store/src/httproutes.rs
@@ -637,8 +637,8 @@ async fn post_index_ann(
             } else {
                 let similarity_scores: Vec<httpapi::SimilarityScore> = distances
                     .iter()
-                    .map(|distance| (*distance).into())
-                    .map(|distance: f32| SimilarityScore::from(distance))
+                    .copied()
+                    .map(SimilarityScore::from)
                     .map(httpapi::SimilarityScore::from)
                     .collect();
 
--- a/crates/vector-store/tests/integration/usearch.rs
+++ b/crates/vector-store/tests/integration/usearch.rs
@@ -1528,3 +1528,144 @@ async fn null_vector_is_not_indexed() {
     )
     .await;
 }
+
+// Regression test for the similarity_scores returned by ann(). These need
+// to be similarity scores (decreasing), not distance scores (increasing).
+//
+// This test inserts 3 vectors at different distances from the query and
+// checks that:
+//  1. similarity_scores are in strictly decreasing order (nearest = highest score).
+//  2. For Euclidean distance, the formula similarity = 1/(1+d) is applied
+//     correctly.
+#[tokio::test]
+#[ntest::timeout(10_000)]
+async fn similarity_scores_are_decreasing_and_correctly_converted() {
+    crate::enable_tracing();
+
+    let node_state = vector_store::new_node_state().await;
+    let internals = vector_store::new_internals();
+    let (db_actor, db) = db_basic::new(node_state.clone());
+
+    // Use a 1-D Euclidean index so distances are easy to predict.
+    let index = IndexMetadata {
+        keyspace_name: "vector".into(),
+        table_name: "items".into(),
+        index_name: "ann".into(),
+        target_column: "embedding".into(),
+        index_type: DbIndexType::Global,
+        filtering_columns: Arc::new(Vec::new()),
+        dimensions: NonZeroUsize::new(1).unwrap().into(),
+        connectivity: Connectivity::default(),
+        expansion_add: ExpansionAdd::default(),
+        expansion_search: ExpansionSearch::default(),
+        space_type: SpaceType::Euclidean,
+        version: Uuid::new_v4().into(),
+        quantization: Quantization::default(),
+    };
+
+    db.add_table(
+        index.keyspace_name.clone(),
+        index.table_name.clone(),
+        Table {
+            primary_keys: Arc::new(vec!["pk".into()]),
+            partition_key_count: 1,
+            columns: Arc::new([("pk".into(), NativeType::Int)].into_iter().collect()),
+            dimensions: [(index.target_column.clone(), index.dimensions)]
+                .into_iter()
+                .collect(),
+        },
+    )
+    .unwrap();
+
+    // Three items with 1-D vectors [0.0], [1.0], [3.0].
+    // Query vector is [0.0]. USearch computes squared L2 distance, so the
+    // squared distances are 0, 1, 9 respectively, and expected similarity
+    // scores (using 1/(1+d)) are 1/(1+0)=1.0, 1/(1+1)=0.5, 1/(1+9)=0.1.
+    db.add_index(
+        index.clone(),
+        Some(db_basic::scan_fn([
+            (
+                [CqlValue::Int(1)].into(),
+                Some(vec![0.0_f32].into()),
+                Timestamp::from_unix_timestamp(10),
+            ),
+            (
+                [CqlValue::Int(2)].into(),
+                Some(vec![1.0_f32].into()),
+                Timestamp::from_unix_timestamp(20),
+            ),
+            (
+                [CqlValue::Int(3)].into(),
+                Some(vec![3.0_f32].into()),
+                Timestamp::from_unix_timestamp(30),
+            ),
+        ])),
+        None,
+    )
+    .unwrap();
+
+    let (_, rx) = watch::channel(Arc::new(Config::default()));
+    let index_factory = vector_store::new_index_factory_usearch(rx).unwrap();
+    let (receivers, _senders) = create_config_channels(test_config()).await;
+    let (_server, addr) =
+        vector_store::run(node_state, db_actor, internals, index_factory, receivers)
+            .await
+            .unwrap();
+    let client = HttpClient::new(addr);
+
+    let keyspace_name = index.keyspace_name.clone().into();
+    let index_name = index.index_name.clone().into();
+
+    wait_for(
+        || async {
+            client
+                .index_status(&keyspace_name, &index_name)
+                .await
+                .is_ok_and(|s| s.status == IndexStatus::Serving && s.count == 3)
+        },
+        "Waiting for 3 vectors to be indexed",
+    )
+    .await;
+
+    let (_primary_keys, distances, similarity_scores) = client
+        .ann(
+            &keyspace_name,
+            &index_name,
+            vec![0.0_f32].into(),
+            None,
+            NonZeroUsize::new(3).unwrap().into(),
+        )
+        .await;
+
+    assert_eq!(similarity_scores.len(), 3);
+    assert_eq!(distances.len(), 3);
+
+    let scores: Vec<f32> = similarity_scores
+        .iter()
+        .map(|s| serde_json::to_value(s).unwrap().as_f64().unwrap() as f32)
+        .collect();
+
+    // Scores must be in strictly decreasing order (nearest item has highest similarity).
+    assert!(
+        scores[0] > scores[1] && scores[1] > scores[2],
+        "similarity_scores must be in decreasing order, got: {scores:?}"
+    );
+
+    // Verify the Euclidean similarity formula 1/(1+d) is correctly applied.
+    let epsilon = 1e-5_f32;
+    assert!(
+        (scores[0] - 1.0).abs() < epsilon,
+        "nearest item (distance=0) should have similarity 1.0, got {}",
+        scores[0]
+    );
+    assert!(
+        (scores[1] - 0.5).abs() < epsilon,
+        "second item (distance=1) should have similarity 0.5, got {}",
+        scores[1]
+    );
+    assert!(
+        (scores[2] - 0.1).abs() < epsilon,
+        "farthest item (squared distance=9) should have similarity 0.1, got {}",
+        scores[2]
+    );
+}
--- a/crates/vector-store/src/similarity.rs
+++ b/crates/vector-store/src/similarity.rs
@@ -5,7 +5,7 @@
 
 use crate::Distance;
 
-#[derive(Clone, Debug, derive_more::From, derive_more::Into)]
+#[derive(Clone, Debug, derive_more::Into)]
 /// Similarity score between vectors derived from the distance. Higher score means more similar.
 pub struct SimilarityScore(f32);
 


crates/vector-store/src/httproutes.rs
```diff
@@ -637,9 +637,7 @@ async fn post_index_ann(
             } else {
                 let similarity_scores: Vec<httpapi::SimilarityScore> = distances
                     .iter()
-                    .map(|distance| (*distance).into())
-                    .map(|distance: f32| SimilarityScore::from(distance))
-                    .map(httpapi::SimilarityScore::from)
+                    .map(|&distance| httpapi::SimilarityScore::from(f32::from(SimilarityScore::from(distance))))
```
ewienik: Good catch a mistake in conversion. The conversion should be done in this way: `Distance -> vector-store::SimilarityScore -> httpapi::SimilarityScore`. There shouldn't be f32 in between (as this makes errors) - conversion should be allowed by the types.
To fix correctness consider removing `derive_more::From` for `vector-store::SimilarityScore`, as SimilarityScore shouldn't be build from raw f32, rather only from `Distance`. Then you can fix this similar to this:
```
let similarity_scores = distances
    .iter()                                                     
    .copied()                                                   
    .map(SimilarityScore::from)                                 
    .map(httpapi::SimilarityScore::from)                        
    .collect_vec();                                                 

```

ewienik (changes requested):

crates/vector-store/src/httproutes.rs
nyh: The fix I sent, written by vscode's claude, works, but I cann't really understand this weird syntax (and I guess my subjective feeling that it's weird just got an objective proof - your code looked fine, but wasn't, but nobody understood it enough to understand it's wrong).

So please tell me explicitly what you want me to do. Do you want me to replace the code that Claude wrote by this new snippet from you and test it? What is the new "copied()" thing, by the way?

ewienik: The old code:
```
let similarity_scores: Vec<httpapi::SimilarityScore> = distances // distances is Vec<Distance>, Distance is a type defined in distance.rs
    .iter()                                                      // iterate over elements in vector, Item = &Distance, next operations are done item by item
    .map(|distance| (*distance).into())                          // converting Distance into f32
    .map(|distance: f32| SimilarityScore::from(distance))        // converting from f32 into vector-store::SimilarityScore
    .map(httpapi::SimilarityScore::from)                         // converting from vector-store::SimilarityScore into httpapi::SimilarityScore                       
    .collect();                                                  // collects all items into a vector defined in the first line `let similarity_scores: Vec<httpapi::SimilarityScore>
```
We shouldn't do this conversion f32 into SimilarityScore - it shouldn't be able to compile. To preserve correctness we shouldn't allow such conversion. For this reason we should remove this conversion, which was implemented in macro `derive_more::From` in similarity.rs:

```
#[derive(Clone, Debug, derive_more::From, derive_more::Into)]
/// Similarity score between vectors derived from the distance. Higher score means more similar.
pub struct SimilarityScore(f32);
```

The macro `derive_more::From` implements a trait `impl From<f32> for SimilarityScore` which we shouldn't allow (SimilarityScore must be created only from `Distance` using `impl From<Distance> for SimilarityScore`. For this reason I propose to remove `derive_more::From` macro from SimilarityScore to avoid further issues.

Proposed code:
```
let similarity_scores = distances // we don't define the vector type - it can be inferenced by the compiler
    .iter()                                                     
    .copied()                                                   // as we iterate over reference &Distance, we take copy of it to proceed with value
    .map(SimilarityScore::from)                                 // conversion from Distance to SimilarityScore using defined trait (impl From<Distance> for SimilarityScore)
    .map(httpapi::SimilarityScore::from)                        // conversion from SimilarityScore into httpapi::SimilarityScore using defined trait (impl From<SimilarityScore> for httpapi::SimilarityScore)
    .collect_vec();                                             // collect all iteration items into the Vec<_> - _ is the last type produced by iterator
```

The claude code proposition uses f32 as intermediate artifact, which shouldn't be allowed. And the nit issue is it uses a stack of functions `httpapi::SimilarityScore::from(f32::from(SimilarityScore::from(distance)))` instead of a chain of functions `.map().map()` which is more Rust idiomatic way work with iterators and it seems to me more clear.

nyh: Regardless of everything else, doesn't the buggy code before this patch demonstrate that a lot can be said about this rust chaining idiom, but clear it isn't? :-(

ewienik: I'd say we were missing a test for the correct behavior. And the buggy was the conversion from f32 into SimilariyScore - it shouldn't be implemented at all. It was introduced in the creation of similarity.rs file, wasn't spot on review.  As you can see there are several macro definition in one line - and it is hard to spot additional functionality provided by several items in line.
If someone implements a specific conversion we can trust it, that it is not buggy. If the conversion from f32 into SimilarityScore was working correctly, we wouldn't have a bug in this chain of conversions.

nyh: > I'd say we were missing a test for the correct behavior. 

Right, this is why I asked the AI to write one and included it.

>And the buggy was the conversion from f32 into SimilariyScore - it shouldn't be implemented at all. It was introduced in the creation of similarity.rs file, wasn't spot on review. As you can see there are several macro definition in one line - and it is hard to spot additional functionality provided by several items in line. If someone implements a specific conversion we can trust it, that it is not buggy. If the conversion from f32 into SimilarityScore was working correctly, we wouldn't have a bug in this chain of conversions.

Ok. I'll implement your suggestions and ask you to review again.

nyh:
By the way, @ewienik what the @$@!@ should I do to please the cargo-fmt bot?
As you can guess, I didn't write any of this Rust code myself, so it appears your bot doesn't like the Rust style of my bot :-(

nyh:
@ewienik I implemented your suggestions. Please take another look, and please help me understand what I need to do to please the bots running in this project's CI.

QuerthDP:
> By the way, @ewienik what the @$@!@ should I do to please the cargo-fmt bot? As you can guess, I didn't write any of this Rust code myself, so it appears your bot doesn't like the Rust style of my bot :-(

`cargo fmt --all`

It's a common practice in Rust programming to use those default formatters (best have it run on save).
It should work out of the box.

crates/vector-store/tests/integration/usearch.rs
```diff
@@ -1616,3 +1616,153 @@ async fn null_vector_is_not_indexed() {
...
+// formula in From<Distance> for SimilarityScore. This caused the response to
+// contain raw distances (lower = closer) instead of similarity scores
+// (higher = closer), making scores appear in increasing order for
+// nearest-first results rather than the expected decreasing order.
+//
+// This test inserts 3 vectors at different distances from the query and
+// checks that:
+//  1. similarity_scores are in strictly decreasing order (nearest = highest score).
+//  2. For Euclidean distance, the formula similarity = 1/(1+d) is applied
+//     correctly, so an item at distance 0 (identical vector) gets score 1.0.
```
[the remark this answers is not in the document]
QuerthDP: Copilot is right here. Please add the timeout macro.

crates/vector-store/tests/integration/usearch.rs
```diff
@@ -1616,3 +1616,153 @@ async fn null_vector_is_not_indexed() {
...
+        scores[1]
+    );
+    // Farthest has squared distance 9 -> similarity 0.1.
+    assert!(
+        (scores[2] - 0.1).abs() < epsilon,
+        "farthest item (squared distance=9) should have similarity 0.1, got {}",
+        scores[2]
+    );
+
+    drop(server);
```
QuerthDP: No need for that. Just initialize the `server` as `_server` since it is not used anywhere here.

crates/vector-store/tests/integration/usearch.rs
```diff
@@ -1616,3 +1616,153 @@ async fn null_vector_is_not_indexed() {
...
+
+    // Scores must be in strictly decreasing order (nearest item has highest similarity).
+    assert!(
+        scores[0] > scores[1] && scores[1] > scores[2],
+        "similarity_scores must be in decreasing order, got: {scores:?}"
+    );
+
+    // Verify the Euclidean similarity formula 1/(1+d) is correctly applied.
+    // The nearest item has distance 0 -> similarity 1.0.
+    let epsilon = 1e-5_f32;
```
QuerthDP, line 1655: Do we need the epsilon check? Are the expected values not reached exactly?

QuerthDP (dismissed):
Please add the `ntest::timeout` to the test. Apart from that it looks fine.

ewienik:
> @ewienik I implemented your suggestions. Please take another look, and please help me understand what I need to do to please the bots running in this project's CI.

There is format linter error. As @QuerthDP wrote, you need to run `cargo fmt --all` to fix all formatting issues.

ewienik (dismissed):

crates/vector-store/tests/integration/usearch.rs
nyh: I didn't check (AI wrote this code), but it makes sense that 1/9 et al. can't be represented exactly in floating point, so we can't expect exact equality.

crates/vector-store/tests/integration/usearch.rs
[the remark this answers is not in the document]
nyh: Oops, missed this comment. I'll do it.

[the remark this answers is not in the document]
nyh: By the way, I don't know why you guys put up with all these "rules". If there's a rule that every test should timeout after 10,000 seconds (or is it 10 seconds? or what?), why does each test need to specify it separately? Why is it not the default of the test framework? That's just ugly :-(

nyh:
> > By the way, @ewienik what the @$@!@ should I do to please the cargo-fmt bot? As you can guess, I didn't write any of this Rust code myself, so it appears your bot doesn't like the Rust style of my bot :-(
> 
> `cargo fmt --all`

Thanks. I didn't even have this garbage installed ;-)
It changed the indentation of one line, in a completely unnecessary way. What a shame.
I'll push the fix.

> It's a common practice in Rust programming to use those default formatters (best have it run on save). It should work out of the box.

I'm still waiting that someone one day explains to me who invented this useless practice that started in Python and now expanded to Rust. This practice never made any sense, it makes even less sense in the age of AI, which doesn't get all OCDed if you break one of 1000 indentation rules - and can understand anything ;-)

crates/vector-store/tests/integration/usearch.rs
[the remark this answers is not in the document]
ewienik: ntest::timeout takes a duration in milliseconds, so this is 10s. All tests can have one limit using specific option (see `cargo test -- --help`), but sometime it seems better to have timeout defined in a specific test (as mostly we run simply `cargo t` without env variables:
```
--ensure-time   Treat excess of the test execution time limit as
                error.
                Threshold values for this option can be configured via
                `RUST_TEST_TIME_UNIT`, `RUST_TEST_TIME_INTEGRATION`
                and
                `RUST_TEST_TIME_DOCTEST` environment variables.
                Expected format of environment variable is
                `VARIABLE=WARN_TIME,CRITICAL_TIME`.
                `CRITICAL_TIME` here means the limit that should not
                be exceeded by test.
```

ewienik (dismissed):

crates/vector-store/tests/integration/usearch.rs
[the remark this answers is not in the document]
nyh: Maybe it makes sense to have a certain test that has a special timeout, but this is not the case, there's nothing particularly slow or fast about this test - it's just normal. Having to remember to add this line on a thousand tests - and asking each test author to choose a number - should it be 10 seconds or 5 seconds or 20 seconds? - doesn't sound great to me.

QuerthDP:
By the way @nyh I assume you're not familiar with that, but in case of this repository we require all PR comments to be resolved to add to the merge queue. I can do it for you since most of them should be resolved, yet there is still one of my comments (https://github.com/scylladb/vector-store/pull/440#discussion_r3201198947) which was not responded to at all.
It does not matter much, it's more a style nit pick (we rather don't drop objects explicitly if not required, the `server` would be dropped automatically by Rust at the end of the scope). Apart from that, you need to rename it to use underscore (`_server`) so that it is marked as unused, which would satisfy the `cargo clippy`. 

I can merge it as is, but I want to know if you consciously ignored the comment.

nyh:
> By the way @nyh I assume you're not familiar with that, but in case of this repository we require all PR comments to be resolved to add to the merge queue. I can do it for you since most of them should be resolved, yet there is still one of my comments ([#440 (comment)](https://github.com/scylladb/vector-store/pull/440#discussion_r3201198947)) which was not responded to at all. It does not matter much, it's more a style nit pick (we rather don't drop objects explicitly if not required, the `server` would be dropped automatically by Rust at the end of the scope). Apart from that, you need to rename it to use underscore (`_server`) so that it is marked as unused, which would satisfy the `cargo clippy`.
> 
> I can merge it as is, but I want to know if you consciously ignored the comment.

No, it was not intentional, I'll take a look again at the comments on what I missed.

crates/vector-store/tests/integration/usearch.rs
nyh: Done. I had to ask the AI to explain to me what this comment even meant, but now I know :-)
I'll push a new version with this fix.

nyh:
> By the way @nyh I assume you're not familiar with that, but in case of this repository we require all PR comments to be resolved to add to the merge queue. I can do it for you since most of them should be resolved, yet there is still one of my comments ([#440 (comment)](https://github.com/scylladb/vector-store/pull/440#discussion_r3201198947)) which was not responded to at all.

I answered your question and pushed a new version. But I don't understand your comment about resolving - is it your practice that the author of the code "resolves" reviewer's issues - or that the reviewers must come back and resolve the issues they opened if they are please with the answer?

nyh:
Github declared "[nyh](https://github.com/nyh) dismissed [ewienik](https://github.com/ewienik)’s [stale review](https://github.com/scylladb/vector-store/pull/440#pullrequestreview-4244349380)". I certainly did no such thing - I don't think @ewienik's review became "stale" because I fixed some silly style issues that don't make any real difference.

Please consider merging this PR, it fixes a real bug, a regression.

QuerthDP:
> > By the way @nyh I assume you're not familiar with that, but in case of this repository we require all PR comments to be resolved to add to the merge queue. I can do it for you since most of them should be resolved, yet there is still one of my comments ([#440 (comment)](https://github.com/scylladb/vector-store/pull/440#discussion_r3201198947)) which was not responded to at all.
> 
> I answered your question and pushed a new version. But I don't understand your comment about resolving - is it your practice that the author of the code "resolves" reviewer's issues - or that the reviewers must come back and resolve the issues they opened if they are please with the answer?

My personal habit is that I respond to the reviewer's comment and resolve it if it does not need any follow ups.
I think it's similar to other developers of Vector Store.

QuerthDP (approved):
Thanks for the fix!

ewienik:
> > > By the way @nyh I assume you're not familiar with that, but in case of this repository we require all PR comments to be resolved to add to the merge queue. I can do it for you since most of them should be resolved, yet there is still one of my comments ([#440 (comment)](https://github.com/scylladb/vector-store/pull/440#discussion_r3201198947)) which was not responded to at all.
> > 
> > 
> > I answered your question and pushed a new version. But I don't understand your comment about resolving - is it your practice that the author of the code "resolves" reviewer's issues - or that the reviewers must come back and resolve the issues they opened if they are please with the answer?
> 
> My personal habit is that I respond to the reviewer's comment and resolve it if it does not need any follow ups. I think it's similar to other developers of Vector Store.
I am using it differently :-) I put my answer to the reviewer's comment showing if this is fixed or if I don't follow suggestion - describing why. I leave comment opened for the reviewer. Only in case the reviewer accepted the PR but comments are still open (and answered by me) I am closing them before merge. So this is not a strict process.

PR #443, opened 2026-05-08
vector-store: fix ALTER INDEX simulator

ewienik:
There is an issue with simulation ALTER INDEX when the new index is created with changed metadata (it is observed with change quantization). The root cause is with function checking if new index should be added, if there are changes within index metadata beside version. This commit fixes it.

The commit introduces also a new function - discovered_indexes_simulator for possibility to test the behavior with unitests.

Fixes: VECTOR-669

### DIFF
--- a/crates/vector-store/src/monitor_indexes.rs
+++ b/crates/vector-store/src/monitor_indexes.rs
@@ -81,20 +81,9 @@ pub(crate) async fn new(
                         };
 
                         if alter_index_simulator {
-                            node_state.send_event(
-                                Event::IndexesDiscovered(
-                                    indexes
-                                        .iter()
-                                        .filter(|idx| !should_delete_simulator(idx, &new_indexes))
-                                        .chain(
-                                            new_indexes
-                                                .iter()
-                                                .filter(|idx| should_add_simulator(idx, &indexes))
-                                        )
-                                        .cloned()
-                                        .collect()
-                                )
-                            ).await;
+                            node_state.send_event(Event::IndexesDiscovered(
+                                discovered_indexes_simulator(&indexes, &new_indexes)
+                            )).await;
                         } else {
                             node_state.send_event(
                                 Event::IndexesDiscovered(new_indexes.clone()),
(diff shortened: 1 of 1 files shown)


crates/vector-store/src/monitor_indexes.rs
```diff
@@ -299,9 +288,27 @@ fn should_delete_simulator(curr_idx: &IndexMetadata, new_indexes: &HashSet<Index
...
+    curr_indexes: &HashSet<IndexMetadata>,
+    new_indexes: &HashSet<IndexMetadata>,
+) -> HashSet<IndexMetadata> {
+    dbg!(curr_indexes)
+        .iter()
+        .filter(|idx| dbg!(!should_delete_simulator(idx, new_indexes)))
+        .chain(
+            dbg!(new_indexes)
+                .iter()
+                .filter(|idx| dbg!(should_add_simulator(idx, curr_indexes))),
```
QuerthDP: Should we use `dbg!` in production code?

crates/vector-store/src/monitor_indexes.rs
```diff
@@ -813,6 +820,14 @@ mod tests {
             },
             &[idx.clone()].into_iter().collect()
         ));
+        assert!(should_delete_simulator(
+            &IndexMetadata {
+                quantization: Quantization::I8,
```
QuerthDP, line 825: Is the issue specific to Quantization?

crates/vector-store/src/monitor_indexes.rs
ewienik: no, leftover, will remove it

crates/vector-store/src/monitor_indexes.rs
ewienik: Not specific to quantization, but it reflects the use case when I wont to test different quantization on the same data.

ewienik:
Changelog for a98a2d53
- removed dbg! from code

<details>
<summary>range diff</summary>

```
1:  b576d59f ! 1:  a98a2d53 vector-store: fix ALTER INDEX simulator
    @@ crates/vector-store/src/monitor_indexes.rs: fn should_delete_simulator(curr_idx:
     +    curr_indexes: &HashSet<IndexMetadata>,
     +    new_indexes: &HashSet<IndexMetadata>,
     +) -> HashSet<IndexMetadata> {
    -+    dbg!(curr_indexes)
    ++    curr_indexes
     +        .iter()
    -+        .filter(|idx| dbg!(!should_delete_simulator(idx, new_indexes)))
    ++        .filter(|idx| !should_delete_simulator(idx, new_indexes))
     +        .chain(
    -+            dbg!(new_indexes)
    ++            new_indexes
     +                .iter()
    -+                .filter(|idx| dbg!(should_add_simulator(idx, curr_indexes))),
    ++                .filter(|idx| should_add_simulator(idx, curr_indexes)),
     +        )
     +        .cloned()
     +        .collect()
```

crates/vector-store/src/monitor_indexes.rs
ewienik: fixed in a98a2d53

crates/vector-store/src/monitor_indexes.rs
QuerthDP: But the change to quantization implies we need to rebuild the index, don't we? Wasn't the whole point of introducing the ALTER INDEX workaround not to rebuilt the index?

ewienik: Yes, not rebuilt the index when not needed - change in rescoring, oversampling shouldn't trigger rebuild. In case of change quantization, we need to rebuild the index - so this is expected behavior.

QuerthDP (approved):

QuerthDP:
Note for future readers as I didn't found it trivial and described enough at first glance:

Before this change - if we wanted to add the index with different metadata, but identical `key()`, `should_add_simulator()` would return false and not proceed with addition.
After the change - we check if the metadata (excluding version) is identical to already existing index, if not `should_add_simulator()` will return true and proceed with addition.

PR #444, opened 2026-05-09
support storing, filtering and retrieving non-key columns

nyh:
This series completes the filtering-column feature in the vector store:
1. Enabling the vector-store to store - in memory - additional non-key columns, not just the key columns and vectors that we store today. This should support both CQL columns and Alternator attributes.
2. Allowing ANN queries to pre-filter on these stored non-key column values, not just on the key columns.
3. Also _optionally_ retrieve some or all of these stored columns with search results.

Alternator vector search will use all three features. CQL's vector search can't yet use the third feature, but in the future it could: If the user's vector search only SELECTs columns which are stored in the vector index, then the entire request can be satisfied from the vector index (using the third feature) and CQL doesn't need to perform additional slow base-table queries.

The vector store already had the data-model bones for this feature: "Table" stored typed per-column values, indexes.rs tracked which columns an index declared as filterable, and the routing logic could prefer an
index whose filtering columns covered all restrictions in a query. What was missing was the actual plumbing: fetching the column values from ScyllaDB during backfill, extracting them from CDC change rows, forwarding
them through DbEmbedding, and storing them into "Table". Without that plumbing the filtering-column slots always contained the sentinel "not-present" value, so every restriction on a non-primary-key column
produced no results.

**Patch 1 — support storing and filtering on Alternator scalar attributes**

Wires up the full pipeline, motivated by Alternator's Projection=INCLUDE for vector indexes (NonKeyAttributes used as KeyConditionExpression pre-filters).

For Alternator, all non-key attributes live in a single `:attrs map<blob,blob>` column with no per-attribute type schema. The two common scalar encodings (S = 0x00 tag + UTF-8 bytes, N = 0x03 tag + CQL decimal) are stored verbatim as CqlValue::Blob so the type tag is preserved for comparison. cql_cmp gains two cross-type arms
(Blob, Text) and (Blob, Decimal) that decode the tag at comparison time, and httproutes.rs maps JSON string/number filter values to Text/Decimal for the HTTP API side.

The CQL path was already partially in place (the scan query and CDC consumer already selected the typed columns directly), but the values were never forwarded into Table::add(). That gap is also closed here as a side-effect.

**Patch 2 — validator: test ANN filtering on a CQL non-primary-key column**

Adds an end-to-end validator test that exercises the newly complete CQL path. A table with `pk INT PRIMARY KEY, category INT, v VECTOR<FLOAT,3>` gets a vector index with category declared as a filtering column. Six rows with alternating category 0/1 are inserted; a `WHERE category = 0` ANN query must return exactly three rows. Verified to pass against a real ScyllaDB node.

**Patch 3 - ann: optionally return stored column values alongside primary keys**
ann() is given the ability to optionally return some of the stored columns together with the item keys which are always returned. Alternator uses this to allow a search that only needs the stored columns to complete without the slow extra step of base table lookup per result. CQL doesn't use it yet, but in the future could when a SELECT only asks for the keys and filtering columns.

**Patch 4 - cdc: fix incorrect vector deletion on Alternator partial updates**
Fix a pre-existing bug where we incorrectly assumed that if an Alternator attribute  - such as the vector column itself - was missing in an update, it meant it was deleted (and when the vector column is deleted, the entire item is deleted). That assumption was not true - CDC gives only deltas, so a certain change might be to some unrelated
attribute and not change the vector column at all. We need to correctly distinguish between elements of ":attr" that were really deleted in the operation - and elements which are missing from the delta because they were really deleted. Luckily (it's not luck, obviously), CDC gives us exactly this information, which we forgot to use.

**Patch 5 - fix a bug in try_to_json() which could lead to crash if an unsupported CQL type was used as a filtering column.**

### DIFF
--- a/api/openapi.json
+++ b/api/openapi.json
@@ -407,6 +407,13 @@
           "limit": {
             "$ref": "#/components/schemas/Limit"
           },
+          "return_columns": {
+            "type": "array",
+            "items": {
+              "$ref": "#/components/schemas/ColumnName"
+            },
+            "description": "Filtering-column names whose stored values should be returned alongside\nthe primary keys.  Empty (the default) means return no column values."
+          },
           "vector": {
             "$ref": "#/components/schemas/Vector"
           }
@@ -420,6 +427,18 @@
           "similarity_scores"
         ],
         "properties": {
+          "column_values": {
+            "type": "object",
+            "description": "Per-column stored values for the filtering columns requested in\n`return_columns`.  Each entry maps a column name to a Vec of\n`Option<Value>` — one entry per returned nearest neighbour, in the\nsame order as `similarity_scores`.  `None` means the value was not\npresent for that row (e.g. the attribute did not exist when the row\nwas indexed).  Absent when `return_columns` was empty.",
+            "additionalProperties": {
+              "type": "array",
+              "items": {}
+            },
+            "propertyNames": {
+              "type": "string",
+              "description": "Name of the column in a db table."
+            }
+          },
           "distances": {
             "type": "array",
             "items": {
--- a/crates/validator/src/cdc.rs
+++ b/crates/validator/src/cdc.rs
@@ -51,6 +51,11 @@ pub(crate) async fn new() -> TestCase<TestActors> {
             timeout,
             cql_per_row_ttl_expires_from_index,
         )
+        .with_test(
+            "cql_explicit_vector_column_deletion_removes_from_index",
+            timeout,
+            cql_explicit_vector_column_deletion_removes_from_index,
+        )
 }
 
 #[framed]
@@ -656,6 +661,97 @@ async fn cql_per_row_ttl_expires_from_index(actors: TestActors) {
         .query_unpaged(format!("DROP KEYSPACE {keyspace}"), ())
         .await
         .expect("failed to drop a keyspace");
+
+    info!("finished");
+}
+
+/// Regression test: explicitly deleting the vector column (`DELETE v FROM t WHERE pk = ?`)
+/// must remove the item from the vector index, not leave it as "unchanged".
+///
+/// Previously the CQL CDC path only treated full row/partition deletions as
+/// removals.  An explicit column tombstone produced `embedding = None`
+/// ("unchanged") instead of `Some(None)` ("delete from index"), so the stale
+/// vector remained searchable.
+#[framed]
+async fn cql_explicit_vector_column_deletion_removes_from_index(actors: TestActors) {
+    info!("started");
+
+    let (session, clients) = prepare_connection_single_vs(&actors).await;
+    let client = clients.first().unwrap();
+    let keyspace = create_keyspace(&session).await;
+    let table = create_table(&session, "pk INT PRIMARY KEY, v VECTOR<FLOAT, 3>", None).await;
+
+    session
+        .query_unpaged(
+            format!("INSERT INTO {table} (pk, v) VALUES (1, [1.0, 0.0, 0.0])"),
+            (),
+        )
+        .await
+        .expect("failed to insert data");
+    session
+        .query_unpaged(
+            format!("INSERT INTO {table} (pk, v) VALUES (2, [0.0, 1.0, 0.0])"),
+            (),
+        )
+        .await
+        .expect("failed to insert data");
+
+    let index = create_index(CreateIndexQuery::new(&session, &clients, &table, "v")).await;
+
+    let status = wait_for_index(client, &index).await;
+    assert_eq!(
+        status.count, 2,
+        "Index should have 2 vectors after full scan"
+    );
+
+    // Explicitly delete only the vector column of pk=1 (not the whole row).
+    // This is a column tombstone, not a row deletion — the CQL CDC path
+    // must recognise it via is_value_deleted() and remove the vector from the index.
+    session
+        .query_unpaged(format!("DELETE v FROM {table} WHERE pk = 1"), ())
+        .await
+        .expect("failed to delete vector column");
+
+    // Wait for the index count to drop to 1.
+    wait_for(
+        || async {
+            let status = client.index_status(&index.keyspace, &index.index).await;
+            matches!(status, Ok(s) if s.count == 1)
+        },
+        "Waiting for index count to drop to 1 after explicit column deletion",
+        FINE_GRAINED_CDC_MAX_LATENCY,
+    )
+    .await;
+
+    // ANN query should only return pk=2.
+    let result = wait_for_value(
+        || async {
+            let result = get_opt_query_results(
+                format!("SELECT pk FROM {table} ORDER BY v ANN OF [0.0, 1.0, 0.0] LIMIT 10"),
+                &session,
+            )
+            .await;
+            result.filter(|r| r.rows_num() == 1)
+        },
+        "Waiting for ANN query to return only pk=2",
+        FINE_GRAINED_CDC_MAX_LATENCY,
+    )
+    .await;
+    let pks: Vec<i32> = result
+        .rows::<(i32,)>()
+        .expect("failed to get rows")
+        .map(|r| r.expect("row error").0)
+        .collect();
+    assert_eq!(
+        pks,
+        vec![2],
+        "Only pk=2 should remain after explicit vector column deletion"
+    );
+
+    session
+        .query_unpaged(format!("DROP KEYSPACE {keyspace}"), ())
+        .await
+        .expect("failed to drop keyspace");
 
     info!("finished");
 }
--- a/crates/validator/src/filtering.rs
+++ b/crates/validator/src/filtering.rs
@@ -100,6 +100,11 @@ pub(crate) async fn new() -> TestCase<TestActors> {
             timeout,
             local_ann_with_timestamp_gte_filter,
         )
+        .with_test(
+            "ann_filter_by_non_pk_filtering_column",
+            timeout,
+            ann_filter_by_non_pk_filtering_column,
+        )
 }
 
 /// Test ANN search filtered by partition key equality.
@@ -1035,6 +1040,87 @@ async fn local_ann_with_timestamp_gte_filter(actors: TestActors) {
         2,
         "Expected two rows with created_at >= 2024-01-01"
     );
+
+    session
+        .query_unpaged(format!("DROP KEYSPACE {keyspace}"), ())
+        .await
+        .expect("failed to drop a keyspace");
+
+    info!("finished");
+}
+
+/// Test ANN search filtered by a non-primary-key CQL filtering column.
+///
+/// The index is created with `category` as a filtering column.  Rows have two
+/// category values (0 and 1).  Querying with `WHERE category = 0` must return
+/// only the rows whose category is 0.
+#[framed]
+async fn ann_filter_by_non_pk_filtering_column(actors: TestActors) {
+    info!("started");
+
+    let (session, clients) = prepare_connection(&actors).await;
+
+    let keyspace = create_keyspace(&session).await;
+    let table = create_table(
+        &session,
+        "pk INT PRIMARY KEY, category INT, v VECTOR<FLOAT, 3>",
+        None,
+    )
+    .await;
+
+    // Insert 6 rows: pk 0..5, alternating category 0 and 1.
+    for pk in 0..6_i32 {
+        session
+            .query_unpaged(
+                format!("INSERT INTO {table} (pk, category, v) VALUES (?, ?, ?)"),
+                (pk, pk % 2, &vec![pk as f32, 0.0, 0.0]),
+            )
+            .await
+            .expect("failed to insert data");
+    }
+
+    let index = create_index(
+        CreateIndexQuery::new(&session, &clients, &table, "v").filter_columns(["category"]),
+    )
+    .await;
+
+    for client in &clients {
+        let index_status = wait_for_index(client, &index).await;
+        assert_eq!(index_status.count, 6, "Expected 6 vectors to be indexed");
+    }
+
+    let result = wait_for_value(
+        || async {
+            let result = get_opt_query_results(
+                format!(
+                    "SELECT pk, category FROM {table} \
+                     WHERE category = 0 \
+                     ORDER BY v ANN OF [0.0, 0.0, 0.0] LIMIT 10 \
+                     ALLOW FILTERING"
+                ),
+                &session,
+            )
+            .await;
+            result.filter(|r| r.rows_num() == 3)
+        },
+        "Waiting for category=0 filtered ANN query to return 3 rows",
+        DEFAULT_OPERATION_TIMEOUT,
+    )
+    .await;
+
+    let rows: Vec<(i32, i32)> = result
+        .rows::<(i32, i32)>()
+        .expect("failed to get rows")
+        .map(|row| row.expect("failed to get row"))
+        .collect();
+
+    assert_eq!(rows.len(), 3, "Expected exactly 3 rows with category=0");
+    for (pk, category) in &rows {
+        assert_eq!(
+            *category, 0,
+            "Expected all rows to have category=0, got pk={pk} category={category}"
+        );
+    }
 
     session
         .query_unpaged(format!("DROP KEYSPACE {keyspace}"), ())
--- a/crates/vector-store/benches/pipeline.rs
+++ b/crates/vector-store/benches/pipeline.rs
@@ -265,7 +265,7 @@ fn scan_fn_mpsc(
                     .send((
                         DbEmbedding {
                             primary_key,
-                            embedding,
+                            embedding: Some(embedding),
                             timestamp,
                             column_values: Default::default(),
                         },
--- a/crates/vector-store/benches/pipeline.rs
+++ b/crates/vector-store/benches/pipeline.rs
@@ -267,6 +267,7 @@ fn scan_fn_mpsc(
                             primary_key,
                             embedding,
                             timestamp,
+                            column_values: Default::default(),
                         },
                         in_progress,
                     ))
@@ -1139,7 +1140,7 @@ where
             })
         })
         .collect_vec();
-    stream::iter(tasks.into_iter())
+    stream::iter(tasks)
         .then(|task| async move { task.await.unwrap() })
         .fold(Duration::ZERO, |acc, x| async move { acc + x })
         .await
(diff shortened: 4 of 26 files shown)


nyh:
Notes to reviewers:
1. The first patch comes from an earlier pull request that Alternator needs - it will be removed in a rebase when it goes in from that other pull request. **(DONE)**
2. This patch to the vector store (or something resembling it) is needed for allowing Alternator to "project" additional non-key columns to the index (this is analogous to the CQL "filtering columns" index-schema feature). The Alternator feature is in https://github.com/scylladb/scylladb/pull/29959.

nyh:
While writing this patch, I discovered that the vector store is missing two very important features regarding "projected columns" (a.k.a. filtering columns):
1. Although the code can store a list of these columns, as far as I can tell there was no code to actually use these columns in a filter. I think the code I added now for Alternator filtering is the first implementation in the vector store?
2. Additionally, there doesn't seem any way to ask the vector store to return the stored projected column values for matching columns. This is important for Alternator which wants to be able to fetch these projected columns efficiently - without the slow extra step of reading all the matching items from the base table.

This patch is currently implementing 1, but not yet 2. I - or somebody else - will need to implement 2 as well, to completely support what Alternator needs. Right now the Alternator feature (https://github.com/scylladb/scylladb/pull/29776) has an xfailing test because of this missing feature. **(DONE BELOW)**

nyh:
I'll push a new version fixing fmt and clippy errors. This is so sad for me to see these things. "fmt" didn't like the way the AI **nicely** formatted some lines (I have no idea what rule was even violated, it looked perfectly nice), and "clippy" didn't like a nested if and insisted that it must be rewritten with "&&". What nonsense :-(

nyh:
Clippy still doesn't accept this patch, and reports issues in code NOT included in this patch.
Despite my better judement I'll include changes for those unrelated style issues :-(

nyh:
Pushed a new version to fix a pre-existing test that started failing. The new paragraph from the commit message:

    Two existing integration tests in `routing.rs` that previously expected
    `BAD_REQUEST` for filter-column restrictions (with a TODO to update them
    once end-to-end support landed) are updated to expect `OK` now that this
    commit provides that support.

crates/vector-store/src/httproutes.rs
```diff
@@ -708,14 +711,29 @@ fn try_from_post_index_ann_filter(
...
+        if matches!(native_type, NativeType::Blob) {
+            return match value {
+                Value::String(s) => Ok(CqlValue::Text(s)),
+                Value::Number(_) => try_from_json(value, &NativeType::Decimal),
+                _ => bail!(
+                    "Column '{column}' is an Alternator attribute column; \
+                     only string (S-type) and number (N-type) filter values are supported"
+                ),
+            };
+        }
```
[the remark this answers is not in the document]
nyh: Good catch. I'll fix this. As far as I understand, this bug would not be reproducible (so I can't add a test), because the code involved only applies to filtering using non-key columns, and currently this feature isn't tested for CQL.

crates/vector-store/src/table.rs
```diff
@@ -1151,6 +1155,12 @@ impl TableAdd for Table {
                     }
                     Ok(())
                 })?;
+                // Update filtering column values for this row (best-effort, no timestamp guard).
+                for (col_name, val) in column_values {
+                    if let Some(col) = self.columns.get_mut(&col_name) {
+                        let _ = col.insert_cqlvalue(primary_id, db_embedding.timestamp, val);
+                    }
```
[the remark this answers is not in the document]
nyh: Done.

nyh:
@ewienik @QuerthDP I called you to review this PR. It started as adding the features I needed for saving non-key filtering columns in Alternator (known as "projected columns" there), but then I realized - this is exactly what is also needed for CQL to allow filtering on non-key columns. And indeed, my PR also fixes CQL :-) I wonder what you think about that - you actually had a feature in CQL that was **almost working**, but change needed the small changes here to fully work.

Still missing is the feature to **return** the projected columns in ANN requests, along with the primary key returned. I need this feature in Alternator, so I'll add this next.

nyh:
> Still missing is the feature to **return** the projected columns in ANN requests, along with the primary key returned. I need this feature in Alternator, so I'll add this next.

Another AI session, and this missing feature is also done, as the last third patch. ann() now has the ability to return some of the stored columns - if requested - together with the item keys always returned. Alternator uses this to allow a search that only needs the stored columns to complete without the slow extra step of base table lookup per result.

crates/vector-store/src/httproutes.rs
```diff
@@ -708,14 +711,29 @@ fn try_from_post_index_ann_filter(
...
         try_from_json(value, native_type)
```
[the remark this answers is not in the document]
nyh: Done.

crates/vector-store/src/table.rs
```diff
@@ -1151,6 +1185,24 @@ impl TableAdd for Table {
...
+                        {
+                            continue;
+                        }
+                        if let Err(err) =
+                            col.insert_cqlvalue(primary_id, db_embedding.timestamp, val)
+                        {
+                            warn!("Failed to update filtering column '{col_name}': {err}");
+                        }
+                    }
+                }
```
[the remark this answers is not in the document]
nyh: Very good point. Will fix and add tests.

crates/vector-store/src/httproutes.rs
```diff
@@ -677,15 +692,35 @@ async fn post_index_ann(
...
+                        > = return_columns
+                            .iter()
+                            .map(|col_name| {
+                                let values = column_values_per_row
+                                    .iter()
+                                    .map(|row_vals| {
+                                        row_vals
+                                            .get(col_name)
+                                            .and_then(|v| try_to_json(v.clone()).ok())
+                                    })
```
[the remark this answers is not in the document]
nyh: I'll do that, although I'm not convinced that this improvement is directly related to this patch.

crates/vector-store/src/db_cdc.rs
```diff
@@ -507,15 +508,143 @@ impl Consumer for CdcConsumer {
...
+                // with `Some(None)` rather than falling through to the "unchanged" path.
+                let emb = if is_deletion_op {
+                    Some(None)
+                } else {
+                    raw_vector_value
+                        .map(|v| source.extract_vector(v))
+                        .transpose()?
+                        .flatten()
+                        .map(Some)
+                };
```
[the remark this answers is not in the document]
nyh: Good catch.

crates/vector-store/src/db.rs
```diff
@@ -991,6 +992,11 @@ fn from_target_option(
     };
 
     let validate_target_type = |target_name: &str| -> anyhow::Result<()> {
+        // For Alternator keyspaces, the vector attribute is stored inside the
+        // `:attrs` map and is not a real CQL column, so skip the type check.
+        if keyspace_name.starts_with("alternator_") {
```
knowack1: there is `db_index_backend` which is suppose to keep all the differences between alternator vs non-alternator index. Prefer to extract to `db_index_backend::validate_target_type`.

knowack1: Please check `keyspace_name.is_alternator()`.

crates/vector-store/src/db_cdc.rs
```diff
@@ -507,12 +507,41 @@ impl Consumer for CdcConsumer {
...
+            bail!("CDC error: column {vector_col} should be deletable");
         }
-        let embedding = row
-            .take_value(column)
+
+        // For Alternator, the whole `:attrs` map is the vector column.  Clone it so we can
+        // also extract filtering column values from it in a second pass.
+        let raw_vector_value = row.take_value(vector_col);
+
+        let column_values = match source {
```
knowack1: Such "ifs" should be places in db_index_backend. E.g. `db_index_backend::get_column_values(...)`.

crates/vector-store/src/db_index.rs
```diff
@@ -586,6 +630,31 @@ impl Statements {
                     return None;
                 }
 
+                // Pop extra filtering columns (appended after writetime) in reverse order.
+                let mut column_values = std::collections::BTreeMap::new();
+                for col_name in scan_filtering_columns.iter().rev() {
+                    let raw = row.columns.pop().unwrap();
+                    if let Some(raw_val) = raw {
+                        let scalar = if is_alternator {
```
knowack1: Also here. We should avoid spreading such ifs around the code and keep all of them in a single place: `db_index_backend`.

crates/vector-store/src/vector.rs
```diff
@@ -350,4 +410,211 @@ mod tests {
...
+    #[test]
+    fn extract_scalar_unknown_tag_returns_none() {
+        assert_eq!(extract_alternator_scalar(&[0x99, b'x']), None);
+    }
+
+    // --- extract_alternator_scalars ---
+
+    #[test]
+    fn extract_scalars_non_map_returns_empty() {
+        let result = extract_alternator_scalars(&CqlValue::Int(42), &["col".into()]);
```
knowack1, line 484: Looks like this code does not compile in this commit.

crates/validator/src/filtering.rs
```diff
@@ -1025,3 +1030,84 @@ async fn local_ann_with_timestamp_gte_filter(actors: TestActors) {
...
     info!("finished");
 }
+
+/// Test ANN search filtered by a non-primary-key CQL filtering column.
+///
+/// The index is created with `category` as a filtering column.  Rows have two
+/// category values (0 and 1).  Querying with `WHERE category = 0` must return
+/// only the rows whose category is 0.
+#[framed]
+async fn ann_filter_by_non_pk_filtering_column(actors: TestActors) {
```
knowack1, line 1058: Should we have the same test for alternator API?

crates/validator/src/filtering.rs
```diff
@@ -1025,3 +1030,84 @@ async fn local_ann_with_timestamp_gte_filter(actors: TestActors) {
 
     info!("finished");
 }
+
```
knowack1, line 1051: It seems that this and the previous commit close VECTOR-419. This should be mentioned in the cover letter.
Preferably, I would split this PR into two:
- PR 1: Commits 1–2 // Support for filtering on non-PK columns
- PR 2: Commits 3–5 // Extended HTTP API

api/openapi.json
```diff
@@ -420,6 +427,18 @@
           "similarity_scores"
         ],
         "properties": {
+          "column_values": {
+            "type": "object",
+            "description": "Per-column stored values for the filtering columns requested in\n`return_columns`.  Each entry maps a column name to a Vec of\n`Option<Value>` — one entry per returned nearest neighbour, in the\nsame order as `similarity_scores`.  `None` means the value was not\npresent for that row (e.g. the attribute did not exist when the row\nwas indexed).  Absent when `return_columns` was empty.",
```
knowack1, line 432: "Vec of `Option<Value>`" - this looks like some rust specifics, while this is description of HTTP API.

knowack1, line 432: What is "None" in json?

crates/validator/src/cdc.rs
```diff
@@ -659,3 +664,94 @@ async fn cql_per_row_ttl_expires_from_index(actors: TestActors) {
...
+
+    // Explicitly delete only the vector column of pk=1 (not the whole row).
+    // This is a column tombstone, not a row deletion — the CQL CDC path
+    // must recognise it via is_value_deleted() and remove the vector from the index.
+    session
+        .query_unpaged(format!("DELETE v FROM {table} WHERE pk = 1"), ())
+        .await
+        .expect("failed to delete vector column");
+
+    // Wait for the index count to drop to 1.
```
knowack1, line 715: This wait is not needed while we have later:
// ANN query should only return pk=2.

crates/validator/src/cdc.rs
```diff
@@ -659,3 +664,94 @@ async fn cql_per_row_ttl_expires_from_index(actors: TestActors) {
...
+        .query_unpaged(
+            format!("INSERT INTO {table} (pk, v) VALUES (2, [0.0, 1.0, 0.0])"),
+            (),
+        )
+        .await
+        .expect("failed to insert data");
+
+    let index = create_index(CreateIndexQuery::new(&session, &clients, &table, "v")).await;
+
+    let status = wait_for_index(client, &index).await;
```
knowack1, line 701: Instead, we can wait until the SELECT returns two vectors. Prefer the CQL API over the Vector Store HTTP API in validator tests to keep them more E2E-oriented.

crates/vector-store/src/lib.rs
```diff
@@ -603,6 +604,9 @@ pub struct DbEmbedding {
     pub primary_key: PrimaryKey,
     pub embedding: Option<Vector>,
     pub timestamp: Timestamp,
+    /// Values for filtering columns, keyed by column name.
+    /// For Alternator tables, these are extracted from the `:attrs` map.
+    pub column_values: BTreeMap<ColumnName, CqlValue>,
```
knowack1: I'm wondering if we also need timestamps for the filtering column values in order to resolve the eventual value.

crates/vector-store/benches/pipeline.rs
```diff
@@ -278,7 +278,7 @@ fn scan_fn_mpsc(
                     .send((
                         DbEmbedding {
                             primary_key,
-                            embedding,
+                            embedding: Some(embedding),
```
knowack1, line 268: Maybe its worth considering to make this struct more generic and keep only column_values - put the embedding also in column_values. This potentialy can be useful  for upcoming full text search changes also.

ewienik:
> @ewienik @QuerthDP I called you to review this PR. It started as adding the features I needed for saving non-key filtering columns in Alternator (known as "projected columns" there), but then I realized - this is exactly what is also needed for CQL to allow filtering on non-key columns. And indeed, my PR also fixes CQL :-) I wonder what you think about that - you actually had a feature in CQL that was **almost working**, but change needed the small changes here to fully work.
> 
> Still missing is the feature to **return** the projected columns in ANN requests, along with the primary key returned. I need this feature in Alternator, so I'll add this next.

We have a ticket for creating local indexes using non-primary key columns and filtering columns from non-primary columns: VECTOR-561. Currently you can filter by all columns from the primary key. We need to design a solution for fullscan and cdc update of non-primary key columns, need to take care of memory usage (all data is stored in RAM), need to use timestamps for modify or discard incoming data. I think the solution for using non-primary key columns should be in separate PRs.

nyh:
Pushed a new rebased version on the recent master (there have been a lot changes to the existing code and the rebase was complicated, but the AI just did it automatically, no sweat :-)).
I have not yet addressed the review comments. I'll do it now.

nyh:
CI continues to fail on random network problems:
```
Error: Unable to download artifact(s): Failed to ListArtifacts: Received non-retryable error: Failed request: (403) Forbidden: Error from intermediary with HTTP status code 403 "Forbidden"
```
Whoever is responsible for these CI jobs, please fix it. I have no idea who to even ask.

crates/vector-store/src/db.rs
nyh: Done.

nyh: Done, but it's not completely pretty. It meant making is_alternator() public for the crate (it wasn't), and converting from a string to a KeyspaceName - `KeyspaceName::from(keyspace_name.as_str())`, so I'll have the right type.

crates/vector-store/src/db_cdc.rs
nyh: I'll do it, but doing this sort of refactoring cleanly (not as an N+1th patch) is very difficult after I already have a series of patches that add more and more features to the same code. I'll get the AI to try to do this, but seriously - maybe we should stop caring that individual patches look clean and just squash patches always :-(

crates/vector-store/src/db_index.rs
nyh: Done.

crates/vector-store/src/lib.rs
nyh: Yes, good spot. We do. I'll change it.

crates/vector-store/src/vector.rs
nyh: Neither I nor my AI understood what you meant by this comment. The code does seem to compile. Maybe you mean if I try to compile this commit alone without the other commits this line will fail?

nyh:
I've addressed the comments in the first commit, but still need to address the rest and probably to split this PR into two as the reviewer asked. I'll continue doing this tomorrow. Even with the AI it takes a lot of time wall-clock time to do such patch reorganizations (I don't even want to think what the Copilot charges for this kind of work will be starting June 1st...).

knowack1:
> I've addressed the comments in the first commit, but still need to address the rest and probably to split this PR into two as the reviewer asked. I'll continue doing this tomorrow. Even with the AI it takes a lot of time wall-clock time to do such patch reorganizations (I don't even want to think what the Copilot charges for this kind of work will be starting June 1st...).

History reorganization is indeed very time-consuming, both for an AI and for a human. I often doubt whether it is worth spending this time just to keep a perfect history of changes. I do not think the investment in a perfect history is ever returned by later, more seamless investigations, code reviews, or easier backporting - but this is only my feeling, not backed by metrics.

However, in this specific case, this patch actually covers two features that could be delivered one by one:
- Support for filtering columns (the missing feature).
- Alternator support for provisioning (filtering columns).

Still, I am fine with keeping this in a single PR just need to be referenced accordingly.

ewienik:
> However, in this specific case, this patch actually covers two features that could be delivered one by one:
> 
>     * Support for filtering columns (the missing feature).
> 
>     * Alternator support for provisioning (filtering columns).
>  
> Still, I am fine with keeping this in a single PR just need to be referenced accordingly.

Support for filtering columns depends on #449 and VECTOR-561. I thought that #449 can be done quit fast, but it seems there are more problems similar to supporting vector search local indexes based on non-primary key columns (problem similar to MV). Both PR tries to extend DBEmbedding with map or vector of values from columns - it seems a challenge. Alsw VECTOR-561 is going to extend in similar way this struct. I think the best way to fight with this is to refactor vector-store with VECTOR-561 issue to allow filtering on custom columns and in next steps allow multi-column targets and local indexes on non-primary columns.

ewienik:
As both #501 and #517 are merged there is support for retrieving and storing non-key columns.

swasik:
@nyh I am closing this PR as plenty of it is already implemented in Vector Store. If you think that some parts of it are still worth merging, please move the relevant commits to the new PR (or start it from scratch).

PR #445, opened 2026-05-11
fix: reconnect Scylla session on hot-reloaded CQL config changes

smoczy123:
Reconnect the DB session when any session-building CQL setting changes, not just the Scylla URI or credentials.

This covers hot-reloaded updates to settings like the CQL URI translation map, which can effectively change the target Scylla port. Add a regression test for translation map changes to ensure reload triggers reconnect.

Fixes: VECTOR-384

### DIFF
--- a/crates/vector-store/src/db.rs
+++ b/crates/vector-store/src/db.rs
@@ -302,6 +302,7 @@ pub(crate) async fn new(
                         warn!("CDC error notification received, cancelling the current ScyllaDB connection...");
                         // Cancel existing session and let reconnection timer handle it
                         session_tx.send(None).ok();
+                        internals.create_session(None).await;
                     }
 
                     // Config changes - check if URI/credentials changed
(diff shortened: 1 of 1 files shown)


crates/vector-store/src/db.rs
```diff
@@ -309,37 +309,18 @@ pub(crate) async fn new(
...
+                            if !reconnect_reasons.is_empty() {
+                                warn!(
+                                    "ScyllaDB session configuration changed ({}), will reconnect...",
+                                    reconnect_reasons.join(", ")
+                                );
                                 config = new_config;
 
                                 // Cancel existing session and let reconnection timer handle it
                                 session_tx.send(None).ok();
+                                internals.create_session(None).await;
```
knowack1, line 324: There is also one other place when we call `session_tx.send(None)`. Should we also call `internals.create_session(None)` there?

knowack1 (dismissed):

crates/vector-store/src/db.rs
ewienik: This change seems not connected to the commit description. Does it fix something?

knowack1: Good point. I think this need to be answered by @smoczy123 .

smoczy123: Correct me if I'm wrong @ewienik, but don't we need to do the `internals.create_session(None)` when we reconnect? The internals don't change if we don't do it.

smoczy123:
Changelog:

- added `internals.create_session(None)` in the second required place

crates/vector-store/src/db.rs
knowack1: If this does not fix anything or/and  we do not have reproduction path/test case, I would skip this change in this PR. Create dedicated one if we think this is important, with well described problem.

smoczy123: I think both changes are needed  to actually fix this issue

knowack1 (approved):

PR #447, opened 2026-05-12
vector-store: fix index size counter desync

m-szymon:
usearch's remove() returns Ok(usize) where the value is the number of entries actually removed (0 if the key was not found). The previous wrapper converted this to Ok(()) unconditionally, causing size.fetch_sub(1) to be called even when nothing was removed from the index.

This manifests when an item is first seen during the range scan with an invalid vector (e.g. wrong dimensionality), causing the usearch add() to return Err. The item is still recorded in primary_ids and vector_timestamps with ETValue::Some — intentionally, to advance the timestamp for CDC deduplication. When a later valid update arrives via CDC, the Entry::Occupied path emits RemoveBeforeAddVector (because vector_already_exists=true) followed by AddVector. The remove hits a key that was never in usearch, returns Ok(0), but the size counter was decremented anyway. The subsequent successful add only brought it back to the pre-update value, so the count never advanced.

Fix: UsearchIndex::remove() now returns Ok(bool) and size is decremented only when an entry was actually removed.

### DIFF
--- a/crates/vector-store/src/index/usearch.rs
+++ b/crates/vector-store/src/index/usearch.rs
@@ -144,7 +144,7 @@ trait UsearchIndex {
     fn size(&self) -> usize;
     fn capacity(&self) -> usize;
     fn add(&self, primary_id: PrimaryId, vector: &Vector) -> anyhow::Result<()>;
-    fn remove(&self, primary_id: PrimaryId) -> anyhow::Result<()>;
+    fn remove(&self, primary_id: PrimaryId) -> anyhow::Result<bool>;
     fn search(
         &self,
         vector: &Vector,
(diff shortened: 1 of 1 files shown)


m-szymon:
Initially it was part of https://github.com/scylladb/vector-store/pull/392 (as some tests added there would fail).
But it deserves separate PR.

Answering remarks from that review:
@ewienik 
> Are you sure that usearch can file silently because of dimension? We should check this before calling usearch add().

Initially I wrote that "usearch add() fail silently." I used wrong wording. Usearch returns error. And it is handled and generates warn. I meant that such vector in general is ignored (not indexed) and considered it a correct behavior.

> What about converting the result into an error if result == 0?

It is not an error. And it would trigger warn, when we actually fix the invalid entry.
Unless we keep information with timestamp that vector is invalid and was not added to index, in which case we will not need to try to remove it. But I think it is overkill.

@QuerthDP 
I used bool as a return type as you suggested.

QuerthDP (approved):

PR #448, opened 2026-05-12
scripts: replace QEMU emulation with cross-compilation for arm64 builds

swasik:
## Changes

- Cross-compile arm64 binaries using `gcc-aarch64-linux-gnu` / `g++-aarch64-linux-gnu` instead of QEMU emulation
- QEMU remains only for Docker image packaging (`docker build --platform`)
- `CARGO_BUILD_TARGET`, `CC_*`, `CXX_*`, `AR_*`, `CARGO_TARGET_*_LINKER` env vars set correctly (underscore-normalized for `cc` crate compatibility)
- Cross container runs as root (for `apt-get`), then `chown`s target dir back to host UID/GID
- `release_dir_for()` helper resolves output path whether cross or native build
- `upload-release` searches multiple candidate paths for tarballs
- Dependency validation step moved from `build-release` to `release` (avoids duplicate checks)

## Tests

- Tested by creating a test release in the fork: https://github.com/swasik/vector-store/releases/tag/1.6.1-dev
- The arm binary was tested on r8g instance

## Dependencies

Depends on #436

Fixes: VECTOR-665


swasik:
This should be resolved as part of https://scylladb.atlassian.net/browse/VECTOR-790.

PR #449, opened 2026-05-13
vector-store: preparation for multi-target-column indexes

knowack1:
In preparation for full-text search (FTS) indexes, which are planned to support multiple target columns in Milestone 3—we need to ensure the existing architecture is ready for this extension.

As validation, we are generalizing the concept of target columns to support multiple columns even for vector indexes. This ensures that the same code path will work for both vector and FTS indexes moving forward.

This approach prevents significant future refactoring when multi-target-column FTS indexes are introduced.

Fixes VECTOR-676


crates/vector-store/src/indexes.rs
```diff
@@ -57,20 +57,20 @@ impl Ord for NeedsFiltering {
...
 
 impl From<&IndexMetadata> for RoutingGroupKey {
     fn from(metadata: &IndexMetadata) -> Self {
         Self {
             keyspace: metadata.keyspace_name.clone(),
             table: metadata.table_name.clone(),
-            column: metadata.target_column.clone(),
+            columns: metadata.target_columns.clone(),
         }
     }
```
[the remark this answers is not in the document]
knowack1: Lets skip this for now. Order must be preserved.

crates/vector-store/src/table.rs
```diff
@@ -894,6 +905,128 @@ impl Index {
...
+                all_operations[col_idx].push(Operation::AddVector {
+                    primary_id,
+                    partition_id,
+                    vector: vector.clone(),
+                    is_update: false,
+                });
+            }
+        }
+        Ok(())
+    }
```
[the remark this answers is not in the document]
knowack1: Fixed

crates/vector-store/src/table.rs
```diff
@@ -1052,11 +1188,11 @@ impl Table {
...
-            .and_then(|index| match index.vector_timestamps.get(primary_id)? {
-                ETValue::Some(epoch, _, _) => Some(*epoch),
-                ETValue::None(_, _) => None,
+            .is_some_and(|index| {
+                index.vector_timestamps.iter().any(|ts| {
+                    matches!(ts.get(primary_id), Some(ETValue::Some(epoch, _, _)) if *epoch == primary_id.epoch())
+                })
             })
-            .is_some_and(|epoch| epoch == primary_id.epoch())
     }
```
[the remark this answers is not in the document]
knowack1: Fixed

crates/vector-store/src/db_index_backend.rs
```diff
@@ -90,7 +90,7 @@ pub(crate) fn range_scan_query(
             "
         )
     } else {
-        let vector = CqlIdentifier::new(target_column.as_ref());
+        let vector = CqlIdentifier::new(target_columns[0].as_ref());
```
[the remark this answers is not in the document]
knowack1: Made code more generic instead of adding todos

crates/vector-store/src/lib.rs
```diff
@@ -604,12 +604,19 @@ impl DbCustomIndex {
...
+pub struct EmbeddingValue {
     pub embedding: Option<Vector>,
     pub timestamp: Timestamp,
 }
 
+#[derive(Clone, Debug, PartialEq)]
+pub struct DbEmbedding {
+    pub primary_key: PrimaryKey,
+    // List of embeddings for each indexed column.
+    // The order corresponds to the order of target columns in IndexMetadata.
```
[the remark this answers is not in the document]
knowack1: Fixed

crates/vector-store/src/monitor_indexes.rs
```diff
@@ -178,7 +178,10 @@ async fn get_indexes(db: &Sender<Db>) -> anyhow::Result<HashSet<IndexMetadata>>
...
                 idx.keyspace.clone(),
                 idx.table.clone(),
-                idx.target_column.clone(),
+                idx.target_columns
+                    .first()
+                    .ok_or_else(|| anyhow::anyhow!("index {idx:?} has empty target_columns"))?
+                    .clone(),
                 idx.index.clone(),
             )
             .await
```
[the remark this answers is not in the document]
knowack1: This is good for this transition period.

crates/vector-store/src/table.rs
```diff
@@ -1076,124 +1215,96 @@ impl TableAdd for Table {
...
+        if let Some((_, index)) = self.indexes.iter().next() {
+            let expected = index.vector_timestamps.len();
+            if num_columns != expected {
+                bail!(
+                    "embeddings length {num_columns} does not match \
+                    expected num_target_columns {expected}"
+                );
+            }
+        }
+        Ok(())
```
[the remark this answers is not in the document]
knowack1: Changed.

crates/vector-store/src/table.rs
```diff
@@ -894,6 +905,128 @@ impl Index {
...
+                    if state.vector_exists {
+                        ops.push(Operation::RemoveVector {
+                            primary_id,
+                            partition_id,
+                        });
+                        if self.data.remove_row(primary_id) {
+                            ops.push(Operation::RemovePartition { partition_id });
+                        }
+                    }
+                }
```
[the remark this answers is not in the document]
knowack1: Removed

crates/vector-store/src/db_index.rs
```diff
@@ -633,6 +615,39 @@ impl Statements {
...
+        };
+        let Ok(vector) =
+            Vector::try_from(vector_value).inspect_err(|err| debug!("range_scan_stream: {err}"))
+        else {
+            return None;
+        };
+        embeddings.push(Some(EmbeddingValue {
+            embedding: Some(vector),
+            timestamp,
+        }));
```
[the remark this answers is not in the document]
knowack1: Fixed.

crates/vector-store/src/table.rs
```diff
@@ -894,6 +905,128 @@ impl Index {
...
+                primary_id,
+                ETValue::Some(primary_id.epoch(), ev.timestamp, ()),
+            )?;
+            if let Some(vector) = &ev.embedding {
+                all_operations[col_idx].push(Operation::AddVector {
+                    primary_id,
+                    partition_id,
+                    vector: vector.clone(),
+                    is_update: false,
+                });
```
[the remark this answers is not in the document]
knowack1: Code removed

crates/vector-store/src/db_index.rs
```diff
@@ -633,6 +615,46 @@ impl Statements {
...
+            }));
+            continue;
+        };
+        let Ok(vector) =
+            Vector::try_from(vector_value).inspect_err(|err| debug!("range_scan_stream: {err}"))
+        else {
+            return None;
+        };
+        embeddings.push(Some(EmbeddingValue {
+            embedding: Some(vector),
```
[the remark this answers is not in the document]
knowack1: This is how it was in original code.

crates/vector-store/src/db_index_backend.rs
```diff
@@ -30,78 +32,119 @@ pub(crate) struct IndexLocation {
...
+                            };
+                            matches.then_some(val)
+                        });
+                        match value {
+                            Some(v) => Ok(Some(Some(Vector::try_from(v.clone())?))),
+                            None => Ok(Some(None)),
+                        }
+                    })
+                    .collect()
             }
```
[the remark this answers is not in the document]
knowack1: This is problem already existing on the master branch. The correct fix can be seen here: https://github.com/scylladb/vector-store/pull/444/changes/088fb06b980e964fd58b7b53dfda8823219e1396. Lets fix this in dedicated PR.

crates/vector-store/src/table.rs
```diff
@@ -1076,122 +1130,126 @@ impl TableAdd for Table {
...
         self.reserve_partition_ids()?;
 
-        let mut operations = vec![];
+        let mut all_operations: Vec<Vec<Operation>> = (0..db_embedding.embeddings.len())
+            .map(|_| Vec::new())
+            .collect();
 
         let primary_key = db_embedding.primary_key;
-        let vector = db_embedding.embedding;
+        let embeddings = db_embedding.embeddings;
```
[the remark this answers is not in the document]
knowack1: Fixed. Check embeddings dimensions at function entry point.

crates/vector-store/src/table.rs
```diff
@@ -1076,122 +1130,126 @@ impl TableAdd for Table {
...
+                                    vector: vector.clone(),
+                                    is_update: false,
+                                });
+                            }
+                        }
+                        Ok(())
+                    },
+                )?;
+
+                Ok(all_operations)
```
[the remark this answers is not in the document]
knowack1: Restored check of any vector in embeddings columns.

crates/vector-store/src/db_index_backend.rs
```diff
@@ -30,35 +30,35 @@ pub(crate) struct IndexLocation {
...
+            Self::Cql { target_columns }
         }
     }
 }
 
 impl DbIndexBackend {
     pub fn vector_column_name(&self) -> &str {
         match self {
-            Self::Cql { target_column } => target_column.as_ref(),
+            Self::Cql { target_columns } => target_columns[0].as_ref(),
```
ewienik: `target_columns[0]` could panic (here and in other places), we should avoid this. Consider creating a newtype `TargetColumns` that would be responsible that only `len > 0` are possible.
As this type `TargetColumns` will be invariant you can define it using `Arc<Box[_]>` to limit memory usage.

crates/vector-store/src/lib.rs
```diff
@@ -604,12 +604,19 @@ impl DbCustomIndex {
 }
 
 #[derive(Clone, Debug, PartialEq)]
-pub struct DbEmbedding {
-    pub primary_key: PrimaryKey,
+pub struct EmbeddingValue {
```
ewienik: Consider rename it to `DbIndexedValue` as for the future it would be more meaningful

crates/vector-store/src/lib.rs
```diff
@@ -604,12 +604,19 @@ impl DbCustomIndex {
...
     pub embedding: Option<Vector>,
     pub timestamp: Timestamp,
 }
 
+#[derive(Clone, Debug, PartialEq)]
+pub struct DbEmbedding {
```
ewienik: Consider rename it to `DbIndexedRow`

crates/vector-store/src/lib.rs
```diff
@@ -604,12 +604,19 @@ impl DbCustomIndex {
...
+    pub primary_key: PrimaryKey,
+    /// List of embeddings for each indexed column.
+    /// The order corresponds to the order of target columns in IndexMetadata.
+    pub embeddings: Vec<Option<EmbeddingValue>>,
```
ewienik: Consider rename it into `values`

crates/vector-store/src/table.rs
```diff
@@ -1076,14 +1088,23 @@ impl TableAdd for Table {
         &mut self,
         _index_key: &IndexKey,
         db_embedding: DbEmbedding,
-    ) -> anyhow::Result<Vec<Operation>> {
+    ) -> anyhow::Result<Vec<Vec<Operation>>> {
```
ewienik: We don't need to create a hierarchy of vectors. Can we use only single vector?

crates/vector-store/src/table.rs
```diff
@@ -784,9 +786,11 @@ impl Index {
             IndexData::Global => {}
             IndexData::Local { ids, .. } => ids.resize_with(new_size, || None),
         }
-        self.vector_timestamps.resize_with(new_size, || {
-            ETValue::None(Epoch::new(), Timestamp::UNIX_EPOCH)
-        });
+        for ts in &mut self.vector_timestamps {
```
ewienik: nit: consider using `iter_mut().for_each()`

crates/vector-store/src/table.rs
```diff
@@ -911,6 +915,7 @@ pub struct Table {
     _index_id_generator: IndexIdGenerator,
     index_ids: BTreeMap<IndexKey, IndexId>,
     indexes: BTreeMap<IndexId, Index>,
+    num_target_columns: usize,
```
ewienik: this is index parameter, not table, it should be in index type.

crates/vector-store/src/table.rs
```diff
@@ -1049,14 +1058,17 @@ impl Table {
         Ok(())
     }
 
+    /// Returns true if at least one column has a vector present at the expected epoch.
+    /// This uses `any` semantics intentionally: a primary id is valid as long as it holds
+    /// a vector in any target column (columns are independently insertable/removable).
```
ewienik: I would say that all target columns must have valid epoch. There shouldn't be mismatched here. Target value exists or doesn't exist. If there was inserts/removes all must be visible with the correct epoch.

crates/vector-store/src/table.rs
```diff
@@ -1092,106 +1113,136 @@ impl TableAdd for Table {
...
+                            let primary_id = primary_id.next_epoch();
+                            ops.push(Operation::AddVector {
                                 primary_id,
                                 partition_id,
+                                vector: vector.clone(),
+                                is_update: vector_already_exists,
                             });
-                            if index.data.remove_row(primary_id) {
-                                operations.push(Operation::RemovePartition { partition_id });
+                            index.vector_timestamps[col_idx]
```
ewienik: `[col_idx]` could panic. You have already `vector_timestamps` prepared

crates/vector-store/src/table.rs
```diff
@@ -1092,106 +1113,136 @@ impl TableAdd for Table {
...
+                                let Some(partition_key) =
+                                    index.partition_key(&self.primary_key_columns, &primary_key)
+                                else {
+                                    return Ok(());
+                                };
+                                index.add(primary_id, partition_key)?
+                            }
+                        };
+                        for (col_idx, embedding) in embeddings.iter().enumerate() {
+                            let Some(ev) = embedding else { continue };
```
ewienik: When there is insert and not all target values exists we should insert some null value. Timestamp could be 0 - any future values will override this. But I think we need to have a logical "null" value.

ewienik (changes requested):
I've already checked the first commit. Needs some refactoring.
Consider running clippy/fmt on the intermediate commit - for better reviewer experience :-)

crates/vector-store/src/db_index_backend.rs
knowack1: This issue was later fixed in commit 727c7b10ab2970f710e2b51b43757e1d5ae22301  After this commit, `db_index_backend` is also capable of handling multi-target columns, eliminating calls like `target_columns[0]`. So, with this, I think there is no need to introduce `TargetColumns`.

I could merge this and the 727c7b10 into single commit, but I tried not to put too much into one commit.

crates/vector-store/src/lib.rs
knowack1: Done in dedicated commit https://github.com/scylladb/vector-store/pull/449/changes/69397c5f71240dd972192f1abc03c640b36120e0

crates/vector-store/src/lib.rs
knowack1: Done in dedicated commit: https://github.com/scylladb/vector-store/pull/449/changes/971086ee66f4c8f42144713e857f2b9eaa8bf785

crates/vector-store/src/lib.rs
knowack1: Done in dedicated commit: https://github.com/scylladb/vector-store/pull/449/changes/69397c5f71240dd972192f1abc03c640b36120e0

crates/vector-store/src/table.rs
knowack1: The returned type has the positional pattern. Every entry in the most upper vector corresponds the the operations for a given target column value. Added comment to the returned type also extracted `Vec<Operation>` into `Operations` alias.

crates/vector-store/src/table.rs
knowack1: Done

crates/vector-store/src/table.rs
knowack1: Fixed

crates/vector-store/src/table.rs
knowack1: Done.

crates/vector-store/src/table.rs
knowack1: Panic fixed. There were 2 calls because there is mut and non-mut versions of vector_timestamps needed in the same scope. Nevertheless I moved the first call into `index::read_column_state` so that its more compact now.

crates/vector-store/src/table.rs
knowack1: Fixed. In this case Now we put:
```
ETValue::None(primary_id.epoch(), Timestamp::UNIX_EPOCH),
```

crates/vector-store/src/db_index_backend.rs
ewienik: For the maintenance it would be better to introduce newtype which will hold contract invariant. The issue might be solved by the later commit but we should limit similar problems in the future.
Additionally newtype gives better semantic by the type directly, not by the variable name.
You can introduce this newtype before this commit. IMHO it is better to fix issues in the commit when they appears, not in the following ones.

crates/vector-store/src/lib.rs
ewienik: Could you move this refactoring before this commit? It would be shorter context of changes in commits.

crates/vector-store/src/table.rs
ewienik: > The returned type has the positional pattern. Every entry in the most upper vector corresponds the the operations for a given target column value. Added comment to the returned type also extracted `Vec<Operation>` into `Operations` alias.

This is not always truth - sometimes one add to table produces two operations for index.

crates/vector-store/src/db_index_backend.rs
ewienik: Similar pattern of a non-empty list of columns we have in `primary_key_columns` and local `key_columns`. Such newtype can be used also for them. So the name could be different - maybe `NonEmptyColumns`? Or another name?

crates/vector-store/src/lib.rs
ewienik: Could you move that refactor before this commit?

crates/vector-store/src/lib.rs
ewienik: Could you move that refactor before this commit?

crates/vector-store/src/table.rs
ewienik: We don't need this hierarchy information - we don't need vector of vectors - we don't need this information further in a pipeline, do we?  `add` can produce flat vector.

crates/vector-store/src/table.rs
```diff
@@ -1092,106 +1128,136 @@ impl TableAdd for Table {
...
-                            missing partition id for index_id {index_id:?} and primary_id {primary_id:?}"
-                        )
-                    })?;
-                    if let Some(vector) = &vector {
-                        if vector_already_exists {
-                            operations.push(Operation::RemoveBeforeAddVector {
-                                primary_id,
-                                partition_id,
-                            });
+                    for (col_idx, embedding) in embeddings.iter().enumerate() {
```
ewienik, line 1131: We should check all embedding if they have something to change, and we should produce single Operation for one row (for all values in targets) - should we update or not.

ewienik (changes requested):
I see an issue with multi-column index values - what should we do when only single column has value? We should provide this information to the pipeline in single operation - monitor_items should now which column is newer and update index only with this value. I think we should provide some bitset to the Add or Remove operation - which columns are affected with this operation. So we shouldn't build Vec<Vec<Operation>> but extend Operation with bitset about columns affected.

knowack1:
> I see an issue with multi-column index values - what should we do when only single column has value? We should provide this information to the pipeline in single operation - monitor_items should now which column is newer and update index only with this value. I think we should provide some bitset to the Add or Remove operation - which columns are affected with this operation. So we shouldn't build Vec<Vec> but extend Operation with bitset about columns affected.

I think this is currently handled correctly (see my comment https://github.com/scylladb/vector-store/pull/449#discussion_r3279599073). When value of specific column was not updated the Operations are empty for this specific position (column) in the `Vec<Operations>`.

crates/vector-store/src/table.rs
knowack1: The current structure of `Vec<Operations>` works as follows. Assuming 4 target columns, `add()` can produce a `Vec<Operations>` like this:
```
[
  [0] = [AddVector{...}],                                                        // New value for target_columns[0]
  [1] = [RemoveBeforeAddVector{...}, AddVector{...}],   // Updated value for target_columns[1]
  [2] = [RemoveVector{...}],                                                 // Removed value for target_columns[2]
  [3] = []                                                                                  // No update (nothing changed) for target_columns[3]
]
```

I think this positional pattern is consistent with what we have in `DBEmbedding` (`DbIndexedRow`).

At this stage, I don't see how to effectively apply a bitset in this case - could you please advise on this? Another approach I see would be to add a target_column_index into the Operation variants.

Also, depending on the chosen approach, I would apply the same pattern to `DBEmbedding` (`DbIndexedRow`) to keep the pattern consistent across the application.

crates/vector-store/src/db_index_backend.rs
knowack1: Makes sense. I will introduce it.

ewienik:
> > I see an issue with multi-column index values - what should we do when only single column has value? We should provide this information to the pipeline in single operation - monitor_items should now which column is newer and update index only with this value. I think we should provide some bitset to the Add or Remove operation - which columns are affected with this operation. So we shouldn't build Vec but extend Operation with bitset about columns affected.
> 
> I think this is currently handled correctly (see my comment [#449 (comment)](https://github.com/scylladb/vector-store/pull/449#discussion_r3279599073)). When value of specific column was not updated the Operations are empty for this specific position (column) in the `Vec<Operations>`.

But this is not performant way - out of cache. We will have a dynamic vector of dynamic vectors - it is hard to check on monitor_items - you have to do several iteration over vector of vectors. Data are not easily grasped as there could be two operations per one cdc insert. It is better to store bitset in Add/Delete operations indicating which column was changed. Additionally there could be more operations - no only add/remove vector - it could be add/remove partition. In the future there could be even more of them.

crates/vector-store/src/table.rs
ewienik: > The current structure of `Vec<Operations>` works as follows. Assuming 4 target columns, `add()` can produce a `Vec<Operations>` like this:
> 
> ```
> [
>   [0] = [AddVector{...}],                                                        // New value for target_columns[0]
>   [1] = [RemoveBeforeAddVector{...}, AddVector{...}],   // Updated value for target_columns[1]
>   [2] = [RemoveVector{...}],                                                 // Removed value for target_columns[2]
>   [3] = []                                                                                  // No update (nothing changed) for target_columns[3]
> ]
> ```

We have also RemovePartition operation. In the future we would have more operations.
Additionally each Operation is independent of target column - `add` method should tell what should be done after the insert of DbIndexRow. 
> 
> I think this positional pattern is consistent with what we have in `DBEmbedding` (`DbIndexedRow`).
> 
> At this stage, I don't see how to effectively apply a bitset in this case - could you please advise on this? Another approach I see would be to add a target_column_index into the Operation variants.
> 
> Also, depending on the chosen approach, I would apply the same pattern to `DBEmbedding` (`DbIndexedRow`) to keep the pattern consistent across the application.

The position in Vec<Operation> shouldn't be coordinated with position of column in DbIndexRow - these two are independent.

knowack1:
These changes requires revisited approach according to https://scylladb.atlassian.net/browse/VECTOR-676?focusedCommentId=60373.

PR #451, opened 2026-05-15
monitor_indexes: make schema poll interval configurable

nyh:
The interval at which the vector store polls Scylla for schema changes (new/removed vector indexes) was hardcoded to 1 second. This introduces up to 1 second of latency before the vector store even notices a newly created index. For real deployments, this is hardly noticable, but for tests - this delay adds valuable seconds to each test.

Add a VECTOR_STORE_MONITOR_INDEXES_INTERVAL environment variable (default: 1s, matching the previous behavior) to allow overriding this interval. The new variable is documented in README.md alongside the other interval configuration options.

For example, before this patch each of Alternator's tests that uses the vector store typically took more than 2 seconds. If the test running sets VECTOR_STORE_MONITOR_INDEXES_INTERVAL to 100ms, each of these tests now takes less than 0.5 seconds.


nyh:
The validator failure is some docker failure that has nothing to do with this PR :-( How can I force it to retry?

ewienik:
> The validator failure is some docker failure that has nothing to do with this PR :-( How can I force it to retry?

You can go directly to the ci failure - GitHub Actions and on the top-right side you have button to re-run failed jobs. If you don't see it I can rerun it.

crates/vector-store/src/monitor_indexes.rs
```diff
@@ -46,15 +45,23 @@ pub(crate) async fn new(
...
+                        .unwrap_or(Duration::from_secs(1)),
+                    config.alter_index_simulator,
+                )
+            };
+            let mut interval = time::interval(interval_duration);
 
             let mut schema_version = SchemaVersion::new();
             let mut indexes = HashSet::new();
-            let mut alter_index_simulator = config_rx.borrow_and_update().alter_index_simulator;
+            let mut alter_index_simulator = alter_index_simulator_init;
```
ewienik: nit: It could be combined with previous change:
```
let (interval_duration, mut alter_index_simulator) = { ... }
```

ewienik (changes requested):
Looks good

crates/vector-store/src/monitor_indexes.rs
nyh: Done.

ewienik (approved):

PR #453, opened 2026-05-18
vector-store: fullscan stops when index is dropped

smoczy123:
Handle None from rx_index.recv() in the db_index fullscan loop by breaking immediately instead of waiting indefinitely for initial_scan. This ensures fullscan terminates when the index is dropped, fixing VECTOR-663.

Also add validator test full_scan_stops_when_index_is_dropped to verify fullscan halts when index is removed by checking for absence of post-drop scan statement executions.

Fixes: VECTOR-663

### DIFF
--- a/crates/validator/src/full_scan.rs
+++ b/crates/validator/src/full_scan.rs
@@ -8,11 +8,18 @@ use crate::common::*;
 use async_backtrace::framed;
 use e2etest::TestCase;
 use e2etest_scylla_proxy_cluster::ScyllaProxyClusterExt;
+use httpapi::IndexInfo;
+use httpapi::IndexStatus;
+use httpclient::HttpClient;
 use scylla_proxy::Condition;
 use scylla_proxy::Reaction;
+use scylla_proxy::RequestFrame;
+use scylla_proxy::RequestOpcode;
 use scylla_proxy::RequestReaction;
 use scylla_proxy::RequestRule;
+use std::collections::HashMap;
 use std::time::Duration;
+use tokio::sync::mpsc;
 use tracing::info;
 
 #[framed]
@@ -25,6 +32,11 @@ pub(crate) async fn new() -> TestCase<TestActors> {
             "full_scan_is_completed_when_responding_to_messages_concurrently",
             timeout,
             full_scan_is_completed_when_responding_to_messages_concurrently,
+        )
+        .with_test(
+            "full_scan_stops_when_index_is_dropped",
+            timeout,
+            full_scan_stops_when_index_is_dropped,
         )
 }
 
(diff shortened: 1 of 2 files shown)


crates/vector-store/src/db_index.rs
```diff
@@ -236,8 +236,16 @@ pub(crate) async fn new(
...
+                    msg = rx_index.recv() => {
+                        match msg {
+                            Some(msg) => {
+                                tokio::spawn(process(Arc::clone(&statements), msg, completed_scan_length.clone()));
+                            }
+                            None => {
+                                // Index was dropped, stop fullscan.
+                                break;
+                            }
+                        }
```
ewienik, line 248: nit: you can use `let Some(msg) = msg else {..}` pattern

crates/validator/src/full_scan.rs
```diff
@@ -127,3 +136,139 @@ async fn full_scan_is_completed_when_responding_to_messages_concurrently(actors:
...
+    info!("started");
+
+    let (session, clients) = prepare_connection_single_vs_no_tls(&actors).await;
+    let client = clients.first().unwrap();
+
+    let keyspace = create_keyspace(&session).await;
+    let table = create_table(
+        &session,
+        "id INT PRIMARY KEY, embedding VECTOR<FLOAT, 3>",
+        Some("CDC = {'enabled': true}"),
```
ewienik: We don't need to directly enable CDC, do we?

crates/validator/src/full_scan.rs
```diff
@@ -127,3 +136,139 @@ async fn full_scan_is_completed_when_responding_to_messages_concurrently(actors:
...
+    let index = create_index(CreateIndexQuery::new(
+        &session,
+        &clients,
+        &table,
+        "embedding",
+    ))
+    .await;
+
+    info!("Waiting a bit for fullscan to be in progress");
+    tokio::time::sleep(Duration::from_millis(500)).await;
```
ewienik: Consider waiting for the specific index status instead of sleep.

crates/validator/src/full_scan.rs
```diff
@@ -127,3 +136,139 @@ async fn full_scan_is_completed_when_responding_to_messages_concurrently(actors:
...
+
+    let mut fullscan_execute_after_drop = 0usize;
+    while let Ok((frame, _)) = rx_feedback.try_recv() {
+        let Some(statement_id) = execute_statement_id(&frame) else {
+            continue;
+        };
+        if statement_id == fullscan_statement_id {
+            fullscan_execute_after_drop += 1;
+        }
+    }
```
ewienik: Consider refactoring to not use sleep. You might use time::timeout and first received packet within 2secons finishes wait.

ewienik (changes requested):

crates/validator/src/full_scan.rs
```diff
@@ -127,3 +136,139 @@ async fn full_scan_is_completed_when_responding_to_messages_concurrently(actors:
...
+
+    assert_eq!(
+        fullscan_execute_after_drop, 0,
```
knowack1, line 259: It is hard to validate this behavior in a deterministic way. There is a race condition 
regarding how long to wait for execute statement NOT to happen. Maybe an internal counters approach could be useful here

crates/validator/src/full_scan.rs
```diff
@@ -127,3 +136,139 @@ async fn full_scan_is_completed_when_responding_to_messages_concurrently(actors:
...
+        }
+        if let Some(statement_id) = execute_statement_id(&frame) {
+            *execute_id_counts.entry(statement_id).or_insert(0) += 1;
+        }
+    }
+
+    let fullscan_statement_id = execute_id_counts
+        .into_iter()
+        .max_by_key(|(_, count)| *count)
+        .map(|(statement_id, _)| statement_id)
```
knowack1: Considering wrapping this code in some nicely named function:
```
    let mut execute_id_counts: HashMap<Vec<u8>, usize> = HashMap::new();
    while let Ok((frame, _)) = rx_feedback_before_drop.try_recv() {
        if frame.opcode != RequestOpcode::Execute {
            continue;
        }
        if let Some(statement_id) = execute_statement_id(&frame) {
            *execute_id_counts.entry(statement_id).or_insert(0) += 1;
        }
    }
```
Similar code is also duplicated later when waiting for execute not to happen.

smoczy123:
Changelog:

- addressed most of the commetns

crates/validator/src/full_scan.rs
smoczy123: I see the issue, but I think it would be really hard to solve it, I think the way it is currently should almost always work, even if not deterministic

crates/validator/src/full_scan.rs
```diff
@@ -127,3 +139,163 @@ async fn full_scan_is_completed_when_responding_to_messages_concurrently(actors:
...
+                Ok(status) if status.status == IndexStatus::Bootstrapping => Some(status),
+                _ => None,
+            }
+        },
+        format!(
+            "index '{}' must reach BOOTSTRAPPING at {}",
+            index.index,
+            client.url()
+        ),
+        Duration::from_secs(60),
```
knowack1: We have DEFAULT_TEST_TIMEOUT and DEFAULT_OPERATION_TIMEOUT maybe some of them could/should be reused here?

crates/validator/src/full_scan.rs
```diff
@@ -127,3 +139,163 @@ async fn full_scan_is_completed_when_responding_to_messages_concurrently(actors:
...
+        .db_proxy
+        .change_request_rules(Some(vec![RequestRule(
+            Condition::RequestOpcode(RequestOpcode::Execute),
+            RequestReaction::noop().with_feedback_when_performed(tx_feedback),
+        )]))
+        .await;
+
+    let mut fullscan_execute_after_drop = 0usize;
+    let settle_until = tokio::time::Instant::now() + Duration::from_millis(500);
+    let deadline = tokio::time::Instant::now() + Duration::from_secs(2);
```
knowack1, line 248: I think its worth commenting why 2s, also mentioning the deterministic verification problem.

crates/validator/src/full_scan.rs
```diff
@@ -127,3 +139,163 @@ async fn full_scan_is_completed_when_responding_to_messages_concurrently(actors:
...
+    session
+        .query_unpaged(format!("DROP INDEX {}", index.index), ())
+        .await
+        .expect("failed to drop an index");
+
+    info!("Waiting for index to be removed from vector-store");
+    wait_for(
+        || async { client.indexes().await.is_empty() },
+        "index must be removed",
+        Duration::from_secs(10),
```
knowack1: We have 10s and 60s timeouts in single test. Is it possible to stick to single value? (maybe DEFAULT_OPERATION_TIMEOUT?)

smoczy123:
Changelog:

- adressed @knowack1 comments

ewienik (approved):

PR #457, opened 2026-05-19
engine: make index-status sync interval configurable

nyh:
The engine's main loop syncs each index's status from node_state into the engine's internal cache every second (hardcoded CHECK_INTERVAL). This cache is what the get_index_status HTTP endpoint reads, which is what Scylla calls to determine whether a vector index is ready.

Because this interval was hardcoded to 1s, a newly-built index would continue to be reported as CREATING for up to a second after the prefill scan finished and node_state marked it as Serving. This caused tests that wait for IndexStatus==ACTIVE to take at least ~1 second even when the actual indexing completed in milliseconds.

Expose the interval as VECTOR_STORE_INDEX_STATUS_UPDATE_INTERVAL (default: 1s, matching the previous hardcoded value) so test environments can set it to a lower value (e.g. 100ms) to reduce test latency.

For example, before this patch Alternator's vector test suite (in the Scylla core repository) took 57 seconds; With this patch, it is down to 28 seconds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ service supports also `.env` files.
 | `VECTOR_STORE_CDC_FINE_SAFETY_INTERVAL`    | Fine-grained CDC reader's safety interval for low-latency updates (ie. `100ms`)                                                                                                      | `100ms`                  |
 | `VECTOR_STORE_CDC_FINE_SLEEP_INTERVAL`     | Fine-grained CDC reader's sleep interval for low-latency updates (ie. `500ms`)                                                                                                       | `500ms`                  |
 | `VECTOR_STORE_MONITOR_INDEXES_INTERVAL`    | How often to poll Scylla for schema changes (new/removed vector indexes). The value is in human readable format (ie. `100ms`)                                                        | `1s`                     |
+| `VECTOR_STORE_INDEX_STATUS_UPDATE_INTERVAL` | How often to sync index status (e.g., BOOTSTRAPPING->SERVING) into the engine's cached state. The value is in human readable format (ie. `100ms`) | `1s`            |
 | `VECTOR_STORE_USEARCH_SIMULATOR`           | Enable simulator for USearch. Provides human readable delays for simulated operations (`search:add-remove:reserve`).                                                                 |                          |
 | `VECTOR_STORE_ALTER_INDEX_SIMULATOR`       | Enable simulator for missing `ALTER INDEX`. When enable indexes aren't deleted and their version is not checked.                                                                     | `false`                  |
 
--- a/crates/vector-store/src/config_manager.rs
+++ b/crates/vector-store/src/config_manager.rs
@@ -491,6 +491,12 @@ pub async fn load_config(env: impl Fn(&str) -> anyhow::Result<String>) -> anyhow
         .transpose()?
         .map(|v| v.into());
 
+    config.engine_status_update_interval = env("VECTOR_STORE_INDEX_STATUS_UPDATE_INTERVAL")
+        .ok()
+        .map(|v| v.parse::<humantime::Duration>())
+        .transpose()?
+        .map(|v| v.into());
+
     config.cql_uri_translation_map = env("VECTOR_STORE_CQL_URI_TRANSLATION_MAP")
         .ok()
         .map(|v| serde_json::from_str(&v))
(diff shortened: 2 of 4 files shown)


nyh:
It turns out https://github.com/scylladb/vector-store/pull/451 was not enough - it really sped up Alternator tests, but there was yet another hard-coded 1-second sleep in the vector store which still kept the tests slower than necessary, so here is another similar patch that makes another hard-coded 1 second sleep configurable.

If you wonder what kind of improvements these patches bring, well, before my configurable-sleep patches, Alternator's vector search test suite (the one in scylladb.git) took 172 seconds. With the new configuration parameters - 23 seconds. Seven times faster. Multi-second tests that did almost nothing besides sleep after creating indexes now take a fraction of a second. This last patch brought down the time from 57 seconds (after my previous PR) to 23 seconds (after this PR).

ewienik (approved):

QuerthDP:
> It turns out #451 was not enough - it really sped up Alternator tests, but there was yet another hard-coded 1-second sleep in the vector store which still kept the tests slower than necessary, so here is another similar patch that makes another hard-coded 1 second sleep configurable.
> 
> If you wonder what kind of improvements these patches bring, well, before my configurable-sleep patches, Alternator's vector search test suite (the one in scylladb.git) took 172 seconds. With the new configuration parameters - 23 seconds. Seven times faster. Multi-second tests that did almost nothing besides sleep after creating indexes now take a fraction of a second. This last patch brought down the time from 57 seconds (after my previous PR) to 23 seconds (after this PR).

Great job @nyh! I hope it speeds up the Validator tests as well. If it's not out of the box, we should change the defaults in our tests as well.

It gives me hope that we'll find some time someday to truly speed them up and refactor to be both effective and developer friendly (VECTOR-381).

nyh:
> Great job @nyh! I hope it speeds up the Validator tests as well.

I hope it does too - especially if those tests create and destroys a lot separate indexes, which I think it does - and my own tests definitely did. It's hard to share global vector search between multiple tests because you can find vectors written by other tests, so it's tempting to create a new vector index for each test, and those full-second sleeps add up.

> If it's not out of the box, we should change the defaults in our tests as well.

In my patches I left the original defaults so it won't speed up unless you override the defaults with environment variables. In https://github.com/scylladb/scylladb/pull/29959/commits/a5b08b129382c34f0630fb7c25b6f3bfe815967f I used:

```
VECTOR_STORE_MONITOR_INDEXES_INTERVAL=50ms
VECTOR_STORE_CDC_FINE_SLEEP_INTERVAL=50ms
VECTOR_STORE_INDEX_STATUS_UPDATE_INTERVAL=50ms
```

You don't really need to go this low, 100ms went most of the way.

You should also verify that your own test code doesn't also have retry loops with full-second sleeps.

> It gives me hope that we'll find some time someday to truly speed them up and refactor to be both effective and developer friendly ([VECTOR-381](https://scylladb.atlassian.net/browse/VECTOR-381)).

As you may have noticed, I've written (in Python) a very big suite of end-to-end tests for Alternator vector search + the vector store. It's now around 130 test functions in 4,000 lines of code, and the entire thing takes (after my latest sleep reductions) 22 seconds. That's still not as fast as I would have hoped, but it's already fast enough to be friendly _enough_. Before my changes, 172 seconds for the test suite was a little unfriendly (although I could, barely, accept it). So definitely, a fast and developer-friendly end-to-end test suite is very doable.

[VECTOR-381]: https://scylladb.atlassian.net/browse/VECTOR-381?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

nyh:
> As you may have noticed, I've written (in Python) a very big suite of end-to-end tests for Alternator vector search + the vector store. It's now around 130 test functions in 4,000 lines of code, and the entire thing takes (after my latest sleep reductions) 22 seconds. That's still not as fast as I would have hoped, but it's already fast enough to be friendly _enough_. Before my changes, 172 seconds for the test suite was a little unfriendly (although I could, barely, accept it). So definitely, a fast and developer-friendly end-to-end test suite is very doable.

By the way, you can see these tests in scylla.git, test/alternator/test_vector.py. In my opinion, they are significantly easier to write than the Rust tests (also because I'm much more familiar with Python - but I think it's not just that, I think Python genuinely is faster to write and easier to understand). They are also heavily commented which I consider important.

PR #452, opened 2026-05-18
validator: add Alternator index creation test

m-szymon:
Introduce the Alternator integration test suite with shared infrastructure (TableShape, Naming patterns, Injectors for customized AWS SDK calls) and test for basic table and index lifecycle operations: CreateTable, DescribeTable, DeleteTable.

PR #458, opened 2026-05-20
validator: add more Alternator index creation tests

m-szymon:
Extends Alternator index creation tests, verifying behaviour in several edge cases, especially validation of input paramteres.

Follow-up #452

### DIFF
--- a/crates/validator/src/alternator/create_table.rs
+++ b/crates/validator/src/alternator/create_table.rs
@@ -13,12 +13,14 @@ use aws_smithy_runtime_api::client::interceptors::context::AfterDeserializationI
 use aws_smithy_runtime_api::client::runtime_components::RuntimeComponents;
 use aws_smithy_types::config_bag::ConfigBag;
 use e2etest::TestCase;
+use httpapi::IndexInfo;
+use httpapi::IndexName;
 use std::sync::Arc;
 use std::sync::Mutex;
 use tracing::info;
 
 /// An SDK interceptor that captures the `VectorIndexes` extension field from
-/// the raw `DescribeTable` JSON response as a [`serde_json::Value`].  The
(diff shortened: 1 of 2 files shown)


crates/validator/src/alternator/create_table.rs
```diff
@@ -192,6 +194,278 @@ async fn create_describe_and_delete_table_with_vector_index(actors: TestActors)
...
+    alternator::delete_table(&client, &table_name).await;
+
+    alternator::wait_for_no_index(&vs_clients, &lower_index).await;
+    alternator::wait_for_no_index(&vs_clients, &upper_index).await;
+
+    info!("finished");
+}
+
+/// Index names are scoped to a CQL keyspace (`alternator_<table>`), so the
+/// same index name on case-distinct tables should be independent.
```
QuerthDP, line 253: Do I understand correctly that each table has its own keyspace in alternator?
And having that assumption we could create same-named indexes on different tables because they're in different keyspaces. If we had one keyspace in such case it would error out like in CQL. Is it possible to have one keyspace but two tables in Alternator?

QuerthDP (approved):
Looks good. 
I've left one comment but AFAIK it's a correct assumption, just curious.

crates/validator/src/alternator/create_table.rs
m-szymon: > Do I understand correctly that each table has its own keyspace in alternator?

Yes. DynamoDB doesn't have concept o keyspace. In Alternator we use it internally for cases exactly like this.

> Is it possible to have one keyspace but two tables in Alternator?

In Alternator keyspace we have other "special" tables - MV, CDC..., but these are not accessible as basic tables. There can be only one "base" table.

PR #459, opened 2026-05-21
validator: add Alternator UpdateTable integration tests

m-szymon:
End-to-end tests for the UpdateTable vector index lifecycle: creating an index on an empty table, on a table with preexisting data (initial-scan path), deleting an index, and verifying that non-indexable rows (wrong type, missing vector, wrong dimensions) are skipped during initial scan.
Introduces shared test infrastructure in alternator::mod — TableContext for table lifecycle management with automatic cleanup, Item builder for DynamoDB items, and ANN result verification helpers — reused across all four tests and subsequent test modules.

### DIFF
--- a/crates/validator/src/alternator/mod.rs
+++ b/crates/validator/src/alternator/mod.rs
@@ -738,14 +738,29 @@ impl TableContext {
 
     /// Creates a table, inserts items, and adds a vector index via
     /// `UpdateTable`. Waits for VS to serve the index with the correct count.
+    /// All `items` must carry a valid vector. Use
+    /// [`Self::create_with_invalid_data`] when the dataset includes items VS
+    /// should skip.
     async fn create_with_data(actors: &TestActors, shape: &TableShape, items: &[Item]) -> Self {
+        Self::create_with_invalid_data(actors, shape, items, &[]).await
+    }
+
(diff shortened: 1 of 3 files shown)


m-szymon:
Is it too big? Should I split it further?
Each commit introduces one test, with infrastructure it uses.
Following commits/PRs will be simpler as all infrastructure will already be there.

crates/validator/src/alternator/mod.rs
```diff
@@ -232,6 +285,132 @@ async fn wait_for_no_index(clients: &[HttpClient], index: &IndexInfo) {
...
+                index.keyspace,
+                index.index,
+                expected_count,
+                client.url()
+            ),
+            Duration::from_secs(60),
+        )
+        .await;
+    }
+}
```
ewienik: nit: It seems a good candidate to move to the common

ewienik (dismissed):

crates/validator/src/alternator/mod.rs
m-szymon: Moved.
Also removed unnecessary comment at `wait_for_no_index`.
And rebased to master.

ewienik (approved):

PR #462, opened 2026-05-25
vector-store: introduce fts index module with document counting

knowack1:
This PR introduces the full-text-search (FTS) index infrastructure alongside the existing vector-search index. The core change is a new `Index` enum that dispatches operations to either a `VsIndex` or `FtsIndex` actor through their respective `mpsc::Sender` channels. Callers match on the enum variant directly rather than going through an adapter with methods, keeping the dispatch explicit at each call site.

The previous `index/` directory is renamed to `vs_index/` and its actor enum is renamed from `Index` to `VsIndex` so both index types follow a consistent naming convention. A new `fts_index` module provides the `FtsIndex` actor with document add/remove/count operations and an extension trait (`FtsIndexExt`) mirroring the existing `IndexExt` pattern.

Fixes VECTOR-622


crates/vector-store/src/fts_index/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+pub(crate) fn new(key: IndexKey) -> mpsc::Sender<FtsIndex> {
+    let (tx, mut rx) = mpsc::channel::<FtsIndex>(perf::channel_size().into());
+    tokio::spawn(async move {
+        debug!("fts index actor starting for {key}");
+        let mut count: usize = 0;
+        while let Some(msg) = rx.recv().await {
+            match msg {
+                FtsIndex::AddDocument { .. } => count += 1,
+                FtsIndex::RemoveDocument { .. } => count = count.saturating_sub(1),
+                FtsIndex::RemovePartition { .. } => count = 0,
```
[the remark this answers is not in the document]
knowack1: This counting feature is temporary. The goal of it is to prove that index building infra works.

crates/vector-store/src/indexes.rs
```diff
@@ -76,7 +76,7 @@ impl From<&IndexMetadata> for RoutingGroupKey {
 
 #[derive(Debug)]
 pub(crate) struct IndexEntry {
-    index: mpsc::Sender<Index>,
+    index: Index,
```
ewienik: We should have different entries for Vs and Fts indexes in `Indexes`, to be able to query for indexes independently:
```
pub(crate) struct Indexes {                          
    vs_entries: HashMap<IndexKey, VsIndexEntry>,
    vs_routing: HashMap<RoutingGroupKey, Vec<IndexKey>>,
    fts_entries: HashMap<IndexKey, FtsIndexEntry>
}                                                    
```
For this reason consider refactoring IndexEntry into `IndexEntry<Index, Options>` or into different VsIndexEntry and FtsIndexEntry. I don't know if we will need RoutingGroupKey for Fts indexes or if we need the same fields in it (probably it will be different for multi-column target).

crates/vector-store/src/httproutes.rs
```diff
@@ -329,13 +331,17 @@ async fn get_index_status(
...
     else {
         let msg = format!("missing index: {keyspace_name}.{index_name}");
         debug!("get_index_status: {msg}");
         return (StatusCode::NOT_FOUND, msg).into_response();
     };
-    match index.count(index_key).await {
+    let count_result = match &index {
+        Index::Vs(sender) => sender.count(index_key).await,
+        Index::Fts(sender) => sender.count(index_key).await,
+    };
```
ewienik: I think we should separate status API for fts and vs. If RoutesInnerState has both `vs_indexes` and `fts_indexes` you can ask only vs indexes without this match at all.

crates/vector-store/src/httproutes.rs
ewienik, line 117: Consider change it to vs_indexes to store only Vs indexes - then there won't be need  to check index type by each request.

crates/vector-store/src/httproutes.rs
```diff
@@ -361,7 +367,10 @@ async fn get_metrics(
         let index_name = crate::IndexName::from(index_name_str);
         let key = IndexKey::new(&keyspace, &index_name);
         if let Some((index, _)) = state.engine.get_index(key.clone()).await
-            && let Ok(count) = index.count(key).await
+            && let Ok(count) = match &index {
+                Index::Vs(sender) => sender.count(key).await,
+                Index::Fts(sender) => sender.count(key).await,
+            }
```
ewienik: Consider refactoring engine.get_index into engine.get_vs_index - we should measure in this metric only vs indexes.

crates/vector-store/src/httproutes.rs
```diff
@@ -551,7 +560,19 @@ async fn post_index_ann(
...
                     return (StatusCode::BAD_REQUEST, msg).into_response();
                 }
-                (routed_key, index, primary_key_columns, table_columns)
+                let Index::Vs(index) = adapter else {
+                    timer.observe_duration();
+
+                    let msg = format!("Index {keyspace}.{index_name} does not support ANN queries");
+                    debug!("post_index_ann: {msg}");
+                    return (StatusCode::BAD_REQUEST, msg).into_response();
+                };
```
ewienik: Consider avoiding this check by distinguish in RouteInnerState between vs_indexes and fts_indexes.

crates/vector-store/src/monitor_items.rs
```diff
@@ -60,7 +61,7 @@ pub(crate) async fn new(
 
 async fn add(
     table: &Arc<RwLock<impl TableAdd>>,
-    index: &Sender<Index>,
+    index: &Index,
```
ewienik: Consider adding a new `impl AsyncFn` parameter to the add function which could take care of processing operations. Then you won't need to pass Index here and dynamically check branches. This AsyncFn could be prepared in new function.

crates/vector-store/src/httproutes.rs
```diff
@@ -24,6 +24,8 @@ use crate::node_state::NodeState;
 use crate::node_state::NodeStateExt;
 use crate::perf;
 use crate::vector;
+use crate::vs_index::IndexExt;
+use crate::vs_index::validator;
```
ewienik: consider exporting vs_index::validator::Error as vs_index::Error

crates/vector-store/src/vs_index/actor.rs
```diff
@@ -77,7 +77,7 @@ pub(crate) trait IndexExt {
     async fn count(&self, index_key: IndexKey) -> CountR;
 }
 
-impl IndexExt for mpsc::Sender<Index> {
+impl IndexExt for mpsc::Sender<VsIndex> {
```
ewienik: Refactor name into VsIndexExt

crates/vector-store/src/vs_index/factory.rs
ewienik: refactor into VsIndexConfiguration

ewienik: refactor into VsIndexFactory

crates/vector-store/src/engine.rs
```diff
@@ -296,6 +279,33 @@ async fn add_index(
...
+                    expansion_search: vs.expansion_search,
+                    space_type: vs.space_type,
+                    quantization: vs.quantization,
+                },
+                table,
+                memory,
+            )?;
+            Ok(Index::Vs(sender))
+        }
+        IndexKind::Fts(_) => Ok(Index::Fts(fts_index::new(metadata.key()))),
```
ewienik: We should have factory for FtsIndex

crates/vector-store/src/fts_index/mod.rs
```diff
@@ -0,0 +1,115 @@
...
+use crate::table::PartitionId;
+use crate::table::PrimaryId;
+use crate::vs_index::actor::CountR;
+use tokio::sync::mpsc;
+use tokio::sync::oneshot;
+use tracing::debug;
+
+/// Actor messages for the full-text-search index.
+#[allow(dead_code)]
+pub(crate) enum FtsIndex {
```
ewienik: Consider moving actor definitions into actor.rs (similar to vs_index)

crates/vector-store/src/fts_index/mod.rs
```diff
@@ -0,0 +1,115 @@
...
+        debug!("fts index actor starting for {key}");
+        while let Some(msg) = rx.recv().await {
+            if let FtsIndex::Count { tx, .. } = msg {
+                let _ = tx.send(Ok(0));
+            }
+        }
+        debug!("fts index actor finished for {key}");
+    });
+    tx
+}
```
ewienik: consider moving this into tantivy.rs and provide factory

crates/vector-store/src/fts_index/mod.rs
```diff
@@ -0,0 +1,115 @@
...
+        })
+        .await
+        .expect("internal actor should receive request");
+    }
+
+    async fn remove_partition(&self, partition_id: PartitionId) {
+        self.send(FtsIndex::RemovePartition { partition_id })
+            .await
+            .expect("internal actor should receive request");
+    }
```
ewienik: Will we have partitions in fts in short term? If not we can remove this api.

ewienik (changes requested):
Consider splitting the first commit into such commits:
- move index dir into vs_index (to preserve git history)
- refactor vs_index names
- create fts_index module (and inside mod.rs, actor.rs, factory.rs and tantivy.rs - similary to the vs_index module; in the future we could have other implementation of fts index)

crates/vector-store/tests/integration/db_basic.rs
```diff
@@ -272,7 +290,10 @@ fn process_db(db: &DbBasic, msg: Db, node_state: Sender<NodeState>) {
                                 target_column: index.metadata.target_column.clone(),
                                 partitioning: index.metadata.partitioning.clone(),
                                 filtering_columns: index.metadata.filtering_columns.clone(),
-                                kind: DbIndexKind::VectorSearch,
+                                kind: match index.metadata.kind {
+                                    IndexKind::Vs(_) => DbIndexKind::VectorSearch,
+                                    IndexKind::Fts(_) => DbIndexKind::FullTextSearch,
+                                },
```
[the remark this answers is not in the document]
knowack1: Code compiles

crates/vector-store/src/monitor_items.rs
```diff
@@ -23,13 +28,98 @@ use tracing::debug;
...
+        partition_id: PartitionId,
+        primary_id: PrimaryId,
+        in_progress: Option<AsyncInProgress>,
+    ) {
+        self.remove_document(partition_id, primary_id, in_progress)
+            .await;
+    }
+
+    async fn remove_partition(&self, _partition_id: PartitionId) {}
+}
```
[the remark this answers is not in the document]
knowack1: We do not plan to support partitions in fts indexes for now. This has been changed as requested by review.

crates/vector-store/src/indexes.rs
knowack1: Done.

crates/vector-store/src/httproutes.rs
knowack1: At this stage, I have kept it as a single endpoint. It works fine for both FTS and VS indexes. If we think it would be more elegant to split this endpoint, we can easily do so with the current implementation.

crates/vector-store/src/httproutes.rs
knowack1: Changed. The implementation retains both VS and FTS indexes but exposes methods to query each individually.

crates/vector-store/src/httproutes.rs
knowack1: Done. However, I think this metric could work well for FTS indexes as well. Is there a specific reason we should limit it to VS indexes?

crates/vector-store/src/httproutes.rs
knowack1: Fixed

crates/vector-store/src/monitor_items.rs
knowack1: Done. The implementation uses trait. Please check `trait IndexDispatch`

crates/vector-store/src/httproutes.rs
knowack1: Done

crates/vector-store/src/vs_index/actor.rs
knowack1: Done

crates/vector-store/src/vs_index/factory.rs
knowack1: Done

knowack1: Done

crates/vector-store/src/engine.rs
knowack1: Done

crates/vector-store/src/fts_index/mod.rs
knowack1: Done

crates/vector-store/src/fts_index/mod.rs
knowack1: Done

crates/vector-store/src/fts_index/mod.rs
knowack1: Done. Is planned on the roadmap but not for MS1.

knowack1:
@ewienik All comments has been addressed, please re-review latest changes.

knowack1:
> Consider splitting the first commit into such commits:
> 
> * move index dir into vs_index (to preserve git history)
> * refactor vs_index names
> * create fts_index module (and inside mod.rs, actor.rs, factory.rs and tantivy.rs - similary to the vs_index module; in the future we could have other implementation of fts index)

Done

crates/vector-store/src/engine.rs
```diff
@@ -248,54 +244,110 @@ async fn add_index(
...
-            tx.send(Err(err))
-                .unwrap_or_else(|_| trace!("add_index: unable to send response"));
-            return;
-        }
-    };
-
-    let monitor_actor = match monitor_items::new(
-        key.clone(),
+    let ctx = AddIndexContext {
+        key: &key,
```
ewienik: key is cloned in both `add_index_*` function so consider move key here. The `key` is used later in `info!` logger - this info could be moved to the start of the `add_index` function

crates/vector-store/src/engine.rs
```diff
@@ -248,54 +244,110 @@ async fn add_index(
...
+    let result = match metadata.kind.clone() {
+        IndexKind::Vs(vs_options) => {
+            add_index_vs(ctx, index_factories, vs_options, memory, metadata).await
+        }
+        IndexKind::Fts(_) => add_index_fts(ctx, index_factories).await,
+    };
+
+    match result {
+        Ok(()) => {
+            info!("creating the index {key}");
```
ewienik: It could be moved to the start of this function

crates/vector-store/src/engine.rs
```diff
@@ -248,54 +244,110 @@ async fn add_index(
...
-    .await
-    {
-        Ok(actor) => actor,
+        db_index,
+        indexes,
+    };
+
+    let result = match metadata.kind.clone() {
+        IndexKind::Vs(vs_options) => {
+            add_index_vs(ctx, index_factories, vs_options, memory, metadata).await
```
ewienik: Consider moving all parameters to the AddIndexContext (index_factories, memory, metadata). vs_options are included in metadata. In this situation you probably don't need `metada.kind.clone()` as you can use reference - you can create AddIndexContext inplace while calling `add_index_*` functions - then you can avoid cloning of IndexKind.

crates/vector-store/src/indexes.rs
```diff
@@ -74,28 +75,72 @@ impl From<&IndexMetadata> for RoutingGroupKey {
...
+    }
+}
+
+impl VsIndexEntry {
     pub(crate) async fn new(
         index: mpsc::Sender<VsIndex>,
         monitor: mpsc::Sender<MonitorItems>,
         db_index: mpsc::Sender<DbIndex>,
         metadata: IndexMetadata,
+        options: crate::IndexOptionsVs,
```
ewienik: options are also in IndexMetadata

crates/vector-store/src/httproutes.rs
ewienik: I was thinking that metrics for Fts will come further in API. Another question is: will we the same metrics for Fts and Vs? If you think the metrics for Fts will be the same as Vs and it will be calculated in the same way, then calculating here also Fts count metric make sense.

crates/vector-store/src/monitor_items.rs
ewienik: `trait IndexDispatch` seems a good approach - not the best as there are still dynamic check the type, but improvements could be done later. I'll have some comments to consider.

crates/vector-store/src/monitor_items.rs
```diff
@@ -23,13 +28,98 @@ use tracing::debug;
 use tracing::error;
 use tracing::error_span;
 
+pub(crate) trait IndexDispatch: Clone + Send + Sync + 'static {
```
ewienik: Consider removing these supertraits as dependency - it seems too strict

crates/vector-store/src/monitor_items.rs
```diff
@@ -23,13 +28,98 @@ use tracing::debug;
...
+
 pub(crate) enum MonitorItems {}
 
-pub(crate) async fn new(
+pub(crate) async fn new<I: IndexDispatch>(
     key: IndexKey,
     table: Arc<RwLock<impl TableAdd + Send + Sync + 'static>>,
     mut embeddings: Receiver<(DbIndexedRow, Option<AsyncInProgress>)>,
-    index: Sender<VsIndex>,
+    index: I,
```
ewienik: Consider using generic of Sender type: `index: mpsc::Sender<T>` to be align with common pattern of actors' handlers. It could be similar to this:
```
pub(crate) async fn new<T>(
    key: IndexKey,
    table: Arc<RwLock<impl TableAdd + Send + Sync + 'static>>,
    mut embeddings: Receiver<(DbIndexedRow, Option<AsyncInProgress>)>,
    index: mpsc::Sender<T>,
    metrics: Arc<Metrics>,
) -> anyhow::Result<Sender<MonitorItems>>
where
    T: Send + 'static,
    mpsc::Sender<T>: IndexDispatch,

```

crates/vector-store/tests/integration/db_basic.rs
```diff
@@ -44,36 +44,54 @@ pub(crate) type RxEmbeddings = mpsc::Receiver<(DbIndexedRow, Option<AsyncInProgr
 pub(crate) type TxEmbeddings = mpsc::Sender<(DbIndexedRow, Option<AsyncInProgress>)>;
 pub(crate) type ScanFn = Box<dyn FnOnce(TxEmbeddings) -> BoxFuture<'static, ()> + Send + Sync>;
 
-pub(crate) fn scan_fn(
-    items: impl IntoIterator<Item = (PrimaryKey, Option<Vector>, Timestamp)>,
-) -> ScanFn {
-    let items = Arc::new(items.into_iter().collect_vec());
+fn make_scan_fn(rows: Vec<DbIndexedRow>) -> ScanFn {
```
ewienik: IMHO there is no need to collect vectors - consider using iterator here - iterator is needed to avoid huge allocation during tests.

ewienik (changes requested):

crates/vector-store/src/httproutes.rs
knowack1: I see. Base on the https://scylladb.atlassian.net/browse/VECTOR-629 the metrics for fts are different to vs metrics. The have different metrics names (labels) specified.

crates/vector-store/src/engine.rs
knowack1: Fixed.

crates/vector-store/src/engine.rs
knowack1: Done.

crates/vector-store/src/engine.rs
knowack1: Done.
1. I created AddIndexContext two lines before the add_index_ call instead of in-place, because doing otherwise would duplicate a long block of struct initialization.
2. Now, add_index_ only accepts AddIndexContext. However, using options from the metadata means I need to unwrap kind to the proper type once again inside add_index_.

crates/vector-store/src/indexes.rs
knowack1: Fixed. Using options from metadata. However in this approach I need to unwrap options to proper type.

crates/vector-store/src/monitor_items.rs
knowack1: Ok, thanks. If agreed lets proceed with the current approach.

crates/vector-store/src/monitor_items.rs
knowack1: Fixed, Indeed these traits are not needed.

crates/vector-store/src/monitor_items.rs
knowack1: Done

crates/vector-store/tests/integration/db_basic.rs
knowack1: Fixed.

crates/vector-store/src/engine.rs
ewienik: add_index_ functions return error, so instead of unwrap you can return error

crates/vector-store/src/engine.rs
```diff
@@ -248,54 +246,98 @@ async fn add_index(
...
 
-    indexes.write().unwrap().insert(key.clone(), index_entry);
-    info!("creating the index {key}");
-    tx.send(Ok(()))
-        .unwrap_or_else(|_| trace!("add_index: unable to send response"));
+async fn add_index_vs(ctx: AddIndexContext<'_>) -> anyhow::Result<()> {
+    let options = ctx
+        .metadata
+        .vs()
+        .expect("add_index_vs must be called with a vector-search index");
```
ewienik: nit: can you refactor this by using `ok().map_err()?`?

ewienik (changes requested):

crates/vector-store/src/engine.rs
knowack1: Fixed.

crates/vector-store/src/engine.rs
knowack1: Done

ewienik (dismissed):
There are merge conflicts yet

knowack1:
> There are merge conflicts yet

Resolved

ewienik (approved):

PR #466, opened 2026-05-27
validator: add Alternator VectorSearch Query and TTL integration tests

m-szymon:
Add end-to-end tests for the Alternator VectorSearch Query extension.
Covers query correctness (index routing, result ordering, filtering),
query options (projection, select modes, limit, large dimensions), and
verifies that TTL expiration events are propagated to VS via an
index-only read path.

### DIFF
--- a/crates/validator/src/alternator/query.rs
+++ b/crates/validator/src/alternator/query.rs
@@ -12,6 +12,8 @@ use aws_sdk_dynamodb::operation::query::QueryOutput;
 use aws_sdk_dynamodb::operation::query::builders::QueryFluentBuilder;
 use aws_sdk_dynamodb::types::AttributeValue;
 use e2etest::TestCase;
+use httpapi::IndexName;
+use std::collections::HashMap;
 use tracing::info;
 
 use crate::alternator;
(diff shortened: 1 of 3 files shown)


nyh:
I have a bit of a deja-vu, didn't you already have these tests? Or is this moved from some older PRs that haven't been merged?

m-szymon:
> I have a bit of a deja-vu, didn't you already have these tests? Or is this moved from some older PRs that haven't been merged?

I was asked to split into smaller PRs and we are still merging :)

crates/validator/src/alternator/query.rs
```diff
@@ -326,6 +327,327 @@ async fn query_with_vector_search_multiple_results_ordering(actors: TestActors)
...
+    );
+
+    ctx.done().await;
+
+    info!("finished");
+}
+
+/// Verifies Limit larger than dataset returns all items without error.
+#[framed]
+async fn query_with_limit_larger_than_dataset(actors: TestActors) {
```
QuerthDP, line 513: Is there a test with limit lower than dataset?

QuerthDP (approved):
Looks good. 

You can dismiss my comments, since I believe there are tests with limit lower than item count, but not just an explicit one.

crates/validator/src/alternator/query.rs
m-szymon: Both `query_with_vector_search` and `query_uses_selected_vector_index` have 3 Items and Limit=1.

PR #468, opened 2026-05-28
validator: add LWT test coverage for Alternator

m-szymon:
Add test coverage for Vector Store indexing through the LWT/Paxos code
path on Alternator writes — conditional expressions, element-level
UpdateItem operations, and a dedicated `always_use_lwt` test suite.

### DIFF
--- a/crates/validator/src/alternator/delete_item.rs
+++ b/crates/validator/src/alternator/delete_item.rs
@@ -8,10 +8,11 @@ use crate::alternator;
 use crate::alternator::Item;
 use crate::alternator::TableContext;
 use crate::common;
+use aws_sdk_dynamodb::operation::delete_item::builders::DeleteItemFluentBuilder;
 use std::sync::Arc;
 use tracing::info;
 
-async fn ctx_delete_item(ctx: &TableContext, item: &Item) {
+fn ctx_delete_item(ctx: &TableContext, item: &Item) -> DeleteItemFluentBuilder {
     let mut req = ctx.client.delete_item().table_name(&ctx.table_name);
     let key_attrs: Vec<&str> = std::iter::once(ctx.shape.pk())
         .chain(ctx.shape.sk())
@@ -21,7 +22,7 @@ async fn ctx_delete_item(ctx: &TableContext, item: &Item) {
             req = req.key(attr_name, attr_val.clone());
         }
     }
-    req.send().await.expect("DeleteItem should succeed");
+    req
 }
 
 /// Inserts items, deletes one via `DeleteItem`, and verifies the VS index
@@ -47,11 +48,29 @@ async fn delete_item_updates_index(actors: Arc<TestActors>) {
             TableContext::create_with_data(&actors, shape, &[a.clone(), b.clone(), c.clone()])
                 .await;
 
-        info!("Deleting item from '{}'", ctx.table_name);
-        ctx_delete_item(&ctx, &a).await;
+        info!("Step 1: deleting item from '{}'", ctx.table_name);
+        ctx_delete_item(&ctx, &a)
+            .send()
+            .await
+            .expect("DeleteItem should succeed");
 
         ctx.wait_for_count(2).await;
         ctx.wait_for_ann([1.0, 1.0, 1.0], &[b.clone(), c]).await;
+
+        // Conditional DeleteItem exercises the LWT/Paxos path under
+        // `only_rmw_uses_lwt` (ConditionExpression makes it RMW).
+        info!(
+            "Step 2: conditional DeleteItem (passing condition) in '{}'",
+            ctx.table_name
+        );
+        ctx_delete_item(&ctx, &b)
+            .expression_attribute_names("#pk", ctx.shape.pk())
+            .condition_expression("attribute_exists(#pk)")
+            .send()
+            .await
+            .expect("conditional DeleteItem with passing condition should succeed");
+
+        ctx.wait_for_count(1).await;
 
         ctx.done().await;
         info!("Shape {shape:?} passed");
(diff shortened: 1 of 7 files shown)


m-szymon:
Rebased and conflicts solved.

crates/validator/src/alternator/delete_item.rs
knowack1: Consider rephrasing  this log to start with "Step 1: ...".

crates/validator/src/alternator/put_item.rs
```diff
@@ -47,6 +47,27 @@ async fn put_item_updates_index(actors: Arc<TestActors>) {
         ctx.put(&a_replaced).await;
         ctx.wait_for_ann([1.0, 1.0, 1.0], &[b, c, a_replaced]).await;
 
+        // Conditional PutItem exercises the LWT/Paxos path under
+        // `only_rmw_uses_lwt` (ConditionExpression makes it RMW).
+        info!(
+            "Step 3: conditional PutItem (passing condition) in '{}'",
```
knowack1, line 58: Consider adding log with Step 1 and Step 2 to have clear indication of test case steps.

crates/validator/src/alternator/update_item.rs
```diff
@@ -247,6 +247,149 @@ async fn update_item_with_invalid_vector_is_not_indexed(actors: Arc<TestActors>)
...
+            &ctx,
+            &valid_a,
+            "SET #vec = list_append(#vec, :val)",
+            Some(AttributeValue::L(vec![AttributeValue::N("1.0".into())])),
+            None,
+        )
+        .await,
+        "ValidationException",
+    );
+    info!("Step 6: SET #vec[2] on 'mixed'");
```
knowack1, line 335: New line before please. Also before other steps.

crates/validator/src/alternator/delete_item.rs
```diff
@@ -53,6 +53,28 @@ async fn delete_item_updates_index(actors: Arc<TestActors>) {
         ctx.wait_for_count(2).await;
         ctx.wait_for_ann([1.0, 1.0, 1.0], &[b.clone(), c]).await;
 
+        // Conditional DeleteItem exercises the LWT/Paxos path under
+        // `only_rmw_uses_lwt` (ConditionExpression makes it RMW).
+        info!(
+            "Step 2: conditional DeleteItem (passing condition) in '{}'",
```
knowack1, line 63: Consider dedicated test case.

crates/validator/src/alternator/delete_item.rs
m-szymon: Fixed.

crates/validator/src/alternator/put_item.rs
m-szymon: Fixed.

crates/validator/src/alternator/update_item.rs
m-szymon: Fixed.

crates/validator/src/alternator/delete_item.rs
m-szymon: I would stay with keeping it here. From the point of view of vector store there should be no difference - in that sense it in not a separate case. It is worth testing, because internally in the scylla there is a large divergence, so in tests focusing on scylla that would be a separate case. But focusing on vector store and API level the split is not necessary. And it would just slow it down on on table/index creation.

But I simplified it a little, by reusing same helper for building a request.

m-szymon (commented):
I changed helpers for put_item, update_item and delete_item.
Now they do not send the request, but just build them.
That way it is easier to reuse these helpers, even if we need to further extend requests (basically what is happening in these patches).
Few preexisting call sites had to be updated to call `.send()`, but I think it is even cleaner, e.g. for asserting errors.

crates/validator/src/alternator/lwt.rs
```diff
@@ -0,0 +1,252 @@
...
+struct Fixture {
+    actors: Arc<TestActors>,
+}
+
+impl e2etest::Fixture for Fixture {
+    async fn setup(setup: &mut impl e2etest::Setup) -> Self {
+        setup.setup::<TestActors>().await;
+        let actors = setup.get::<TestActors>().await.unwrap();
+
+        let mut scylla_configs = common::get_default_scylla_node_configs(&actors).await;
```
knowack1: nit: this setup looks very the same as `alternator::init`

crates/validator/src/alternator/put_item.rs
```diff
@@ -32,21 +32,47 @@ async fn put_item_updates_index(actors: Arc<TestActors>) {
...
+        let mut req = ctx
+            .client
+            .put_item()
+            .table_name(&ctx.table_name)
+            .condition_expression("attribute_not_exists(#pk)")
+            .expression_attribute_names("#pk", ctx.shape.pk());
+        for (attr_name, attr_val) in &d.0 {
+            req = req.item(attr_name, attr_val.clone());
+        }
+        req.send()
```
knowack1: nit: This code:
```
        let mut req = ctx
            .client
            .put_item()
            .table_name(&ctx.table_name)
            .condition_expression("attribute_not_exists(#pk)")
            .expression_attribute_names("#pk", ctx.shape.pk());
        for (attr_name, attr_val) in &d.0 {
            req = req.item(attr_name, attr_val.clone());
        }
```
looks like `TableContext::put`. Consider extending `TableContext::put` to accept condition_expression.

knowack1 (dismissed):

crates/validator/src/alternator/put_item.rs
m-szymon: I forgot to do that? That was the reason I modified `ctx.put()`...
Fixed.

crates/validator/src/alternator/lwt.rs
m-szymon: The difference is in one config arg.
I refactored this now.

knowack1 (approved):

PR #469, opened 2026-05-28
vector-store: fix flaky memory_limit test

ewienik:
The root cause of flakiness was the expectation that memory usage is stable during test. As it cannot be guarantee in CI or local environment sometimes test failed, because memory usage dropped during test (probably other tests in background finished).

The commit fixes this by setup memory limit to 20MB of all usage (single vector-store engine takes more than that). It divides stream of vectors into two parts - first should be not indexed, then switch off memory limit and push the second part of vectors. Only the second part should be indexes.

To check if the memory limit is observed correctly this commit adds internal counter to count memory usage above or below limit.

In the future this test could be moved to the validator as it doesn't depend on internal knowledge (it depends on internal counters available by http).

Fixes: VECTOR-667

### DIFF
--- a/crates/vector-store/tests/integration/memory_limit.rs
+++ b/crates/vector-store/tests/integration/memory_limit.rs
@@ -7,6 +7,7 @@ use crate::create_config_channels;
 use crate::db_basic;
 use crate::db_basic::Table;
 use crate::usearch::test_config;
+use futures::FutureExt;
 use httpapi::NodeStatus;
 use httpclient::HttpClient;
 use scylla::cluster::metadata::NativeType;
@@ -14,12 +15,14 @@ use scylla::value::CqlValue;
 use std::num::NonZeroUsize;
 use std::sync::Arc;
 use std::time::Duration;
-use sysinfo::System;
+use tokio::sync::mpsc;
 use tracing::info;
 use uuid::Uuid;
 use vector_store::Config;
 use vector_store::Connectivity;
 use vector_store::DbIndexPartitioning;
+use vector_store::DbIndexedRow;
+use vector_store::DbIndexedValue;
 use vector_store::ExpansionAdd;
 use vector_store::ExpansionSearch;
 use vector_store::HttpServerExt;
@@ -79,42 +82,45 @@ async fn memory_limit_during_index_build() {
     )
     .unwrap();
 
-    const VECTOR_COUNT: i32 = 1_000;
+    let (tx_outer, mut rx_outer) = mpsc::channel(1);
     db.add_index(
         index.clone(),
-        Some(db_basic::scan_fn((0..VECTOR_COUNT).map(|i| {
-            (
-                [CqlValue::Int(i)].into(),
-                Some(vec![0.0, 0.0, 0.0].into()),
-                Timestamp::from_unix_timestamp(10),
-            )
-        }))),
+        Some(Box::new(move |tx_inner| {
+            async move {
+                let mut pk = 0;
+                while let Some(item) = rx_outer.recv().await {
+                    pk += 1;
+                    let (tx_in_progress, mut rx_in_progress) = mpsc::channel(1);
+                    tx_inner
+                        .send((
+                            DbIndexedRow {
+                                primary_key: [CqlValue::Int(pk)].into(),
+                                value: Some(DbIndexedValue::Vector(item)),
+                                timestamp: Timestamp::from_unix_timestamp(10),
+                            },
+                            Some(tx_in_progress.clone().into()),
+                        ))
+                        .await
+                        .unwrap();
+                    // wait until in-progress marker is dropped
+                    drop(tx_in_progress);
+                    while rx_in_progress.recv().await.is_some() {}
+                }
+            }
+            .boxed()
+        })),
         None,
     )
     .unwrap();
 
-    // Set memory limit for Vector Store
-    let mut system_info = System::new_all();
-    system_info.refresh_memory();
-    let used_memory = if let Some(cgroup) = system_info.cgroup_limits() {
-        cgroup.rss
-    } else {
-        system_info.used_memory()
-    };
-
-    const LIMIT_MEMORY: u64 = 20 * 1024 * 1024; // 20 MB - it shouldn't be enough to index all vectors
-    let limit_memory = used_memory + LIMIT_MEMORY;
-    info!(
-        "Setting VS memory limit to {LIMIT_MEMORY} bytes, current used memory is {used_memory} bytes, "
-    );
-
-    let config = Config {
-        memory_limit: Some(limit_memory),
+    const LIMIT_MEMORY: u64 = 20 * 1024 * 1024; // 20 MB - it shouldn't be enough to index any vector
+    let mut config = Config {
+        memory_limit: Some(LIMIT_MEMORY),
         memory_usage_check_interval: Some(Duration::from_millis(10)),
         ..test_config()
     };
 
-    let (receivers, _senders) = create_config_channels(config).await;
+    let (receivers, senders) = create_config_channels(config.clone()).await;
     let index_factory = vector_store::new_index_factory_usearch(receivers.config.clone()).unwrap();
 
     let node_state = node_state.clone();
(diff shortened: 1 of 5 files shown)


crates/vector-store/tests/integration/memory_limit.rs
```diff
@@ -126,26 +132,75 @@ async fn memory_limit_during_index_build() {
...
+                .status()
+                .await
+                .is_ok_and(|status| status == NodeStatus::Serving)
+        },
+        "Waiting for index to be serving",
+    )
+    .await;
+
+    assert!(
+        dbg!(
```
m-szymon: Why `dbg!` here?

m-szymon (approved):

ewienik:
Changelog for fc350e74
- remove dbg! logging

<details>
<summary>range diff</summary>

```
1:  ff9aba26 ! 1:  fc350e74 vector-store: fix flaky memory_limit test
    @@ crates/vector-store/tests/integration/memory_limit.rs: async fn memory_limit_dur
          let client = HttpClient::new(addr);
      
     +    info!("Waiting for index to be bootstrapping");
    -     crate::wait_for(
    -         || async {
    -             client
    -                 .status()
    -                 .await
    --                .ok()
    --                .map(|status| status == NodeStatus::Serving)
    --                .unwrap_or(false)
    ++    crate::wait_for(
    ++        || async {
    ++            client
    ++                .status()
    ++                .await
     +                .is_ok_and(|status| status == NodeStatus::Bootstrapping)
    -         },
    --        "Waiting for index to be build",
    ++        },
     +        "Waiting for index to be bootstrapping",
    -     )
    -     .await;
    - 
    --    assert!(
    --        client
    --            .index_status(&index.keyspace_name.into(), &index.index_name.into())
    ++    )
    ++    .await;
    ++
     +    const VECTOR_COUNT: usize = 10;
     +
     +    info!("Send vectors to be indexed - they shouldn't be indexed, because of the memory limit");
    @@ crates/vector-store/tests/integration/memory_limit.rs: async fn memory_limit_dur
     +    for i in 0..VECTOR_COUNT {
     +        tx_outer
     +            .send(vec![i as f32, i as f32, i as f32].into())
    -             .await
    --            .unwrap()
    --            .count
    --            < VECTOR_COUNT as usize,
    --        "Expected less than {VECTOR_COUNT} vectors to be indexed"
    ++            .await
     +            .unwrap();
     +    }
     +
     +    info!("Waiting for index to be serving");
     +    drop(tx_outer);
    -+    crate::wait_for(
    -+        || async {
    -+            client
    -+                .status()
    -+                .await
    +     crate::wait_for(
    +         || async {
    +             client
    +                 .status()
    +                 .await
    +-                .ok()
    +-                .map(|status| status == NodeStatus::Serving)
    +-                .unwrap_or(false)
     +                .is_ok_and(|status| status == NodeStatus::Serving)
    -+        },
    +         },
    +-        "Waiting for index to be build",
     +        "Waiting for index to be serving",
    -+    )
    -+    .await;
    -+
    -+    assert!(
    -+        dbg!(
    -+            client
    -+                .index_status(&index.keyspace_name.into(), &index.index_name.into())
    -+                .await
    -+                .unwrap()
    -+                .count
    -+        ) == VECTOR_COUNT,
    +     )
    +     .await;
    + 
    +@@ crates/vector-store/tests/integration/memory_limit.rs: async fn memory_limit_during_index_build() {
    +             .await
    +             .unwrap()
    +             .count
    +-            < VECTOR_COUNT as usize,
    +-        "Expected less than {VECTOR_COUNT} vectors to be indexed"
    ++            == VECTOR_COUNT,
     +        "Expected only last part of vectors to be indexed"
          );
      }
```

crates/vector-store/tests/integration/memory_limit.rs
ewienik: fixed in fc350e74

crates/vector-store/tests/integration/memory_limit.rs
```diff
@@ -126,16 +132,64 @@ async fn memory_limit_during_index_build() {
...
+
+    const VECTOR_COUNT: usize = 10;
+
+    info!("Send vectors to be indexed - they shouldn't be indexed, because of the memory limit");
+    for i in 0..VECTOR_COUNT {
+        tx_outer
+            .send(vec![i as f32, i as f32, i as f32].into())
+            .await
+            .unwrap();
+    }
```
QuerthDP, line 155: We don't check that those are not indexed.

crates/vector-store/tests/integration/memory_limit.rs
```diff
@@ -145,7 +199,7 @@ async fn memory_limit_during_index_build() {
             .await
             .unwrap()
             .count
-            < VECTOR_COUNT as usize,
-        "Expected less than {VECTOR_COUNT} vectors to be indexed"
+            == VECTOR_COUNT,
+        "Expected only last part of vectors to be indexed"
```
QuerthDP, line 215: We don't know here if the first or the last part of vectors were actually indexed. Maybe we did index the first part which is also the VECTOR_COUNT size? Or maybe we did index some of those and rest of the other that summed up to VECTOR_COUNT?

QuerthDP (commented):
Looks ok, but I'm not sure if the test really expresses the intention. We should possibly add one more count which would check the index is empty before adding the second part of vectors.

Apart from that, this PR reminds me that we had planned to add a feature flag for all the `internals`. Since we start seeing it being useful I think we could focus on doing that, so we can freely add more counters in other places when needed.

ewienik:
Changelog for 41d2e906
- add additional check if vectors are not indexed

<details>
<summary>range diff</summary>

```
1:  fc350e74 ! 1:  41d2e906 vector-store: fix flaky memory_limit test
    @@ crates/vector-store/tests/integration/memory_limit.rs: async fn memory_limit_dur
     +            .unwrap();
     +    }
     +
    ++    let keyspace_name = index.keyspace_name.into();
    ++    let index_name = index.index_name.into();
    ++    assert_eq!(
    ++        client
    ++            .index_status(&keyspace_name, &index_name)
    ++            .await
    ++            .unwrap()
    ++            .count,
    ++        0,
    ++        "Expected none vector to be indexed"
    ++    );
    ++
     +    info!("Remove memory limit - it should allow to index all vectors");
     +    client
     +        .internals_start_counter("memory-usage-below-limit".to_string())
    @@ crates/vector-store/tests/integration/memory_limit.rs: async fn memory_limit_dur
          )
          .await;
      
    -@@ crates/vector-store/tests/integration/memory_limit.rs: async fn memory_limit_during_index_build() {
    +-    assert!(
    ++    assert_eq!(
    +         client
    +-            .index_status(&index.keyspace_name.into(), &index.index_name.into())
    ++            .index_status(&keyspace_name, &index_name)
                  .await
                  .unwrap()
    -             .count
    +-            .count
     -            < VECTOR_COUNT as usize,
     -        "Expected less than {VECTOR_COUNT} vectors to be indexed"
    -+            == VECTOR_COUNT,
    ++            .count,
    ++        VECTOR_COUNT,
     +        "Expected only last part of vectors to be indexed"
          );
      }
```

crates/vector-store/tests/integration/memory_limit.rs
ewienik: fixed in 41d2e906

crates/vector-store/tests/integration/memory_limit.rs
ewienik: Now in 41d2e906 we check that the first part of vectors was not indexed.

QuerthDP (dismissed):

ewienik:
Changelog for c93742ec
- fix test description

<details>
<summary>range diff</summary>

```
1:  41d2e906 ! 1:  c93742ec vector-store: fix flaky memory_limit test
    @@ crates/vector-store/tests/integration/memory_limit.rs: async fn memory_limit_dur
     +            .unwrap()
     +            .count,
     +        0,
    -+        "Expected none vector to be indexed"
    ++        "Expected all vectors not to be indexed"
     +    );
     +
     +    info!("Remove memory limit - it should allow to index all vectors");
```

QuerthDP (approved):

ewienik:
> Apart from that, this PR reminds me that we had planned to add a feature flag for all the `internals`. Since we start seeing it being useful I think we could focus on doing that, so we can freely add more counters in other places when needed.

This add to the internal counters is on the really slow path - so there is no performance degradation. We've added such flag for internals counters on the hotpath. Feature flags brings its own complexities in CI/testing/running and spaghetti code so maybe it won't be necessary to introduce this flag in all places where internal counters is used.

ewienik:
CI: github docker issue:

> Unable to find image 'scylladb/scylla-nightly@sha256:ff7411ed4eacc2205cb5b1ff92bac55f542155cd536e6dfc14cca4e87991cdd8' locally
docker: Error response from daemon: Get "https://registry-1.docker.io/v2/scylladb/scylla-nightly/manifests/sha256:ff7411ed4eacc2205cb5b1ff92bac55f542155cd536e6dfc14cca4e87991cdd8": Get "https://auth.docker.io/token?account=githubactions&scope=repository%3Ascylladb%2Fscylla-nightly%3Apull&service=registry.docker.io": net/http: request canceled (Client.Timeout exceeded while awaiting headers)

PR #473, opened 2026-06-08
validator: reject global ANN query on local-only vector index

smoczy123:
Add a regression test for VECTOR-609 verifying that a global ANN query (one without a full partition-key equality restriction) issued against a column whose only vector index is local fails instead of silently returning empty or incorrect results.

Fixes: VECTOR-609

### DIFF
--- a/crates/validator/src/filtering.rs
+++ b/crates/validator/src/filtering.rs
@@ -839,6 +839,73 @@ async fn local_index_filter_returns_no_results_when_nothing_matches(actors: Arc<
     info!("finished");
 }
 
+/// Regression test for VECTOR-609: a global ANN query (one without a full
+/// partition key equality restriction) issued against a column whose only
+/// vector index is local must fail rather than silently returning empty or
+/// incorrect results.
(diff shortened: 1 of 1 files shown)


crates/validator/src/filtering.rs
```diff
@@ -907,6 +912,66 @@ async fn local_index_filter_returns_no_results_when_nothing_matches(actors: Test
...
+        assert_eq!(index_status.count, 20, "Expected 20 vectors to be indexed");
+    }
+
+    session
+        .query_unpaged(
+            format!("SELECT pk, ck FROM {table} ORDER BY v ANN OF [1.0, 0.0, 0.0] LIMIT 20"),
+            (),
+        )
+        .await
+        .expect_err("global ANN query on a local-only vector index should fail");
```
knowack1, line 892: Should we assert specific error message?

smoczy123: added

knowack1 (approved):

PR #474, opened 2026-06-09
httpapi: serialize ANN infinities as float max

m-szymon:
Saturate non-finite ANN distances and similarity scores at the HTTP API boundary so responses stay valid JSON while preserving the sign of overflowed dot-product values. Add a focused httpapi regression test covering +inf and -inf serialization for ANN responses.

Fixes VECTOR-660.

### DIFF
--- a/crates/httpapi/src/lib.rs
+++ b/crates/httpapi/src/lib.rs
@@ -4,6 +4,8 @@
  */
 
 use macros::ToEnumSchema;
+use serde::Serialize;
+use serde::Serializer;
 use serde_json::Value;
 use std::borrow::Cow;
 use std::collections::HashMap;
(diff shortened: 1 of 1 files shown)


crates/httpapi/src/lib.rs
```diff
@@ -279,13 +289,68 @@ pub struct PostIndexAnnResponse {
...
+{
+    let value = if value == f32::INFINITY {
+        f32::MAX
+    } else if value == f32::NEG_INFINITY {
+        -f32::MAX
+    } else {
+        value
+    };
+    serializer.serialize_f32(value)
+}
```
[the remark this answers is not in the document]
m-szymon: Currently encountering NaN triggers error. While this is not the best handling, I am not intending to change that in this PR.

crates/httpapi/src/lib.rs
```diff
@@ -279,13 +289,73 @@ pub struct PostIndexAnnResponse {
...
+
+fn serialize_saturated_f32<S>(value: f32, serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: Serializer,
+{
+    let value = if value == f32::INFINITY {
+        f32::MAX
+    } else if value == f32::NEG_INFINITY {
+        -f32::MAX
```
QuerthDP: Why not `f32::MIN`?

QuerthDP (dismissed):

crates/httpapi/src/lib.rs
m-szymon: Right. I can only guess `f32::MIN` is preferred in in Rust. In C++ `min` is "smallest positive" and it is easily confused with `lowest` - `-max` is unambiguous.
I changed it.

QuerthDP (approved):

PR #476, opened 2026-06-11
vector-store: split table.rs

ewienik:
`table` module is getting bigger and bigger. There are plans for more refactoring of this module - let's split it into several submodules. This PR divides this module without business logic change.

The first commit is a git move to preserve git history. The rest of commit are simple splitting commits.

Fixes: VECTOR-688


knowack1 (approved):
Great refactor.

PR #477, opened 2026-06-15
vector-store: implement BM25 search HTTP API

knowack1:
Implements the `POST /indexes/{keyspace}/{index_name}/bm25` HTTP endpoint for full-text search using BM25 ranking. The endpoint was previously a stub returning `NOT_IMPLEMENTED`; it now resolves the requested index, validates its serving status, executes the BM25 query against the FTS engine, and returns ranked primary keys with scores.

The HTTP client gains corresponding methods for calling the new endpoint. Integration tests covering the BM25 search path are added, including error conditions and expected ranking behavior.

Closes: VECTOR-617

### DIFF
--- a/crates/vector-store/src/fts_index/tantivy.rs
+++ b/crates/vector-store/src/fts_index/tantivy.rs
@@ -315,8 +315,8 @@ pub(crate) fn new(
                 FtsIndex::Count { tx, index_key, .. } => {
                     let result = get_state(&states, table.as_ref(), &index_key)
                         .map(|s| s.reader.searcher().num_docs() as usize)
-                        .ok_or_else(|| anyhow!("fts: index not found for {index_key:?}"));
-                    let _ = tx.send(result);
+                        .unwrap_or(0);
+                    _ = tx.send(Ok(result));
                 }
                 FtsIndex::Search {
                     index_key,
@@ -325,7 +325,7 @@ pub(crate) fn new(
                     tx,
                 } => {
                     let Some(state) = get_state(&states, table.as_ref(), &index_key) else {
-                        let _ = tx.send(Err(anyhow!("fts: index not found for {index_key:?}")));
+                        _ = tx.send(Ok((vec![], vec![])));
                         continue;
                     };
                     let table = Arc::clone(&table);
@@ -333,7 +333,7 @@ pub(crate) fn new(
                         .spawn_blocking(move || {
                             let result =
                                 handle_search(&state, table.as_ref(), &index_key, &query, limit);
-                            let _ = tx.send(result);
+                            _ = tx.send(result);
                         })
                         .await;
                 }
--- a/crates/vector-store/src/httproutes.rs
+++ b/crates/vector-store/src/httproutes.rs
@@ -765,13 +765,126 @@ If TLS is enabled on the server, clients must connect using a HTTPS protocol.",
     )
 )]
 async fn post_index_bm25(
-    Path((_keyspace, _index_name)): Path<(httpapi::KeyspaceName, httpapi::IndexName)>,
+    State(state): State<RoutesInnerState>,
+    extensions: Extensions,
+    Path((keyspace, index_name)): Path<(httpapi::KeyspaceName, httpapi::IndexName)>,
+    extract::Json(request): extract::Json<httpapi::PostIndexBm25Request>,
 ) -> Response {
-    (
-        StatusCode::NOT_IMPLEMENTED,
-        "Full-text search is not yet implemented",
-    )
-        .into_response()
+    let keyspace: crate::KeyspaceName = keyspace.into();
+    let index_name: crate::IndexName = index_name.into();
+    if state.use_tls
+        && extensions
+            .get::<Protocol>()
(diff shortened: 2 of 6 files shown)


crates/vector-store/tests/integration/fts.rs
```diff
@@ -138,3 +140,203 @@ async fn fts_index_returns_proper_count() {
...
+        .map(|(pk, doc, ts)| {
+            (
+                pk.into_iter().collect(),
+                Some(doc.to_string()),
+                Timestamp::from_unix_timestamp(ts),
+            )
+        })
+        .collect();
+
+    let (run, index, _db, _node_state) = setup_fts_store(
```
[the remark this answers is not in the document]
knowack1: Fixed

crates/vector-store/tests/integration/fts.rs
```diff
@@ -138,3 +140,203 @@ async fn fts_index_returns_proper_count() {
...
+        },
+        "Waiting for FTS index to be serving",
+    )
+    .await;
+
+    (
+        client,
+        keyspace_name,
+        index_name,
+        (server, config_tx, _db, _node_state),
```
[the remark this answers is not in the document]
knowack1: Fixed.

crates/vector-store/src/httproutes.rs
```diff
@@ -765,13 +732,90 @@ If TLS is enabled on the server, clients must connect using a HTTPS protocol.",
...
+    let search_result = fts_sender
+        .search(index_key, request.query, request.limit.into())
+        .await;
+
+    match search_result {
+        Err(err) => {
+            let msg = format!("index.bm25 request error: {err}");
+            debug!("post_index_bm25: {msg}");
+            (StatusCode::INTERNAL_SERVER_ERROR, msg).into_response()
+        }
```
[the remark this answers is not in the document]
knowack1: This comment makes sense. But lets see if this will be real problem in the production. Eventually there is error message in the response, so the exact root cause of the problem will be known for the end user.

crates/vector-store/src/fts_index/tantivy.rs
```diff
@@ -280,10 +280,10 @@ pub(crate) fn new(
                         .await;
                 }
                 FtsIndex::Count { tx, index_key, .. } => {
-                    let result = get_state(&states, table.as_ref(), &index_key)
+                    let count = get_state(&states, table.as_ref(), &index_key)
                         .map(|s| s.reader.searcher().num_docs() as usize)
-                        .ok_or_else(|| anyhow!("fts: index not found for {index_key:?}"));
-                    let _ = tx.send(result);
+                        .unwrap_or(0);
```
ewienik: Does this mask a nonexistent index? If so, we shouldn't mask it.

crates/vector-store/src/httproutes.rs
```diff
@@ -765,13 +766,125 @@ If TLS is enabled on the server, clients must connect using a HTTPS protocol.",
...
+                    );
+                    debug!("post_index_bm25: {msg}");
+                    return (StatusCode::INTERNAL_SERVER_ERROR, msg).into_response();
+                }
+            }
+        }
+        (entry.index().clone(), entry.db_index())
+    };
+
+    let primary_key_columns = db_index_sender.get_primary_key_columns().await;
```
ewienik: `primary_key_columns` may be stored within `state.indexes.get_fts()` similar to vt_index.

ewienik (changes requested):

crates/vector-store/src/fts_index/tantivy.rs
knowack1: Fixed.
I think there was an implementation issue. This error concerns the `IndexState` lookup, which is created when the first document is added. Both count and search assumed that this state would already exist.

In the test case, count were called before a document was added, this should lead to returning 0 count instead of failure. I changed the count/search implementation to lazily create the state, just like we do for adding or removing a document.

crates/vector-store/src/httproutes.rs
knowack1: Done.
I move to this field to IndexEntry which holds common properties for both fts and vs indexes.

crates/vector-store/src/fts_index/tantivy.rs
ewienik: IMHO `count` and `search` shouldn't  create an index. The index ought to be created only when a document is added. If anyone asked index to get count or search it doesn't mean the index should be there.

knowack1: I moved these Tantivy-related changes regarding count/search to a dedicated commit, as they are in fact not related to the other changes in this PR: https://github.com/scylladb/vector-store/pull/477/changes/6243c03412d3f61336c0f12ed4cb36cb9da500da.

I moved back to the first implementation that returns 0 from count and Ok([],[]) from search. This behavior is the same as in usearch.rs and it seems to be the correct behavior, as this is when an index is actually created but has no documents/vectors yet.

ewienik (approved):

PR #480, opened 2026-06-19
vector-store: fix cdc reader shutdown

ewienik:
This PR provides two commits for fixing unexpected CDC reader shutdown in production. The first fixes wrong assumption that all primary key columns in CDC Row must exists (for static columns it is not true). The second fixes limited CDC restart tries - it introduces unlimited restarts for both wide and fine CDC readers.

Fixes: VECTOR-693

### DIFF
--- a/crates/validator/src/cdc.rs
+++ b/crates/validator/src/cdc.rs
@@ -5,6 +5,9 @@
 
 use crate::TestActors;
 use crate::common::*;
+use itertools::Itertools;
+use std::array;
+use std::collections::BTreeSet;
 use std::sync::Arc;
 use std::time::Duration;
 use std::time::SystemTime;
@@ -639,6 +642,88 @@ async fn cql_per_row_ttl_expires_from_index(actors: Arc<TestActors>) {
         .query_unpaged(format!("DROP KEYSPACE {keyspace}"), ())
         .await
         .expect("failed to drop a keyspace");
+
+    info!("finished");
+}
+
+/// Check if null clustering keys are skipped by the index.
+///
+/// Use static columns as updating a static column gives a CDC's row with null clustering key.
+///
+/// Steps:
+/// 1. Create a table with a clustering key and a static column.
+/// 2. Create an index on the vector column.
+/// 3. Insert rows with a clustering key and a static column value.
+/// 4. Verify that the index count is correct and the ANN query returns the inserted rows.
+/// 5. Drop the keyspace and verify that the index is dropped.
+#[e2etest::test(group = cdc)]
+async fn skip_null_ck(actors: Arc<TestActors>) {
+    info!("started");
+
+    let (session, clients) = prepare_connection_single_vs(&actors).await;
+    let client = clients.first().unwrap();
+    let keyspace = create_keyspace(&session).await;
+    let table = create_table(
+        &session,
+        "pk INT, ck INT, v VECTOR<FLOAT, 3>, s INT STATIC, PRIMARY KEY (pk, ck)",
+        None,
+    )
+    .await;
+    let index = create_index(CreateIndexQuery::new(&session, &clients, &table, "v")).await;
+
+    let status = wait_for_index(client, &index).await;
+    assert_eq!(status.count, 0, "Index should start empty");
+
+    info!("Insert data to the table for CDC");
+    let pks: BTreeSet<_> = array::from_fn::<i32, 10, _>(|i| i as i32)
+        .into_iter()
+        .collect();
+    for pk in &pks {
+        session
+            .query_unpaged(
+                format!("INSERT INTO {table} (pk, ck, v, s) VALUES ({pk}, 2, [1.0, 2.0, 3.0], 3)"),
+                (),
+            )
+            .await
+            .expect("failed to insert data");
+    }
+
+    wait_for(
+        || async {
+            let status = client.index_status(&index.keyspace, &index.index).await;
+            matches!(status, Ok(s) if s.count == pks.len())
+        },
+        "Waiting for all rows to be indexed",
+        DEFAULT_OPERATION_TIMEOUT,
+    )
+    .await;
+
+    // Verify ANN query returns the inserted row
+    let rows = wait_for_value(
+        || async {
+            let result = get_opt_query_results(
+                format!("SELECT pk FROM {table} ORDER BY v ANN OF [1.0, 2.0, 3.0] LIMIT 100"),
+                &session,
+            )
+            .await?;
+            let rows = result
+                .rows::<(i32,)>()
+                .ok()?
+                .map_ok(|row| row.0)
+                .collect::<Result<BTreeSet<_>, _>>()
+                .ok()?;
+            (rows.len() == pks.len()).then_some(rows)
+        },
+        "Waiting for ANN query after CDC insert",
+        DEFAULT_OPERATION_TIMEOUT,
+    )
+    .await;
+    assert_eq!(rows, pks);
+
+    session
+        .query_unpaged(format!("DROP KEYSPACE {keyspace}"), ())
+        .await
+        .expect("failed to drop keyspace");
 
     info!("finished");
 }
--- a/crates/validator/src/cdc.rs
+++ b/crates/validator/src/cdc.rs
@@ -5,7 +5,14 @@
 
 use crate::TestActors;
 use crate::common::*;
+use bytes::BytesMut;
+use e2etest_scylla_proxy_cluster::ScyllaProxyClusterExt;
 use itertools::Itertools;
+use scylla_proxy::Condition;
+use scylla_proxy::Reaction;
+use scylla_proxy::ResponseOpcode;
+use scylla_proxy::ResponseReaction;
+use scylla_proxy::ResponseRule;
 use std::array;
 use std::collections::BTreeSet;
 use std::sync::Arc;
@@ -19,13 +26,17 @@ const CDC_MAX_LATENCY: Duration = Duration::from_secs(60);
 const CDC_ACTOR_STOP_TIMEOUT: Duration = Duration::from_secs(10);
 const TTL_EXPIRATION_TIMEOUT: Duration = Duration::from_secs(10);
 
-e2etest::group!(name = cdc, fixtures = (Fixture), parent = crate::validator);
+e2etest::group!(
+    name = cdc_direct,
+    fixtures = (FixtureDirect),
+    parent = crate::validator
+);
 
-struct Fixture {
+struct FixtureDirect {
     actors: Arc<TestActors>,
 }
 
-impl e2etest::Fixture for Fixture {
+impl e2etest::Fixture for FixtureDirect {
     async fn setup(setup: &mut impl e2etest::Setup) -> Self {
         setup.setup::<TestActors>().await;
         let actors = setup.get::<TestActors>().await.unwrap();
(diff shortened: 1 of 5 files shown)


crates/validator/src/cdc.rs
```diff
@@ -642,3 +645,75 @@ async fn cql_per_row_ttl_expires_from_index(actors: Arc<TestActors>) {
...
     info!("finished");
 }
+
+/// Check if null clustering keys are skipped by the index.
+///
+/// Use static columns as updating a static column gives a CDC's row with null clustering key.
+///
+/// Steps:
+/// 1. Create a table with a clustering key and a static column.
+/// 2. Create an index on the static column.
```
m-szymon: The index is not created on static column, right? It is created on vector column.

crates/validator/src/cdc.rs
```diff
@@ -642,3 +645,75 @@ async fn cql_per_row_ttl_expires_from_index(actors: Arc<TestActors>) {
 
     info!("finished");
 }
+
+/// Check if null clustering keys are skipped by the index.
+///
+/// Use static columns as updating a static column gives a CDC's row with null clustering key.
```
m-szymon, line 686: I am not sure if I understand how this test works.
Does the `INSERT INTO {table} (pk, ck, v, s) VALUES ({pk}, 2, [1.0, 2.0, 3.0], 3)` result in two CDC events?
One with `pk,ck,v` and second with `pk,null,s`?
Would that test fail if we didn't skip ck=null?

ewienik:
Changelog for 612c63204
- reset backoff counter after session restart

<details>
<summary>range diff</summary>

```
1:  e7717a82a ! 1:  612c63204 vector-store: refactor cdc restart behaviour
    @@ crates/vector-store/src/db_cdc.rs: pub(crate) fn new(
     +            let mut err_counter = 0usize;
     +
                  loop {
    -+
                      tokio::select! {
                          // Shut down when all senders are dropped
    -                     _ = rx.recv() => { break; }
     @@ crates/vector-store/src/db_cdc.rs: pub(crate) fn new(
    +                             session_opt, &config_rx, &metadata,
    +                             &tx_embeddings, &internals,
    +                         ).await;
    ++                        err_counter = 0;
    ++                        backoff_duration = None;
                          }
      
                          _ = reader.error_notify.notified() => {
```

crates/vector-store/src/db_cdc.rs
```diff
@@ -161,15 +155,21 @@ pub(crate) fn new(
...
                     }
 
                     _ = reader.error_notify.notified() => {
-                        if reader.handle_error() {
-                            break;
-                        }
+                        err_counter += 1;
+                        backoff_duration = Some(DEFAULT_BACKOFF_DURATION);
+                        warn!(
+                            "{name} CDC reader error {err_counter}, restarting after {duration:?} backoff",
```
m-szymon, line 167: nit: if I understand correctly, with several errors within same backoff window, each will produce error message, all stating "restarting after {duration}", but there will be one restart, right?

crates/validator/src/cdc.rs
QuerthDP, line 1: nit: made CdcReader task finished with error -> made CdcReader task finish with error

QuerthDP (commented):
Conceptually it looks fine.
It should fix the bug that was observed in production and additionally changes the policy of restarting CDC which was wrong for stopping the restart after 3 retries.
I'm thinking of possibly changing that to exponential retry, but 5s each is fine as well.

I'll continue the review later with the more detailed look.

m-szymon (approved):
Fix seems correct, I am a little hesitant about tests.

Other related questions:
Can you confirm what happens with other CDC events? I believed partition deletes and range deletes were simply ignored (https://scylladb.atlassian.net/browse/VECTOR-647), but not I don't know.
Another issue I never tested before, although I thought about it. Can we have index on static vector column? Now I think not, but do we prevent it in any way?

crates/validator/src/cdc.rs
ewienik: Yes, it will fail, because we try to retrieve primary key + target for every CDC Row. So for the second one with ck = null the current version would fail

crates/vector-store/src/db_cdc.rs
ewienik: No, error handling here means CDC Reader isn't already working, so there will be only one such log message.

ewienik:
> Other related questions: Can you confirm what happens with other CDC events? I believed partition deletes and range deletes were simply ignored (https://scylladb.atlassian.net/browse/VECTOR-647), but not I don't know. Another issue I never tested before, although I thought about it. Can we have index on static vector column? Now I think not, but do we prevent it in any way?

The processing CDC row is simple now - we get target column (Option) and primary key columns. If any of the primary key columns doesn't exists it means it is not correct update of the value.

crates/vector-store/src/db_cdc.rs
ewienik: Use `is_deleted()` to detect if the target column exists in the row - as there could be update on different column, not target column

crates/vector-store/src/db_cdc.rs
```diff
@@ -536,22 +536,16 @@ impl Consumer for CdcConsumer {
...
-                }
-                row.take_value(column.as_ref()).ok_or(anyhow!(
-                    "CDC error: primary key column {column} value should exist"
-                ))
-            })
-            .collect::<anyhow::Result<_>>()?;
+            .map(|column| row.take_value(column.as_ref()))
+            .collect()
+        else {
+            // If any primary key column is missing skip this row.
```
knowack1, line 497: I recommend debug log here. Could be essential for investigating why certain  rows/values are not seen in the index.

crates/validator/src/cdc.rs
```diff
@@ -642,3 +645,75 @@ async fn cql_per_row_ttl_expires_from_index(actors: Arc<TestActors>) {
...
+///
+/// Steps:
+/// 1. Create a table with a clustering key and a static column.
+/// 2. Create an index on the static column.
+/// 3. Insert rows with a clustering key and a static column value.
+/// 4. Verify that the index count is correct and the ANN query returns the inserted rows.
```
knowack1, line 692: "Verify that the index count is correct..."  - I do not see this in test case. I see only the second part "...and the ANN query returns the inserted rows".

crates/vector-store/src/db_cdc.rs
```diff
@@ -55,12 +56,8 @@ const DEFAULT_CONSISTENT_SLEEP_INTERVAL: Duration = Duration::from_secs(10);
...
 const DEFAULT_REALTIME_SLEEP_INTERVAL: Duration = Duration::from_millis(500);
 
-/// Maximum consecutive errors before a CDC reader with Backoff policy escalates to session teardown.
-const DEFAULT_MAX_CONSECUTIVE_ERRORS: u32 = 3;
-
-/// Base backoff duration for CDC reader restarts after errors.
-/// The actual backoff is `DEFAULT_BACKOFF_BASE * consecutive_error_count`.
-const DEFAULT_BACKOFF_BASE: Duration = Duration::from_secs(5);
+/// Default backoff duration for CDC reader restarts after errors.
+const DEFAULT_BACKOFF_DURATION: Duration = Duration::from_secs(5);
```
knowack1, line 61: I think exponential backoff makes sense for CDC recovery. Consider restoring it.  Could apply to both wide and fine CDC readers.

crates/validator/src/cdc.rs
```diff
@@ -717,3 +752,147 @@ async fn skip_null_ck(actors: Arc<TestActors>) {
...
+    let wide_stopped = format!("{index_ident}-wide-cdc-handler-stopped");
+    let fine_started = format!("{index_ident}-fine-cdc-handler-started");
+    let fine_stopped = format!("{index_ident}-fine-cdc-handler-stopped");
+
+    client.internals_clear_counters().await.unwrap();
+    for name in [&wide_started, &wide_stopped, &fine_started, &fine_stopped] {
+        client.internals_start_counter(name.clone()).await.unwrap();
+    }
+
+    let log_counters = async || {
```
knowack1, line 955: Consider a standalone function to avoid making the test case function unnecessarily long.

crates/vector-store/src/db_cdc.rs
ewienik: In case there is missing primary key column we are not able to track the specific row that was not processed.

crates/vector-store/src/db_cdc.rs
ewienik: IMHO, we need to try restarting forever. So we would need to limit exponentiality at a reasonable value, so eventually we will have non-exponential delay. To simplify implementation we could stay with const delay, as it won't overwhelm server.

crates/validator/src/cdc.rs
ewienik: I considered this, but this function needs to retrieve 5 references: client and 4 names of counters, that make each call of the log_counter function multi-line call. Eventually it makes function not shorter but more complicated.

ewienik:
Changelog for e657f93e1
- rebase to master
- add commit: fix cdc processing with null target
- fix comments
- add check for index size

<details>
<summary>range diff</summary>

```
1:  8961e31c1 ! 1:  347dcd9e2 vector-store: fix cdc processing with null pk
    @@ Commit message
     
         CDC Row can store null for clustering key in specific situations - and this is
         not an error. The db_cdc module handled this situation as an error and made
    -    CdcReader task finished with error. This commit fixes this by collecting all
    +    CdcReader task finish with error. This commit fixes this by collecting all
         primary columns - if any is missing it means the row should be skipped.
     
      ## crates/validator/src/cdc.rs ##
    @@ crates/validator/src/cdc.rs: async fn cql_per_row_ttl_expires_from_index(actors:
     +///
     +/// Steps:
     +/// 1. Create a table with a clustering key and a static column.
    -+/// 2. Create an index on the static column.
    ++/// 2. Create an index on the vector column.
     +/// 3. Insert rows with a clustering key and a static column value.
     +/// 4. Verify that the index count is correct and the ANN query returns the inserted rows.
     +/// 5. Drop the keyspace and verify that the index is dropped.
    @@ crates/validator/src/cdc.rs: async fn cql_per_row_ttl_expires_from_index(actors:
     +    let status = wait_for_index(client, &index).await;
     +    assert_eq!(status.count, 0, "Index should start empty");
     +
    -+    // Insert data to the table for CDC
    ++    info!("Insert data to the table for CDC");
     +    let pks: BTreeSet<_> = array::from_fn::<i32, 10, _>(|i| i as i32)
     +        .into_iter()
     +        .collect();
    @@ crates/validator/src/cdc.rs: async fn cql_per_row_ttl_expires_from_index(actors:
     +            .expect("failed to insert data");
     +    }
     +
    ++    wait_for(
    ++        || async {
    ++            let status = client.index_status(&index.keyspace, &index.index).await;
    ++            matches!(status, Ok(s) if s.count == pks.len())
    ++        },
    ++        "Waiting for all rows to be indexed",
    ++        DEFAULT_OPERATION_TIMEOUT,
    ++    )
    ++    .await;
    ++
     +    // Verify ANN query returns the inserted row
     +    let rows = wait_for_value(
     +        || async {
-:  --------- > 2:  3a404bed6 vector-store: fix cdc processing with null target
2:  612c63204 ! 3:  e657f93e1 vector-store: refactor cdc restart behaviour
    @@ crates/validator/src/cdc.rs: async fn cql_per_row_ttl_expires_from_index(actors:
          info!("started");
      
     @@ crates/validator/src/cdc.rs: async fn skip_null_ck(actors: Arc<TestActors>) {
    + /// 6. Verify that the index count after index and the ANN query returns the updated and inserted
    + ///    rows.
    + /// 7. Drop the keyspace and verify that the index is dropped.
    +-#[e2etest::test(group = cdc)]
    ++#[e2etest::test(group = cdc_direct)]
    + async fn skip_null_target(actors: Arc<TestActors>) {
    +     info!("started");
    + 
    +@@ crates/validator/src/cdc.rs: async fn skip_null_target(actors: Arc<TestActors>) {
      
          info!("finished");
      }
    @@ crates/validator/src/cdc.rs: async fn skip_null_ck(actors: Arc<TestActors>) {
     +}
     
      ## crates/vector-store/src/db_cdc.rs ##
    -@@ crates/vector-store/src/db_cdc.rs: use scylla_cdc::consumer::CDCRow;
    - use scylla_cdc::consumer::Consumer;
    +@@ crates/vector-store/src/db_cdc.rs: use scylla_cdc::consumer::Consumer;
      use scylla_cdc::consumer::ConsumerFactory;
    + use scylla_cdc::consumer::OperationType;
      use scylla_cdc::log_reader::CDCLogReaderBuilder;
     +use std::future;
      use std::sync::Arc;
```

crates/validator/src/cdc.rs
ewienik: fixed in e657f93e1

crates/validator/src/cdc.rs
ewienik: fixed in e657f93e1

crates/vector-store/src/db_cdc.rs
ewienik: fixed in e657f93e1

crates/validator/src/cdc.rs
ewienik: fixed in e657f93e1

knowack1 (approved):

PR #482, opened 2026-06-22
vector-store: monitor CDC indexing lag for vector search index

smoczy123:
Add an `indexing_lag_seconds` histogram metric that records the time
between when a change was written in ScyllaDB (its CDC writetime) and
when it is indexed in the vector store. The lag is observed per indexed
value, labelled by keyspace and index name.

Only CDC-driven changes are measured: rows replayed during the initial
full scan carry an in-progress marker, so they are excluded.

Fixes: VECTOR-684

### DIFF
--- a/crates/vector-store/src/monitor_items.rs
+++ b/crates/vector-store/src/monitor_items.rs
@@ -37,14 +37,14 @@ pub(crate) trait IndexDispatch {
         partition_id: PartitionId,
         primary_id: PrimaryId,
         value: DbIndexedValue,
-        in_progress: Option<AsyncInProgress>,
+        in_progress: AsyncInProgress,
     ) -> impl Future<Output = ()> + Send;
 
     fn remove_value(
         &self,
         partition_id: PartitionId,
         primary_id: PrimaryId,
-        in_progress: Option<AsyncInProgress>,
+        in_progress: AsyncInProgress,
     ) -> impl Future<Output = ()> + Send;
 
     fn remove_partition(&self, partition_id: PartitionId) -> impl Future<Output = ()> + Send;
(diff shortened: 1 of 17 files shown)


crates/vector-store/src/monitor_items.rs
```diff
@@ -188,6 +195,13 @@ async fn add<I: IndexDispatch>(
                     .modified
                     .with_label_values(&[key.keyspace().as_ref(), key.index().as_ref(), op_label])
                     .inc();
+                if from_cdc {
+                    metrics.observe_indexing_lag(
+                        key.keyspace().as_ref(),
+                        key.index().as_ref(),
+                        change_timestamp,
+                    );
+                }
```
ewienik: This is not the the point, where the vector is added. Consider using for this purpose AsyncInProgress by refactor it into
```
pub enum AsyncInProgress {
    None,
    Fullscan(mpsc::Sender<()>,
    Cdc(Arc<Metrics>, Timestamp>,
}
```
Consider refactoring this type into a separate module, implement Drop for it and for Cdc calculate metrics in drop. Try to find the smallest size_of for this AsyncInProgress.

ewienik (changes requested):

smoczy123:
Changelog:

- rebased onto master
- refactored according to @ewienik comment

smoczy123:
It looks like a lot of changes but most of them are cause by refactor of AsyncInProgress

crates/vector-store/src/monitor_items.rs
ewienik: Changes LGTM

ewienik (approved):
We will need to measure impact of retrieving current time for each CDC insert. But it could be measured later.

PR #493, opened 2026-06-29
vector-store/validator: fix silently failing tests

knowack1:
The validator was returning a success exit code on test failures, which meant failing tests were not noticed in CI. Once the exit code was corrected, the following test cases were revealed to be failing:

- `types::query_with_binary_key`
- `types::query_with_number_key`
- `types::query_with_string_key`
- `types::query_with_optimized_vector_type`
- `lwt::alternator_with_always_use_lwt`
- `update_item::update_item_with_invalid_vector_is_not_indexed`
- `put_item::put_item_updates_index`
- `put_item::put_item_with_invalid_vector_is_not_indexed`
- `batch_write_item::batch_write_item_updates_index`
- `batch_write_item::batch_write_item_with_invalid_vector`
- `query::query_with_large_dimensions`
- `filtering::global_ann_query_on_local_only_index_fails`

This PR fixes the exit code handling and all of the above test cases.

### DIFF
--- a/crates/validator/src/lib.rs
+++ b/crates/validator/src/lib.rs
@@ -38,6 +38,7 @@ use e2etest_vector_store_cluster::VectorStoreClusterExt;
 use std::env;
 use std::net::Ipv4Addr;
 use std::path::PathBuf;
+use std::process::ExitCode;
 use std::sync::Arc;
 use tokio::runtime::Handle;
 use tokio::sync::mpsc;
@@ -170,7 +171,7 @@ fn validate_different_subnet(dns_ip: Ipv4Addr, base_ip: Ipv4Addr) {
 
 e2etest::group!(name = validator, fixtures = (TestActors));
 
-pub async fn run() {
+pub async fn run() -> ExitCode {
     let args = Args::parse();
     let root = validator();
     let stats = match args.command {
@@ -178,7 +179,7 @@ pub async fn run() {
             root.test_names().into_iter().for_each(|name| {
                 println!("{name}");
             });
-            return;
+            return ExitCode::SUCCESS;
         }
         Command::Run(args) => e2etest::run(init(args), root).await,
     };
@@ -202,7 +203,10 @@ pub async fn run() {
             "{error_list}",
             error_list = format_failed_names(&stats.failed_names())
         );
+        return ExitCode::FAILURE;
     }
+
+    ExitCode::SUCCESS
 }
 
 pub(crate) fn format_failed_names(failed_names: &[String]) -> String {
--- a/crates/vector-store/src/db_index_backend.rs
+++ b/crates/vector-store/src/db_index_backend.rs
@@ -20,8 +20,21 @@ use scylla::client::session::Session;
 use scylla::statement::prepared::PreparedStatement;
 use scylla::value::CqlValue;
 use scylla_cdc::CqlIdentifier;
+use scylla_cdc::consumer::CDCRow;
 use std::collections::BTreeMap;
 use std::num::NonZeroUsize;
+
+/// Alternator tables store all user attributes in a single map column `:attrs`.
+/// Because Alternator (DynamoDB-compatible) is schemaless, different items can
+/// have different types for the same attribute name, so a map is used instead
+/// of dedicated typed columns.
+const ALTERNATOR_ATTRS_COLUMN: &str = ":attrs";
+
+pub(crate) enum CdcValueStatus {
+    NewValue(CqlValue),
+    Deleted,
+    Skip,
+}
 
 pub(crate) struct IndexLocation {
     pub keyspace: KeyspaceName,
@@ -49,7 +62,7 @@ impl DbIndexBackend {
     pub fn target_column_name(&self) -> &str {
         match self {
             Self::Cql { target_column } => target_column.as_ref(),
-            Self::Alternator { .. } => ":attrs",
+            Self::Alternator { .. } => ALTERNATOR_ATTRS_COLUMN,
         }
     }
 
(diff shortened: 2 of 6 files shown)


crates/vector-store/src/db_index_backend.rs
```diff
@@ -76,6 +83,41 @@ impl DbIndexBackend {
...
+                    Some(value) => CdcValueStatus::NewValue(value),
+                    None => CdcValueStatus::Skip,
+                }
+            }
+            Self::Alternator { target_column } => {
+                // In Alternator, the vector attribute is stored in the `:attrs` map column.
+                let column = ":attrs";
+                // Check If the `:attrs` map column is deleted, the target_column value is also considered deleted.
+                if row.is_value_deleted(column) {
+                    return CdcValueStatus::Deleted;
```
m-szymon: With PutItem :attrs is first deleted and than added, but this is single CDC event.
So it can have both is_value_deleted and a value. We should check value first.

crates/validator/src/lib.rs
```diff
@@ -170,15 +170,15 @@ fn validate_different_subnet(dns_ip: Ipv4Addr, base_ip: Ipv4Addr) {
 
 e2etest::group!(name = validator, fixtures = (TestActors));
 
-pub async fn run() {
+pub async fn run() -> std::process::ExitCode {
```
ewienik: consider `use std::process::ExitCode`

crates/vector-store/src/db_index_backend.rs
```diff
@@ -76,6 +83,44 @@ impl DbIndexBackend {
...
+                    None => CdcValueStatus::Skip,
+                }
+            }
+            Self::Alternator { target_column } => {
+                // Check take_value before is_value_deleted.
+                // For map columns, is_value_deleted can be true even when new data is present.
+                let column = ":attrs";
+                match row.take_value(column) {
+                    Some(value) => CdcValueStatus::NewValue(value),
+                    None if row.is_value_deleted(column) => CdcValueStatus::Deleted,
```
ewienik, line 113: Should it be before take_value as a first check?

crates/vector-store/src/db_index_backend.rs
```diff
@@ -76,6 +83,41 @@ impl DbIndexBackend {
...
+                    return CdcValueStatus::Deleted;
+                }
+                match row.take_value(column) {
+                    Some(value) => CdcValueStatus::NewValue(value),
+                    None => CdcValueStatus::Skip,
+                }
+            }
+            Self::Alternator { target_column } => {
+                // In Alternator, the vector attribute is stored in the `:attrs` map column.
+                let column = ":attrs";
```
m-szymon: `target_column_name()` already defines ":attrs" literal. Any way to reuse it?

ewienik (changes requested):

crates/vector-store/src/db_index_backend.rs
knowack1: That was my initial implementation be this does not work this way. I put some comment on this:
```
 // Check take_value before is_value_deleted.
// For map columns, is_value_deleted can be true even when new data is present.
```

crates/vector-store/src/db_index_backend.rs
```diff
@@ -76,6 +83,40 @@ impl DbIndexBackend {
...
+                    Some(value) => CdcValueStatus::NewValue(value),
+                    None => CdcValueStatus::Skip,
+                }
+            }
+            Self::Alternator { target_column } => {
+                // Check take_value before is_value_deleted.
+                // For map columns, is_value_deleted can be true even when new data is present.
+                let column = ":attrs";
+                match row.take_value(column) {
+                    Some(value) => CdcValueStatus::NewValue(value),
```
m-szymon, line 112: I think ":attrs" can have a value with some unrelated attributes, but the target vector attribute may still be absent or deleted.

crates/vector-store/src/db_index_backend.rs
ewienik: Consider providing better comment describing the problem

m-szymon (commented):
To be extra pedantic commits should be in reverse order.
After first commit tests start failing.

crates/vector-store/src/db_index_backend.rs
knowack1: There is also a 3rd place where this occurs. I extracted this into a dedicated constant with an explanatory comment. I hesitated to use `target_columns_name()` directly in this function, as having two target columns would be misleading (the one coming from `Self::Alternator { target_column }`).

crates/vector-store/src/db_index_backend.rs
knowack1: Fixed.

crates/vector-store/src/db_index_backend.rs
knowack1: This comment makes sense and you are probably right, but I haven't checked this yet. Because all current test cases pass with the current code, I propose handling this problem in a dedicated Jira ticket with proper test case coverage.

crates/validator/src/lib.rs
knowack1: Done

crates/vector-store/src/db_index_backend.rs
```diff
@@ -76,6 +89,42 @@ impl DbIndexBackend {
...
+                // so is_value_deleted can be true even when a new value is present.
+                // We must check for a value first to avoid misreporting it as deleted.
+                let column = ALTERNATOR_ATTRS_COLUMN;
+                match row.take_value(column) {
+                    Some(value) => CdcValueStatus::NewValue(value),
+                    None if row.is_value_deleted(column) => CdcValueStatus::Deleted,
+                    None => {
+                        let target_deleted = row.take_deleted_elements(column).iter().any(|el| {
+                            el.as_text().map(String::as_str) == Some(target_column.as_ref())
+                        });
```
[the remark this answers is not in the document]
knowack1: CqlValue::as_text() returns Option<String>

crates/vector-store/src/indexes.rs
```diff
@@ -394,7 +397,17 @@ impl Indexes {
...
+                let has_serving = candidates
+                    .iter()
+                    .filter_map(|key| self.vs_entries.get(key))
+                    .any(|entry| entry.status == IndexStatus::Serving);
+                if has_serving {
+                    BestIndexState::NoGlobalIndex
+                } else {
+                    BestIndexState::NotServing(requested_entry.progress)
+                }
+            }
```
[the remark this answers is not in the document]
knowack1: That is true, but this change doesn't make it worse. Currently, in those other cases, the user still gets a misleading (non-serving) message. We can handle more sophisticated cases later if needed.

ewienik (approved):

PR #495, opened 2026-06-30
fix: Alternator CDC handling for vector attribute updates

m-szymon:
Fix Alternator CDC processing so vector index updates are based on the indexed attribute inside `:attrs`, not on any physical `:attrs` map change.

Alternator stores user attributes in `:attrs` as `map<text, blob>` in CQL terms, represented internally as `map<utf8, bytes>`. Because of that, CDC events for unrelated attributes can still modify the same physical `:attrs` column. The vector-store CDC path now extracts the configured indexed attribute first, then decides whether the event is a new value, deletion, or should be skipped.

This prevents unrelated Alternator `UpdateItem`/`PutItem` changes from de-indexing vectors, while preserving replacement semantics for `PutItem`.

This PR also adds validator coverage for Alternator `PutItem` and `UpdateItem` CDC edge cases, including invalid vector replacements, unrelated attribute updates, mixed target/unrelated attribute deltas, and target attribute deletion. It also removes the obsolete whole-`:attrs` vector extractor so Alternator schema mapping stays in the index backend and vector parsing only handles logical vector values.

### DIFF
--- a/crates/validator/src/alternator/put_item.rs
+++ b/crates/validator/src/alternator/put_item.rs
@@ -86,80 +86,63 @@ async fn put_item_with_invalid_vector_is_not_indexed(actors: Arc<TestActors>) {
 
     let ctx = TableContext::create(&actors, shape).await;
 
-    let vec_with_string_elem = AttributeValue::L(vec![
-        AttributeValue::N("1.0".into()),
-        AttributeValue::N("2.0".into()),
-        AttributeValue::S("3.0".into()),
-    ]);
-    let vec_with_null_elem = AttributeValue::L(vec![
-        AttributeValue::N("1.0".into()),
-        AttributeValue::N("2.0".into()),
-        AttributeValue::Null(true),
-    ]);
-
     let no_vec = Item::key(pk, None, "pk", "no-vec");
     ctx.put(&no_vec)
         .send()
         .await
         .expect("PutItem should succeed");
 
-    let wrong_type_string = Item::key(pk, None, "pk", "wrong-type-string")
-        .attr(vec_attr, AttributeValue::S("not-a-vector".into()));
-    alternator::assert_service_error(
-        ctx.put(&wrong_type_string).send().await,
-        "ValidationException",
-    );
-
-    let wrong_type_mostly_float_one_s = Item::key(pk, None, "pk", "wrong-type-mostly-float-one-s")
-        .attr(vec_attr, vec_with_string_elem);
-    alternator::assert_service_error(
-        ctx.put(&wrong_type_mostly_float_one_s).send().await,
-        "ValidationException",
-    );
-
-    let wrong_type_mostly_float_one_null =
-        Item::key(pk, None, "pk", "wrong-type-mostly-float-one-null")
-            .attr(vec_attr, vec_with_null_elem);
-    alternator::assert_service_error(
-        ctx.put(&wrong_type_mostly_float_one_null).send().await,
-        "ValidationException",
-    );
-
-    let wrong_type_too_short = Item::key(pk, None, "pk", "wrong-type-too-short")
-        .attr(vec_attr, alternator::float_list([1.0_f32, 1.0]));
-    alternator::assert_service_error(
-        ctx.put(&wrong_type_too_short).send().await,
-        "ValidationException",
-    );
-
-    let wrong_type_too_long = Item::key(pk, None, "pk", "wrong-type-too-long")
-        .attr(vec_attr, alternator::float_list([1.0_f32, 1.0, 1.0, 1.0]));
-    alternator::assert_service_error(
-        ctx.put(&wrong_type_too_long).send().await,
-        "ValidationException",
-    );
-
-    // valid is put last so that wait_for_ann acts as a sequencing barrier:
-    // when VS has indexed valid it must have already processed the no_vec CDC
-    // event before it (CDC events are ordered), proving that the item without
-    // a vector attribute was correctly ignored.
-    let valid = Item::key(pk, None, "pk", "valid").vec(vec_attr, [1.0, 1.0, 1.0]);
-    ctx.put(&valid)
+    let try_replace_valid =
+        Item::key(pk, None, "pk", "try-replace-valid").vec(vec_attr, [1.0, 1.0, 1.0]);
+    ctx.put(&try_replace_valid)
         .send()
         .await
         .expect("PutItem should succeed");
-    ctx.wait_for_ann([1.0, 1.0, 1.0], &[valid]).await;
+    ctx.wait_for_ann([1.0, 1.0, 1.0], std::slice::from_ref(&try_replace_valid))
+        .await;
 
-    let valid_no_vec = Item::key(pk, None, "pk", "valid");
-    ctx.put(&valid_no_vec)
+    let invalid_items = [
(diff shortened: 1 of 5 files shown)


crates/validator/src/alternator/put_item.rs
knowack1, line 1: Should we break lines in commit messages at a reasonable length (like 120 characters)?

crates/validator/src/alternator/update_item.rs
```diff
@@ -176,55 +165,31 @@ async fn update_item_with_invalid_vector_is_not_indexed(actors: Arc<TestActors>)
...
+            AttributeValue::N("1.0".into()),
+            AttributeValue::N("2.0".into()),
+            AttributeValue::Null(true),
+        ]),
+        alternator::float_list([1.0_f32, 1.0]),
+        alternator::float_list([1.0_f32, 1.0, 1.0, 1.0]),
+    ];
+
+    for invalid_value in invalid_values {
+        alternator::assert_service_error(
```
knowack1, line 195: Consider logging `invalid_value`, otherwise it could be hard to determine which iteration failed.

crates/validator/src/alternator/update_item.rs
```diff
@@ -195,6 +200,117 @@ async fn update_item_with_invalid_vector_is_not_indexed(actors: Arc<TestActors>)
...
+    ctx.wait_for_ann([1.0, 1.0, 1.0], &[a.clone(), b.clone()])
+        .await;
+
+    update_item_key(&ctx, &a, "SET unrelated = :val")
+        .expression_attribute_values(":val", AttributeValue::S("changed".into()))
+        .send()
+        .await
+        .expect("UpdateItem on unrelated attribute should succeed");
+
+    let b_updated = Item::key(shape.pk(), shape.sk(), "pk", "b").vec(vec_attr, [1.0, 1.0, 1.0]);
```
knowack1, line 229: I was wondering why we need this action in this test, but I realized it's necessary to ensure all changes are processed. I now understand the 'barrier' wording from the excerpt. Adding a small comment here would be helpful.

crates/vector-store/src/db_index_backend.rs
```diff
@@ -67,14 +69,7 @@ impl DbIndexBackend {
...
     pub fn extract_vector(&self, value: CqlValue) -> anyhow::Result<Option<Vector>> {
-        match self {
-            Self::Cql { .. } => Vector::try_from(value).map(Some),
-            Self::Alternator { target_column } => vector::AlternatorAttrs {
-                attrs: value,
-                target_column: target_column.as_ref(),
-            }
-            .try_into(),
-        }
+        Vector::try_from(value).map(Some)
```
knowack1, line 84: This method is no longer needed in `db_index_backend` because, after this change, the code is the same for both the CQL and Alternator paths.

crates/vector-store/src/db_index_backend.rs
```diff
@@ -67,14 +69,7 @@ impl DbIndexBackend {
...
     }
 
     pub fn extract_document(&self, value: CqlValue) -> anyhow::Result<Option<String>> {
```
knowack1, line 87: I think this method is also not needed here and could have same impl for Cql and Alternator case:
```
match value {
                CqlValue::Text(s) => Ok(Some(s)),
                CqlValue::Ascii(s) => Ok(Some(s)),
                other => anyhow::bail!("expected text column, got {:?}", other),
            },
```

knowack1 (commented):
Looks good. Some comments for consideration.

crates/vector-store/src/db_index_backend.rs
m-szymon: It could be the same now because FTS is not implemented for Alternator, so I am not sure if that branch is even reachable.
But once we implement it, for Alternator we will need to decode text from blob.
So this function will change anyway - it will be something similar to Vector::try_from.
I don't want to change it now.

knowack1: Ok, I see the difference. In case of Alternator the value here is always a blob.

nyh:
I'm had a huge feeling of deja-vu here. I felt like **you** warned me in the past about this problem, that I wrote tests for it, and fixed them!

And indeed, it is true, I did test and fix these problems more than a month ago :-(

See commit https://github.com/scylladb/vector-store/pull/444/commits/7c444c2445f4ecc3df34aa0ee13e3c28d62f7e56 in https://github.com/scylladb/vector-store/pull/444/commits

And in addition to the tests in your Rust framework, I also added tests for it in Alternator's own tests: https://github.com/scylladb/scylladb/pull/29776/commits/236d81f6b463d015432d5700194529f26df96082 part of unmerged https://github.com/scylladb/scylladb/pull/29776. 

At least, I think it's the same thing? Or maybe it was some other related but not identical problem?

m-szymon:
> I'm had a huge feeling of deja-vu here. I felt like **you** warned me in the past about this problem, that I wrote tests for it, and fixed them!
> 
> And indeed, it is true, I did test and fix these problems more than a month ago :-(
> 
> See commit [7c444c2](https://github.com/scylladb/vector-store/commit/7c444c2445f4ecc3df34aa0ee13e3c28d62f7e56) in https://github.com/scylladb/vector-store/pull/444/commits
> 
> And in addition to the tests in your Rust framework, I also added tests for it in Alternator's own tests: [scylladb/scylladb@236d81f](https://github.com/scylladb/scylladb/commit/236d81f6b463d015432d5700194529f26df96082) part of unmerged [scylladb/scylladb#29776](https://github.com/scylladb/scylladb/pull/29776).
> 
> At least, I think it's the same thing? Or maybe it was some other related but not identical problem?

Yes, the commit message indicates the same issue and it says it is pre-existing bug. But because it is bundled with filtering it is not easy to notice it is relevant not only for filtering.

nyh (approved):

crates/vector-store/src/db_cdc.rs
```diff
@@ -430,9 +431,9 @@ fn extract_indexed_value(
     kind: &IndexKind,
 ) -> anyhow::Result<Option<DbIndexedValue>> {
     match kind {
-        IndexKind::Vs(_) => backend
-            .extract_vector(value)
-            .map(|opt| opt.map(DbIndexedValue::Vector)),
+        IndexKind::Vs(_) => {
+            Vector::try_from(value).map(|vector| Some(DbIndexedValue::Vector(vector)))
```
knowack1: Re-reviewing this change, I think it's worth preserving extract_vector. Instead of accepting a blob for non-Alternator indexes, it would be better to dispatch based on the index kind and enforce the expected type. That said, this is just a minor.

knowack1 (dismissed):

crates/vector-store/src/db_cdc.rs
m-szymon: I restored `extract_vector` as a simple wrapper to Vector::try_from.
I am not going to refactor Vector::try_from in this patch.

knowack1 (dismissed):

m-szymon:
Conflict fixed.

m-szymon:
This PR was already queued, but had conflicts with "non-empty collections".
Now I rebased it again on master.
@ewienik @knowack1

ewienik (approved):

PR #504, opened 2026-07-06
vector-store: minimal DiskANN integration

Akvear:
This PR are the first steps towards integrating DiskANN. 
There is 1 new imported crate: `diskann`.
Implemented end-to-end getter for version of diskann reported via GET.

The USearch and OpenSearch code paths behave unchanged.

Fixes: VECTOR-713
Fixes: VECTOR-716
Fixes: VECTOR-718
Fixes: VECTOR-712

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,7 @@ console-subscriber = "0.5.0"
 criterion = { version = "0.8.2", features = ["async_tokio"] }
 dashmap = "6.1.0"
 derive_more = { version = "2.0.1", features = ["full"] }
+diskann = { git = "https://github.com/microsoft/DiskANN.git", rev = "541472a8d0bbfa4d458ec1b909e1d41a18adaabc" }
 dotenvy = "0.15.7"
 e2etest = "0.2.0"
 e2etest-dns = "0.1.0"
--- a/README.md
+++ b/README.md
@@ -34,6 +34,8 @@ service supports also `.env` files.
 | `VECTOR_STORE_MONITOR_INDEXES_INTERVAL`    | How often to poll Scylla for schema changes (new/removed vector indexes). The value is in human readable format (ie. `100ms`)                                                        | `1s`                     |
 | `VECTOR_STORE_INDEX_STATUS_UPDATE_INTERVAL` | How often to sync index status (e.g., BOOTSTRAPPING->SERVING) into the engine's cached state. The value is in human readable format (ie. `100ms`) | `1s`            |
 | `VECTOR_STORE_USEARCH_SIMULATOR`           | Enable simulator for USearch. Provides human readable delays for simulated operations (`search:add-remove:reserve`).                                                                 |                          |
+| `VECTOR_STORE_USE_DISKANN`                 | Use DiskANN as the indexing engine instead of USearch.                                                | `false`                  |
+| `VECTOR_STORE_DISKANN_ALPHA`               | DiskANN parameter that controls the trade-off between index quality and build time. | (DiskANN default)                    |
 | `VECTOR_STORE_ALTER_INDEX_SIMULATOR`       | Enable simulator for missing `ALTER INDEX`. When enable indexes aren't deleted and their version is not checked.                                                                     | `false`                  |
 
 ## Development builds
--- a/crates/vector-store/Cargo.toml
+++ b/crates/vector-store/Cargo.toml
@@ -40,6 +40,7 @@ const-hex.workspace = true
 console-subscriber = {workspace = true, optional = true}
 dashmap.workspace = true
 derive_more.workspace = true
+diskann.workspace = true
 dotenvy.workspace = true
 futures.workspace = true
 hotpath.workspace = true
--- a/crates/vector-store/src/config_manager.rs
+++ b/crates/vector-store/src/config_manager.rs
@@ -5,6 +5,7 @@
 
 use crate::Config;
 use crate::Credentials;
+use crate::DiskannAlpha;
 use crate::file_monitor::TlsFilesMonitor;
 use crate::tls;
 use crate::tls::TlsServerConfig;
@@ -435,6 +436,23 @@ pub async fn load_config(env: impl Fn(&str) -> anyhow::Result<String>) -> anyhow
         .map(|v| v.split(':').map(|s| s.parse::<humantime::Duration>()).map_ok(|v| v.into()).collect::<Result<Vec<_>, _>>().map_err(|err| {
             anyhow!("Unable to parse VECTOR_STORE_USEARCH_SIMULATOR env (search_us:add_us:delete_us:...): {err}")
         })).transpose()?;
+
+    if let Ok(diskann_alpha) = env("VECTOR_STORE_DISKANN_ALPHA") {
+        let alpha = diskann_alpha
+            .trim()
+            .parse::<f32>()
+            .map_err(|_| anyhow!("Unable to parse VECTOR_STORE_DISKANN_ALPHA env (float)"))?;
+        config.diskann_alpha = Some(
+            DiskannAlpha::new(alpha)
+                .map_err(|e| anyhow!("Invalid VECTOR_STORE_DISKANN_ALPHA: {e}"))?,
+        );
+    }
+
+    config.use_diskann = env("VECTOR_STORE_USE_DISKANN")
+        .unwrap_or("false".into())
+        .trim()
+        .parse()
+        .map_err(|_| anyhow!("Unable to parse VECTOR_STORE_USE_DISKANN env (bool)"))?;
 
     config.alter_index_simulator = env("VECTOR_STORE_ALTER_INDEX_SIMULATOR")
         .unwrap_or("false".into())
@@ -878,6 +896,22 @@ mod tests {
                 .to_string()
                 .contains("Unable to parse VECTOR_STORE_FULLTEXT_INDEXES")
         );
+    }
+
+    #[tokio::test]
+    async fn load_config_diskann() {
+        let env = mock_env(HashMap::new());
+        let config = load_config(env).await.unwrap();
+        assert!(config.diskann_alpha.is_none());
+        assert!(!config.use_diskann);
+
+        let env = mock_env(HashMap::from([
+            ("VECTOR_STORE_DISKANN_ALPHA", "1.2".into()),
+            ("VECTOR_STORE_USE_DISKANN", "true".into()),
+        ]));
+        let config = load_config(env).await.unwrap();
+        assert_eq!(config.diskann_alpha, Some(DiskannAlpha::new(1.2).unwrap()));
+        assert!(config.use_diskann);
     }
 
     #[test]
--- a/crates/vector-store/src/lib.rs
+++ b/crates/vector-store/src/lib.rs
@@ -157,6 +157,25 @@ impl std::fmt::Display for TableIdentifier {
     }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, PartialOrd)]
+pub struct DiskannAlpha(f32);
+
+impl DiskannAlpha {
+    pub fn new(value: f32) -> Result<Self, String> {
+        if !value.is_finite() {
+            Err(format!("DiskannAlpha must be finite, got {value}"))
+        } else if value <= 0.0 {
+            Err(format!("DiskannAlpha must be > 0, got {value}"))
+        } else {
+            Ok(Self(value))
+        }
+    }
+
+    pub fn get(self) -> f32 {
+        self.0
+    }
+}
+
 #[derive(Clone, Debug)]
 pub struct Config {
     pub vector_store_addr: std::net::SocketAddr,
@@ -167,6 +186,8 @@ pub struct Config {
     pub opensearch_addr: Option<String>,
     pub credentials: Option<Credentials>,
     pub usearch_simulator: Option<Vec<Duration>>,
+    pub diskann_alpha: Option<DiskannAlpha>,
+    pub use_diskann: bool,
     pub alter_index_simulator: bool,
     pub fulltext_indexes: bool,
     pub cql_connection_timeout: Option<Duration>,
@@ -198,6 +219,8 @@ impl Default for Config {
             opensearch_addr: None,
             credentials: None,
             usearch_simulator: None,
+            diskann_alpha: None,
+            use_diskann: false,
             alter_index_simulator: false,
             fulltext_indexes: true,
             disable_colors: false,
@@ -765,6 +788,12 @@ pub fn new_index_factory_opensearch(
     Ok(Box::new(vs_index::opensearch::new_opensearch(
         &addr, config_rx,
     )?))
+}
+
+pub fn new_index_factory_diskann(
+    config_rx: watch::Receiver<Arc<Config>>,
+) -> anyhow::Result<Box<dyn VsIndexFactory + Send + Sync>> {
+    Ok(Box::new(vs_index::diskann::new_diskann(config_rx)?))
 }
 
 pub fn openapi() -> OpenApi {
--- a/crates/vector-store/src/vs_index/diskann.rs
+++ b/crates/vector-store/src/vs_index/diskann.rs
@@ -1,0 +1,78 @@
+/*
+ * Copyright 2026-present ScyllaDB
+ * SPDX-License-Identifier: LicenseRef-ScyllaDB-Source-Available-1.1
+ */
+
+use crate::Config;
+use crate::VsIndexFactory;
+use crate::memory::Memory;
+use crate::perf;
+use crate::table::Table;
+use crate::vs_index::actor::VsIndex;
+use crate::vs_index::factory::VsIndexConfiguration;
+use std::sync::Arc;
+use std::sync::RwLock;
+use tokio::sync::mpsc;
+use tokio::sync::watch;
+use tracing::Instrument;
+use tracing::debug;
+use tracing::debug_span;
+use tracing::warn;
+
+pub struct DiskannIndexFactory;
+
+impl VsIndexFactory for DiskannIndexFactory {
+    fn create_index(
+        &self,
+        index: VsIndexConfiguration,
+        _table: Arc<RwLock<Table>>,
+        _memory: mpsc::Sender<Memory>,
+    ) -> anyhow::Result<mpsc::Sender<VsIndex>> {
+        new(index.key)
+    }
+
+    fn index_engine_version(&self) -> String {
+        format!("diskann-{}", diskann::version())
+    }
+}
+
+pub fn new_diskann(
+    _config_rx: watch::Receiver<Arc<Config>>,
+) -> anyhow::Result<DiskannIndexFactory> {
+    Ok(DiskannIndexFactory)
+}
+
+fn new(index_key: crate::IndexKey) -> anyhow::Result<mpsc::Sender<VsIndex>> {
+    let (tx, mut rx) = mpsc::channel(perf::channel_size().into());
+
+    tokio::spawn(perf::hotpath_async(
+        {
+            async move {
+                debug!("starting");
+
+                while let Some(msg) = rx.recv().await {
+                    match msg {
+                        VsIndex::AddVector { .. }
+                        | VsIndex::RemoveVector { .. }
+                        | VsIndex::RemovePartition { .. } => {
+                            warn!("not implemented yet");
+                        }
+                        VsIndex::Ann { tx, .. } | VsIndex::FilteredAnn { tx, .. } => {
+                            _ = tx
+                                .send(Err(anyhow::anyhow!("DiskANN index is not implemented yet")));
+                        }
+                        VsIndex::Count { tx, .. } => {
+                            _ = tx
+                                .send(Err(anyhow::anyhow!("DiskANN index is not implemented yet")));
+                        }
+                    }
+                }
+
+                debug!("finished");
+            }
+        }
+        .instrument(debug_span!("diskann", "{index_key}")),
+    ));
+
+    Ok(tx)
+}
--- a/crates/vector-store/tests/integration/info.rs
+++ b/crates/vector-store/tests/integration/info.rs
@@ -55,3 +55,20 @@ async fn get_application_info_opensearch() {
     assert_eq!(info.service, env!("CARGO_PKG_NAME"));
     assert_eq!(info.engine, "opensearch");
 }
+
+#[tokio::test]
+async fn get_application_info_diskann() {
+    let diskann_config = vector_store::Config {
+        ..Default::default()
+    };
+
+    let (_, config_rx) = watch::channel(Arc::new(diskann_config));
+    let (client, _server, _config_senders) =
+        run_vs(vector_store::new_index_factory_diskann(config_rx).unwrap()).await;
+
+    let info = client.info().await;
+
+    assert_eq!(info.version, env!("CARGO_PKG_VERSION"));
+    assert_eq!(info.service, env!("CARGO_PKG_NAME"));
+    assert_eq!(info.engine, format!("diskann-{}", diskann::version()));
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -993,6 +993,20 @@ name = "bytemuck"
 version = "1.24.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1fbdf580320f38b612e485521afda1ee26d10cc9884efaaa750d383e13e3c5f4"
+dependencies = [
+ "bytemuck_derive",
+]
+
+[[package]]
+name = "bytemuck_derive"
+version = "1.10.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f9abbd1bc6865053c427f7198e6af43bfdedc55ab791faed4fbd361d789575ff"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.117",
+]
 
 [[package]]
 name = "byteorder"
@@ -1759,6 +1773,60 @@ dependencies = [
 ]
 
 [[package]]
+name = "diskann"
+version = "0.55.0"
+source = "git+https://github.com/microsoft/DiskANN.git?rev=541472a8d0bbfa4d458ec1b909e1d41a18adaabc#541472a8d0bbfa4d458ec1b909e1d41a18adaabc"
+dependencies = [
+ "anyhow",
+ "bytemuck",
+ "diskann-utils",
+ "diskann-vector",
+ "diskann-wide",
+ "futures-util",
+ "half",
+ "hashbrown 0.16.0",
+ "num-traits",
+ "rand 0.9.4",
+ "thiserror 2.0.18",
+ "tokio",
+ "tracing",
+]
+
+[[package]]
+name = "diskann-utils"
+version = "0.55.0"
+source = "git+https://github.com/microsoft/DiskANN.git?rev=541472a8d0bbfa4d458ec1b909e1d41a18adaabc#541472a8d0bbfa4d458ec1b909e1d41a18adaabc"
+dependencies = [
+ "bytemuck",
+ "cfg-if",
+ "diskann-vector",
+ "diskann-wide",
+ "half",
+ "rand 0.9.4",
+ "rand_distr",
+ "thiserror 2.0.18",
+]
+
+[[package]]
+name = "diskann-vector"
+version = "0.55.0"
+source = "git+https://github.com/microsoft/DiskANN.git?rev=541472a8d0bbfa4d458ec1b909e1d41a18adaabc#541472a8d0bbfa4d458ec1b909e1d41a18adaabc"
+dependencies = [
+ "cfg-if",
+ "diskann-wide",
+ "half",
+]
+
+[[package]]
+name = "diskann-wide"
+version = "0.55.0"
+source = "git+https://github.com/microsoft/DiskANN.git?rev=541472a8d0bbfa4d458ec1b909e1d41a18adaabc#541472a8d0bbfa4d458ec1b909e1d41a18adaabc"
+dependencies = [
+ "cfg-if",
+ "half",
+]
+
+[[package]]
 name = "displaydoc"
 version = "0.2.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2303,9 +2371,12 @@ version = "2.7.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6ea2d84b969582b4b1864a92dc5d27cd2b77b622a8d79306834f1be5ba20d84b"
 dependencies = [
+ "bytemuck",
  "cfg-if",
  "crunchy",
  "num-traits",
+ "rand 0.9.4",
+ "rand_distr",
  "zerocopy",
 ]
 
@@ -4187,6 +4258,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "63b8176103e19a2643978565ca18b50549f6101881c443590420e4dc998a3c69"
 
 [[package]]
+name = "rand_distr"
+version = "0.5.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6a8615d50dcf34fa31f7ab52692afec947c4dd0ab803cc87cb3b0b4570ff7463"
+dependencies = [
+ "num-traits",
+ "rand 0.9.4",
+]
+
+[[package]]
 name = "rand_pcg"
 version = "0.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -6001,6 +6082,7 @@ dependencies = [
  "criterion",
  "dashmap 6.1.0",
  "derive_more",
+ "diskann",
  "dotenvy",
  "futures",
  "hotpath",
--- a/crates/vector-store/src/main.rs
+++ b/crates/vector-store/src/main.rs
@@ -77,10 +77,14 @@ fn main() -> anyhow::Result<()> {
 
         let config_rx = config_receivers.config.clone();
         let opensearch_addr = config_rx.borrow().opensearch_addr.clone();
+        let use_diskann = config_rx.borrow().use_diskann;
 
         let index_factory = if let Some(addr) = opensearch_addr {
             tracing::info!("Using OpenSearch index factory at {addr}");
             vector_store::new_index_factory_opensearch(addr, config_rx.clone())?
+        } else if use_diskann {
+            tracing::info!("Using DiskANN index factory");
+            vector_store::new_index_factory_diskann(config_rx.clone())?
         } else {
             tracing::info!("Using Usearch index factory");
             vector_store::new_index_factory_usearch(config_rx.clone())?
--- a/crates/vector-store/src/vs_index/mod.rs
+++ b/crates/vector-store/src/vs_index/mod.rs
@@ -11,5 +11,7 @@ pub(crate) use actor::VsIndex;
 pub(crate) use actor::VsIndexExt;
 pub(crate) use validator::Error;
 
+#[allow(dead_code)]
+pub(crate) mod diskann;
 pub(crate) mod opensearch;
 pub(crate) mod usearch;
--- a/crates/vector-store/src/vs_index/mod.rs
+++ b/crates/vector-store/src/vs_index/mod.rs
@@ -11,7 +11,6 @@ pub(crate) use actor::VsIndex;
 pub(crate) use actor::VsIndexExt;
 pub(crate) use validator::Error;
 
-#[allow(dead_code)]
 pub(crate) mod diskann;
 pub(crate) mod opensearch;
 pub(crate) mod usearch;


crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -0,0 +1,479 @@
...
+
+    dataset.write_all(&SEED_DATASET_POINTS.to_le_bytes())?;
+    dataset.write_all(&dimensions_u32.to_le_bytes())?;
+
+    for i in 0..SEED_DATASET_POINTS {
+        for j in 0..usize::from(dimensions.0) {
+            let dummy_val: f32 = (i as f32) + (j as f32 * 0.1);
+            dataset.write_all(&dummy_val.to_le_bytes())?;
+        }
+    }
```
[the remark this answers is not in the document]
Akvear: Not important, this need to be overhauled either way.

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -0,0 +1,479 @@
...
+    /// Creates a new provider. The directory should be provisioned before calling this.
+    pub fn new(base_dir: PathBuf) -> Self {
+        Self { base_dir }
+    }
+
+    /// Helper to safely resolve the DiskANN string identifier (e.g., "index.graph")
+    /// to a physical path within the SSD directory.
+    fn get_path(&self, item_identifier: &str) -> PathBuf {
+        self.base_dir.join(item_identifier)
+    }
```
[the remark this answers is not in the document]
Akvear: I don't think this is necessary.

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -0,0 +1,741 @@
...
+                    embedding.len()
+                );
+                error!("{err}");
+                return Some(DiskannState::Fail(err));
+            }
+
+            collector.vectors.push((primary_id, embedding));
+            if collector.vectors.len() == BUILD_DATASET_POINTS {
+                return Some(transition_to_serving(collector).await);
+            }
```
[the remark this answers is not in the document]
Akvear: Mutating requests should be blocking, for now it's easier to make queries also blocking.

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -0,0 +1,83 @@
+/*
+ * Copyright 2025-present ScyllaDB
```
QuerthDP: nit: 2026

crates/vector-store/Cargo.toml
QuerthDP, line 1: consider specifying the scope of this commit:

vector-store: add diskann crates ?
treewide: add discann crates ?
chore: add discann crates ?

I'd prefer the 3rd.

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -81,3 +88,97 @@ fn new(index_key: crate::IndexKey) -> anyhow::Result<mpsc::Sender<VsIndex>> {
...
+
+impl TryFrom<(&VsIndexConfiguration, f32, usize)> for DiskannParams {
+    type Error = anyhow::Error;
+
+    fn try_from(
+        (cfg, alpha, max_points): (&VsIndexConfiguration, f32, usize),
+    ) -> Result<Self, Self::Error> {
+        anyhow::ensure!(
+            cfg.quantization == Quantization::F32,
+            "DiskANN engine (v1) only supports F32 quantization; got {:?}",
```
QuerthDP: What is the (v1)? Is it some specific version of the crate?
I think we shall not specify that. If we will support it, we won't need this message.
Maybe it would be not supported in (v2) as well?

crates/vector-store/src/vs_index/diskann.rs
QuerthDP, line 1: Let's try to add some explanation/motivation/summary in the commit messages. It helps to understand deeper what the patch tries to improve/change.

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -81,3 +88,97 @@ fn new(index_key: crate::IndexKey) -> anyhow::Result<mpsc::Sender<VsIndex>> {
...
+                nbits: 8,
+                standard_deviation: None,
+            }),
+            Quantization::F16 | Quantization::BF16 => Ok(QuantizationType::PQ {
+                num_chunks: BUILD_PQ_CHUNKS,
+            }),
+            Quantization::B1 => anyhow::bail!("DiskANN does not support B1 quantization"),
+        }
+    }
+}
```
QuerthDP: I don't understant. Once you mention that we don't support Quantization, and later you add those quantization support conversions.

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -182,3 +182,90 @@ impl TryFrom<Quantization> for QuantizationType {
...
+        assert_eq!(
+            params.config.pruned_degree(),
+            NonZeroUsize::new(16).unwrap()
+        );
+        assert_eq!(params.dim, 3);
+        assert_eq!(params.l_search_default, NonZeroUsize::new(32).unwrap());
+        assert_eq!(params.config.l_build(), NonZeroUsize::new(64).unwrap());
+        assert_eq!(params.metric, Metric::L2);
+    }
+}
```
QuerthDP: This should be included in previous commit. Also think if we need them now, if we don't support Quantization.

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -4,43 +4,70 @@
...
+use diskann::graph::config::defaults::ALPHA as DISKANN_DEFAULT_ALPHA;
+use diskann::graph::config::{Builder, MaxDegree};
 use diskann::utils::ONE;
+use diskann_disk::build::builder::build::DiskIndexBuilder;
+use diskann_disk::data_model::AdHoc;
+use diskann_disk::disk_index_build_parameter::{DISK_SECTOR_LEN, MemoryBudget, NumPQChunks};
+use diskann_disk::storage::DiskIndexWriter;
+use diskann_disk::{DiskIndexBuildParameters, QuantizationType};
 use diskann_providers::model::configuration::IndexConfiguration;
+use diskann_providers::storage::{StorageReadProvider, StorageWriteProvider};
```
QuerthDP: Let's keep the one-line-per-import conversion

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -49,35 +76,61 @@ impl VsIndexFactory for DiskannIndexFactory {
...
-    _config_rx: watch::Receiver<Arc<Config>>,
+    mut config_rx: watch::Receiver<Arc<Config>>,
 ) -> anyhow::Result<DiskannIndexFactory> {
-    Ok(DiskannIndexFactory)
+    let config = config_rx.borrow_and_update();
+
+    let diskann_index_path = config
+        .diskann_index_path
+        .clone()
+        .ok_or_else(|| anyhow::anyhow!("DiskANN index path should be set"))?;
```
QuerthDP: ok_or()?

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -89,12 +142,263 @@ fn new(index_key: crate::IndexKey) -> anyhow::Result<mpsc::Sender<VsIndex>> {
...
+            let build_result = build_disk_index(params, dimensions, vectors, &index_dir);
+            match build_result {
+                Ok(()) => {
+                    debug!("DiskANN index built from collected vectors");
+                    DiskannState::Serving
+                }
+                Err(err) => {
+                    let err = err.to_string();
+                    error!("DiskANN index build failed: {err}");
+                    if let Err(e) = std::fs::remove_dir_all(&index_dir) {
```
QuerthDP: Should we remove all data or retry?

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -89,12 +142,263 @@ fn new(index_key: crate::IndexKey) -> anyhow::Result<mpsc::Sender<VsIndex>> {
...
+        MemoryBudget::try_from_gb(BUILD_MEMORY_LIMIT_GB)
+            .context("failed to create DiskANN build memory budget")?,
+        params.quantization,
+        NumPQChunks::new_with(BUILD_PQ_CHUNKS, usize::from(dimensions.0))
+            .context("failed to create DiskANN PQ chunk configuration")?,
+    );
+
+    let index_configuration = IndexConfiguration::from(params);
+
+    let dataset_path = index_dir.join("dataset.bin");
```
QuerthDP: Why .bin?

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -103,13 +407,8 @@ impl TryFrom<(&VsIndexConfiguration, f32, usize)> for DiskannParams {
...
         (cfg, alpha, max_points): (&VsIndexConfiguration, f32, usize),
     ) -> Result<Self, Self::Error> {
-        anyhow::ensure!(
-            cfg.quantization == Quantization::F32,
-            "DiskANN engine (v1) only supports F32 quantization; got {:?}",
-            cfg.quantization
-        );
-
         let metric: Metric = cfg.space_type.try_into()?;
+        let quantization_type = QuantizationType::try_from(cfg.quantization)?;
```
QuerthDP: So do we plan to support quantization now? Why did we need the error before at all

crates/vector-store/src/vs_index/diskann.rs
QuerthDP, line 1: > vector-store: create DiskANN index and materialize index files

This commit feels huge, with no proper explanaition.
Please consider if this could be split further and add proper commit message.

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -552,7 +553,17 @@ mod tests {
...
+    use crate::vs_index::VsIndexExt;
+    use diskann_providers::storage::{
+        get_compressed_pq_file, get_disk_index_file, get_pq_pivot_file,
+    };
     use std::num::NonZeroUsize;
+    use std::path::Path;
+    use tempfile::tempdir;
+    use tokio::sync::watch;
+    use tokio::task;
+    use tokio::time::{self, Duration};
```
QuerthDP: nit: one-import-per-line

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -627,4 +642,171 @@ mod tests {
...
+            tmp_dir.path(),
+            table,
+            memory_tx,
+        )
+        .unwrap();
+
+        let err = actor.count(index_key).await.unwrap_err();
+        assert_eq!(
+            err.to_string(),
+            "DiskANN index still collecting vectors (0/1000)"
```
QuerthDP: why 0/1000?

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -627,4 +642,171 @@ mod tests {
...
+                PrimaryId::from(1),
+                vec![1.0, 2.0].into(),
+                crate::AsyncInProgress::None,
+            )
+            .await;
+
+        let err = actor.count(index_key).await.unwrap_err();
+        assert_eq!(
+            err.to_string(),
+            "DiskANN collector vector dimensions mismatch: expected 3, got 2"
```
QuerthDP: Where is the expected dimension 3? I cannot find it

crates/vector-store/src/config_manager.rs
```diff
@@ -604,7 +604,7 @@ mod tests {
     use secrecy::ExposeSecret;
     use std::collections::HashMap;
     use std::io::Write;
-    use tempfile::NamedTempFile;
+    use tempfile::{NamedTempFile, tempdir};
```
QuerthDP: nit: one-import-per-line

crates/vector-store/tests/integration/info.rs
```diff
@@ -55,3 +56,23 @@ async fn get_application_info_opensearch() {
     assert_eq!(info.service, env!("CARGO_PKG_NAME"));
     assert_eq!(info.engine, "opensearch");
 }
+
+#[tokio::test]
+async fn get_application_info_diskann() {
+    // DiskANN is set up to fail if the index path is not set.
```
QuerthDP: Why do we need this comment here?

QuerthDP (commented):
At the first glance it looks really good.
Please consider my suggestions of how to improve the readability/maintainability of this code IMO.
I'm not an expert in DiskANN (not even a rookie), but from the existing indexing perspective the wiring looks fine.

crates/vector-store/src/vs_index/diskann.rs
Akvear: It's deleted either way later on, but this whole section is a bit misleading when going commit by commit, will fix.

crates/vector-store/src/vs_index/diskann.rs
Akvear: retry will likely results on the same error, data is removed to avoid manual cleanup if build doesn't succeed

crates/vector-store/src/vs_index/diskann.rs
Akvear: It doesn't really matter, but it's raw data so i think .bin is suitable.

crates/vector-store/src/vs_index/diskann.rs
Akvear: The commit history is a bit messed up with this.

crates/vector-store/src/vs_index/diskann.rs
Akvear: Since no `VsIndex::AddVector` was present during the collecting phase, `actor.count()` will return an error stating that vectors are still being collected and 0 have been collected so far. The 1000 should be changed to the BUILD_DATASET_POINTS constant.

crates/vector-store/src/vs_index/diskann.rs
Akvear: The `test_index()` methods returns a `VsIndexConfiguration` with dimension 3.

crates/vector-store/tests/integration/info.rs
Akvear: not needed.

crates/vector-store/src/vs_index/diskann.rs
Akvear: I split it up into 3 - should be more understandable now.

QuerthDP:
Please provide a changelog and rerequest review once you finish responding to the round of commits.

Cargo.toml
```diff
@@ -36,6 +36,10 @@ console-subscriber = "0.5.0"
 criterion = { version = "0.8.2", features = ["async_tokio"] }
 dashmap = "6.1.0"
 derive_more = { version = "2.0.1", features = ["full"] }
+diskann = { git = "https://github.com/microsoft/DiskANN.git", rev = "541472a8d0bbfa4d458ec1b909e1d41a18adaabc" }
```
knowack1, line 39: Explain in the commit message why using version from git sha.

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -0,0 +1,83 @@
+/*
+ * Copyright 2026-present ScyllaDB
+ * SPDX-License-Identifier: LicenseRef-ScyllaDB-Source-Available-1.0
```
knowack1: A new licence is LicenseRef-ScyllaDB-Source-Available-1.**1**. Also in other places.

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -0,0 +1,83 @@
...
+const BUILD_MEMORY_LIMIT_GB: f64 = 2.0;
+const BUILD_PQ_CHUNKS: usize = 1;
+
+pub struct DiskannIndexFactory;
+
+impl VsIndexFactory for DiskannIndexFactory {
+    fn create_index(
+        &self,
+        index: VsIndexConfiguration,
+        _table: Arc<RwLock<Table>>,
```
knowack1, line 28: Remove unused parameters. Add them only when needed. I see that table and memory are not used even after the last commit in this PR.

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -0,0 +1,83 @@
...
+use tokio::sync::watch;
+use tracing::Instrument;
+use tracing::debug;
+use tracing::debug_span;
+use tracing::warn;
+
+const NUM_THREADS: usize = 1;
+const MAX_POINTS: usize = 1_000_000;
+const BUILD_MEMORY_LIMIT_GB: f64 = 2.0;
+const BUILD_PQ_CHUNKS: usize = 1;
```
knowack1: nit: these constants are not used in this commit.

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -81,3 +88,184 @@ fn new(index_key: crate::IndexKey) -> anyhow::Result<mpsc::Sender<VsIndex>> {
...
+
+#[derive(Clone)]
+pub(crate) struct DiskannParams {
+    pub(crate) config: DiskannConfig,
+    pub(crate) metric: Metric,
+    pub(crate) dim: usize,
+    pub(crate) max_points: usize,
+    #[allow(dead_code)]
+    // AFAIK l_search is used per query, but we don't have queries yet, so this is unused for now
+    pub(crate) l_search_default: NonZeroUsize,
```
knowack1: 1. The code doesn't compile.
2. Commit includes code that is not needed at this stage.

Each commit in a PR needs to be atomic, self-contained and  minimal code necessary to satisfy the desired scope/task.

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -81,3 +88,184 @@ fn new(index_key: crate::IndexKey) -> anyhow::Result<mpsc::Sender<VsIndex>> {
 
     Ok(tx)
 }
+
+#[derive(Clone)]
+pub(crate) struct DiskannParams {
```
knowack1: It is pub(crate). Is this ever referenced from outside the module?

crates/vector-store/src/vs_index/diskann.rs
knowack1, line 1: Any lines following the commit title should be free-form text providing additional details about the changes, if necessary. Here is has same form as commit title: "vector-store: add test for mappings"

crates/vector-store/src/vs_index/diskann.rs
QuerthDP: Should we update it treewide?

crates/vector-store/src/vs_index/diskann.rs
Akvear: I think there was a reason for this but it was not affecting it anymore, it's now module only.

Akvear:
Changes:
- split up a big commit into 3 logically separate commits
- deleted a commit with unnecessary StorageProvider implementation, instead used a DiskANN FileStorageProvider
- squashed mapping tests into the previous commit
- fixed commit history to not mislead about the quantization
- kept one line per import style
- put new imports and new constants in appropriate commits
- changed commit order to make all of them compile

crates/vector-store/src/vs_index/diskann.rs
Akvear: Fixed, there are now only 2 places where clippy complains about unused, but the commit history is clearer this way.

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -0,0 +1,763 @@
...
+    _table: Arc<RwLock<impl TableSearch + Send + Sync + 'static>>,
+    _memory: mpsc::Sender<Memory>,
+) -> anyhow::Result<mpsc::Sender<VsIndex>> {
+    let index_dir = diskann_index_path.join(index_key.as_ref());
+
+    if index_dir.exists() && index_dir.read_dir()?.next().is_some() {
+        anyhow::bail!("DiskANN index directory is non-empty (refusing to delete): {index_dir:?}");
+    }
+
+    std::fs::create_dir_all(&index_dir).context("failed to create DiskANN index directory")?;
```
knowack1: For I/O, I think we should use the async Tokio counterparts. Instead of blocking functions.

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -0,0 +1,763 @@
...
+    mut config_rx: watch::Receiver<Arc<Config>>,
+) -> anyhow::Result<DiskannIndexFactory> {
+    let config = config_rx.borrow_and_update();
+
+    let diskann_index_path = config
+        .diskann_index_path
+        .clone()
+        .ok_or(anyhow::anyhow!("DiskANN index path should be set"))?;
+
+    if !diskann_index_path.exists() {
```
knowack1: There are already similar check in config_manager. Do we need to check this twice?

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -0,0 +1,763 @@
...
+            SpaceType::Cosine => Ok(Self::Cosine),
+            SpaceType::DotProduct => Ok(Self::InnerProduct),
+            SpaceType::Hamming => {
+                anyhow::bail!("DiskANN does not support Hamming space type")
+            }
+        }
+    }
+}
+
+impl TryFrom<Quantization> for QuantizationType {
```
knowack1: According to https://github.com/swasik/vector-store/blob/e5199b8cbcaaea945488bb36e989e10a25e06a58/docs/dev/diskann/sep-diskann.md, this PQ should be independent on quantization:
```
PQ is configured internally to the engine with sensible defaults in this increment — DiskANN's PQ is independent of the existing Quantization enum (f32/f16/bf16/i8/b1) and is not surfaced in CQL yet; exposing PQ parameters through CQL is a follow-up iteration.
```

Also I couldn't find this mapping requirements in Jira tickets.

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -0,0 +1,763 @@
...
+    fn try_from(quantization: Quantization) -> Result<Self, Self::Error> {
+        match quantization {
+            Quantization::F32 => Ok(QuantizationType::PQ {
+                num_chunks: BUILD_PQ_CHUNKS,
+            }),
+            Quantization::I8 => Ok(QuantizationType::SQ {
+                nbits: 8,
+                standard_deviation: None,
+            }),
+            Quantization::F16 | Quantization::BF16 => Ok(QuantizationType::PQ {
```
knowack1: Same as Quantization::F32 arm

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -0,0 +1,763 @@
...
+    Ok(DiskannIndexFactory {
+        diskann_index_path,
+        alpha: config.diskann_alpha.unwrap_or(DISKANN_DEFAULT_ALPHA),
+    })
+}
+
+fn new(
+    params: DiskannParams,
+    index_key: IndexKey,
+    dimensions: Dimensions,
```
knowack1: `dimensions` are also in `params`.

README.md
```diff
@@ -34,6 +34,8 @@ service supports also `.env` files.
 | `VECTOR_STORE_MONITOR_INDEXES_INTERVAL`    | How often to poll Scylla for schema changes (new/removed vector indexes). The value is in human readable format (ie. `100ms`)                                                        | `1s`                     |
 | `VECTOR_STORE_INDEX_STATUS_UPDATE_INTERVAL` | How often to sync index status (e.g., BOOTSTRAPPING->SERVING) into the engine's cached state. The value is in human readable format (ie. `100ms`) | `1s`            |
 | `VECTOR_STORE_USEARCH_SIMULATOR`           | Enable simulator for USearch. Provides human readable delays for simulated operations (`search:add-remove:reserve`).                                                                 |                          |
+| `VECTOR_STORE_DISKANN_INDEX_PATH`          | Path to a directory where DiskANN indexes are stored. If set, DiskANN will be used as the indexing backend instead of USearch. The path must exist and be a directory.               |                          |
```
knowack1: VECTOR_STORE_DISKANN_INDEX_PATH reads like "the path to one index," but it's actually a root/base directory under which the engine creates one subdirectory per index.

AI suggestions:
```
Name	Why less good
VECTOR_STORE_DISKANN_ROOT_PATH	Accurate ("root of a tree of index dirs"), but "root" is a bit generic/ambiguous with filesystem root, or the app's install root.
VECTOR_STORE_DISKANN_DATA_DIR	Clear, but loses the "this is where SSD-backed index data specifically lives" nuance; also inconsistent suffix (_DIR vs the rest of the codebase's _PATH/_URI convention).
VECTOR_STORE_DISKANN_SSD_PATH	Closest to the SEP's literal phrase "SSD index path," but ties the name to a storage medium (SSD) rather than role — slightly awkward if someone points it at, say, a different NVMe-backed mount that isn't literally "SSD" branded, or a test tmp dir.
VECTOR_STORE_DISKANN_INDEXES_PATH (pluralized)	Minimal fix, technically correct, but doesn't really clarify to a reader that it's a base/root dir vs. "the path where all my indexes' data happens to be."
```

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -0,0 +1,763 @@
...
+        Self {
+            params,
+            dimensions,
+            index_dir,
+            vectors: Vec::with_capacity(BUILD_DATASET_POINTS),
+        }
+    }
+}
+
+enum DiskannState {
```
knowack1: I suggest to add some comment. What each state means and why we need this.

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -0,0 +1,763 @@
...
+
+fn build_disk_index(
+    params: DiskannParams,
+    dimensions: Dimensions,
+    vectors: Vec<(PrimaryId, Vector)>,
+    index_dir: &Path,
+) -> anyhow::Result<()> {
+    let storage_provider = FileStorageProvider;
+
+    // TODO: make these DiskANN build constants configurable or agree on
```
knowack1: Instead of leaving a TODO, please create a Jira ticket. With just a TODO, it's highly likely it will never be addressed.

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -0,0 +1,763 @@
...
+    for (_, vector) in vectors {
+        for value in vector.as_slice() {
+            dataset.write_all(&value.to_le_bytes())?;
+        }
+    }
+
+    let dataset = dataset
+        .into_inner()
+        .map_err(|e| anyhow::anyhow!("failed to flush DiskANN dataset buffer: {}", e))?;
+    dataset.sync_all()?;
```
knowack1: I think we do not need to sync.  In case of restart we will rebuild the index from scratch.

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -0,0 +1,763 @@
...
+        ))?
+        .to_string();
+    let index_path_prefix_str = prefix_path
+        .to_str()
+        .ok_or(anyhow::anyhow!(
+            "DiskANN index prefix path is not valid UTF-8: {prefix_path:?}"
+        ))?
+        .to_string();
+
+    write_dataset(&dataset_path, dimensions, &vectors)?;
```
knowack1: I wonder how to write/remove vectors from index when it is built. Could you please provide a snippet?

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -0,0 +1,763 @@
...
+
+    Ok(())
+}
+
+fn write_dataset(
+    dataset_path: &Path,
+    dimensions: Dimensions,
+    vectors: &[(PrimaryId, Vector)],
+) -> anyhow::Result<()> {
+    let dataset = File::create(dataset_path).map_err(|e| {
```
knowack1: I'm not sure whether we should use the blocking or async versions of the I/O functions. Using the blocking versions might make sense here for performance reasons. @ewienik , could you please advise?

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -0,0 +1,763 @@
...
+                {
+                    break;
+                }
+                task::yield_now().await;
+            }
+        })
+        .await
+        .unwrap();
+
+        drop(actor);
```
knowack1: it's redundant. This happens automatically.

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -0,0 +1,763 @@
...
+
+    #[tokio::test]
+    async fn dimension_mismatch_transitions_to_fail() {
+        let tmp_dir = tempdir().unwrap();
+        let cfg = test_index();
+        let index_key = cfg.key.clone();
+        let dimensions = cfg.dimensions;
+        let params = DiskannParams::try_from((&cfg, ALPHA, MAX_POINTS)).unwrap();
+        let table = Arc::new(RwLock::new(MockTableSearch::new()));
+        let (memory_tx, _memory_rx) = mpsc::channel(1);
```
knowack1: This boilerplate appears multiple times. Consider moving to dedicated utility function:
```
    let tmp_dir = tempdir().unwrap();
        let cfg = test_index();
        let index_key = cfg.key.clone();
        let dimensions = cfg.dimensions;
        let params = DiskannParams::try_from((&cfg, ALPHA, MAX_POINTS)).unwrap();
        let table = Arc::new(RwLock::new(MockTableSearch::new()));
        let (memory_tx, _memory_rx) = mpsc::channel(1);
```

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -0,0 +1,763 @@
...
+            }
+        })
+        .await
+        .unwrap();
+
+        drop(actor);
+    }
+
+    #[tokio::test]
+    async fn collecting_returns_error_before_threshold() {
```
knowack1: The name is a bit off because it doesn't describe what's actually being tested. The test doesn't test collecting" behavior at all - it tests that calling .count() while the index is still in the Collecting state (before it has accumulated BUILD_DATASET_POINTS vectors) returns an error.

A more accurate name would be something like:                                                                                                                                                                                                    
     - count_fails_while_collecting                                                                                                                                                                                                                                                        
     - count_returns_error_while_collecting                                                                                                                                                                                                                                                
     - count_errors_before_build_threshold_reached

We should test also other methods: search, search_filtered for this behavior.

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -0,0 +1,763 @@
...
+            NonZeroUsize::new(16).unwrap()
+        );
+        assert_eq!(params.dim, 3);
+        assert_eq!(params.l_search_default, NonZeroUsize::new(32).unwrap());
+        assert_eq!(params.config.l_build(), NonZeroUsize::new(64).unwrap());
+        assert_eq!(params.metric, Metric::L2);
+    }
+
+    #[test]
+    fn new_diskann_path_validation() {
```
knowack1: Is this test somehow duplicate of what is in config_manager?

knowack1 (commented):
Please find more comments from my side.

Given the scope, I'd suggest splitting this into two separate PRs:
1. Scaffolding for the DiskANN index instance — everything needed to create the index where all methods return "not implemented" (or similar stub). This can be verified via the info.version integration test already included in this PR.
2. Actual DiskANN index implementation — the collector, build pipeline, serving state, etc.

Akvear:
Changes:
- Changed the scope of this PR

Comments regarding the part delegated to a follow up will be addressed in the follow up.

Cargo.toml
knowack1: @Akvear I see https://github.com/microsoft/DiskANN/pull/1240. Nice!

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -4,12 +4,17 @@
...
 use crate::Config;
+use crate::Quantization;
+use crate::SpaceType;
 use crate::VsIndexFactory;
 use crate::memory::Memory;
 use crate::perf;
 use crate::table::Table;
 use crate::vs_index::actor::VsIndex;
 use crate::vs_index::factory::VsIndexConfiguration;
+
```
ewienik: nit: don't leave space between `use` blocks

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -19,6 +24,17 @@ use tracing::debug;
 use tracing::debug_span;
 use tracing::warn;
 
+use diskann::graph::Config as DiskannConfig;
```
ewienik: nit: don't leave space between `use` blocks

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -76,3 +92,186 @@ fn new(index_key: crate::IndexKey) -> anyhow::Result<mpsc::Sender<VsIndex>> {
 
     Ok(tx)
 }
+
+#[derive(Clone)]
+struct DiskannParams {
+    config: DiskannConfig,
+    metric: Metric,
+    dim: usize,
```
ewienik: Could it be Dimension type?

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -76,3 +92,186 @@ fn new(index_key: crate::IndexKey) -> anyhow::Result<mpsc::Sender<VsIndex>> {
...
+    // quantization is important for the index creation, but index creation is not implemented yet
+    quantization: QuantizationType,
+}
+
+impl TryFrom<(&VsIndexConfiguration, f32, usize)> for DiskannParams {
+    type Error = anyhow::Error;
+
+    fn try_from(
+        (cfg, alpha, max_points): (&VsIndexConfiguration, f32, usize),
+    ) -> Result<Self, Self::Error> {
```
ewienik: Could it be better to have DiskannParams::new() -> anyhow::Result<_>?

crates/vector-store/src/vs_index/diskann.rs
Akvear: I will look into how quantization in DiskANN works, it's not very well documented. It might be better to leave this for the follow up

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -76,3 +92,186 @@ fn new(index_key: crate::IndexKey) -> anyhow::Result<mpsc::Sender<VsIndex>> {
 
     Ok(tx)
 }
+
+#[derive(Clone)]
+struct DiskannParams {
+    config: DiskannConfig,
+    metric: Metric,
+    dim: usize,
+    max_points: usize,
```
ewienik: Is zero possible? If not, consider change to NonZeroUsize. Or better create a newtype.

crates/vector-store/src/vs_index/diskann.rs
Akvear: It's there for a unit test, but the test is somewhat of a duplicate so this can be safely removed.

crates/vector-store/src/lib.rs
```diff
@@ -167,6 +167,8 @@ pub struct Config {
     pub opensearch_addr: Option<String>,
     pub credentials: Option<Credentials>,
     pub usearch_simulator: Option<Vec<Duration>>,
+    pub diskann_index_path: Option<std::path::PathBuf>,
+    pub diskann_alpha: Option<f32>,
```
ewienik: Consider providing a newtype instead of f32, as this value has its own restrictions (finite positive float)

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -53,12 +62,43 @@ impl VsIndexFactory for DiskannIndexFactory {
...
 
-fn new(index_key: crate::IndexKey) -> anyhow::Result<mpsc::Sender<VsIndex>> {
+fn new(index_key: IndexKey, diskann_index_path: &Path) -> anyhow::Result<mpsc::Sender<VsIndex>> {
+    let index_dir = diskann_index_path.join(index_key.as_ref());
+
+    if index_dir.exists() && index_dir.read_dir()?.next().is_some() {
+        anyhow::bail!("DiskANN index directory is non-empty (refusing to delete): {index_dir:?}");
+    }
+
+    std::fs::create_dir_all(&index_dir).context("failed to create DiskANN index directory")?;
```
ewienik: Consider changing into asynchronous tokio fs

ewienik (changes requested):

crates/vector-store/src/vs_index/diskann.rs
knowack1: I think yes. But of course in dedicated PR.

Akvear:
Changes:
- removed quantization from the scope
- fixed newlines between imports
- changed usize and f32 to more restricitve types
- changed std::fs to tokio::fs
- updated env variable name in README

crates/vector-store/src/vs_index/diskann.rs
Akvear: It is

crates/vector-store/src/config_manager.rs
```diff
@@ -436,6 +438,39 @@ pub async fn load_config(env: impl Fn(&str) -> anyhow::Result<String>) -> anyhow
             anyhow!("Unable to parse VECTOR_STORE_USEARCH_SIMULATOR env (search_us:add_us:delete_us:...): {err}")
         })).transpose()?;
 
+    if let Ok(raw_diskann_index_path) = env("VECTOR_STORE_DISKANN_INDEX_PATH") {
```
knowack1: Please add unit tests.

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -0,0 +1,78 @@
...
+use tracing::warn;
+
+pub struct DiskannIndexFactory;
+
+impl VsIndexFactory for DiskannIndexFactory {
+    fn create_index(
+        &self,
+        index: VsIndexConfiguration,
+        _table: Arc<RwLock<Table>>,
+        _memory: mpsc::Sender<Memory>,
```
knowack1, line 29: Remove unused parameters.

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -49,19 +59,48 @@ impl VsIndexFactory for DiskannIndexFactory {
...
             async move {
                 debug!("starting");
 
+                async {
+                    if tokio::fs::try_exists(&index_dir).await.unwrap_or(false) {
+                        let mut dir = tokio::fs::read_dir(&index_dir).await?;
+                        if dir.next_entry().await?.is_some() {
+                            anyhow::bail!("DiskANN index directory is non-empty: {index_dir:?}");
+                        }
+                    }
```
knowack1: Move to dedicated function:
```
verify_is_empty()?;
```

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -49,19 +59,48 @@ impl VsIndexFactory for DiskannIndexFactory {
...
+
     let (tx, mut rx) = mpsc::channel(perf::channel_size().into());
 
     tokio::spawn(perf::hotpath_async(
         {
             async move {
                 debug!("starting");
 
+                async {
+                    if tokio::fs::try_exists(&index_dir).await.unwrap_or(false) {
```
knowack1: I think we need to wipe the existing index data at bootup - as they can be out of sync. However, this can't be done at the individual index level because we also need to wipe data for indices that no longer exist.

crates/vector-store/src/config_manager.rs
QuerthDP: Let's describe this env variables in commit message

crates/vector-store/src/config_manager.rs
```diff
@@ -436,6 +438,39 @@ pub async fn load_config(env: impl Fn(&str) -> anyhow::Result<String>) -> anyhow
...
+    if let Ok(diskann_alpha) = env("VECTOR_STORE_DISKANN_ALPHA") {
+        let alpha = diskann_alpha
+            .trim()
+            .parse::<f32>()
+            .map_err(|_| anyhow!("Unable to parse VECTOR_STORE_DISKANN_ALPHA env (float)"))?;
+        if !alpha.is_finite() || alpha <= 0.0 {
+            anyhow::bail!(
+                "VECTOR_STORE_DISKANN_ALPHA must be a finite positive float; got {alpha}"
+            );
+        }
```
QuerthDP: Do we need this check here? This is duplicated in the `PositiveFiniteF32::new`

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -19,6 +29,8 @@ use tracing::debug;
 use tracing::debug_span;
 use tracing::warn;
 
+const NUM_THREADS: usize = 1;
```
QuerthDP: Why 1?

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -76,3 +88,140 @@ fn new(index_key: crate::IndexKey) -> anyhow::Result<mpsc::Sender<VsIndex>> {
...
+
+#[derive(Clone)]
+struct DiskannParams {
+    config: DiskannConfig,
+    metric: Metric,
+    dim: Dimensions,
+    max_points: NonZeroUsize,
+    #[allow(dead_code)]
+    // AFAIK l_search is used per query, but we don't have queries yet, so this is unused for now
+    l_search_default: NonZeroUsize,
```
QuerthDP: Maybe we don't need it now and we'll add it when queries are here?

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -76,3 +88,140 @@ fn new(index_key: crate::IndexKey) -> anyhow::Result<mpsc::Sender<VsIndex>> {
...
+    fn diskann_params_try_from_index_configuration() {
+        let params = DiskannParams::try_from((
+            &test_index(),
+            PositiveFiniteF32::new(DISKANN_DEFAULT_ALPHA).unwrap(),
+            MAX_POINTS,
+        ))
+        .unwrap();
+
+        assert_eq!(
+            params.config.pruned_degree(),
```
QuerthDP: What is a pruned_degree() and how does it apply to the MAX_POINTS?

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -76,3 +88,140 @@ fn new(index_key: crate::IndexKey) -> anyhow::Result<mpsc::Sender<VsIndex>> {
...
+    use crate::Connectivity;
+    use crate::ExpansionAdd;
+    use crate::ExpansionSearch;
+    use crate::IndexKey;
+    use crate::IndexName;
+    use crate::KeyspaceName;
+    use crate::Quantization;
+    use std::num::NonZeroUsize;
+
+    const MAX_POINTS: NonZeroUsize = NonZeroUsize::new(1_000_000).unwrap();
```
QuerthDP: Do we check it anywhere in the code not the tests?

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -31,7 +36,12 @@ use tracing::warn;
 
 const NUM_THREADS: usize = 1;
 
-pub struct DiskannIndexFactory;
+pub struct DiskannIndexFactory {
+    diskann_index_path: PathBuf,
+    #[allow(dead_code)]
+    // alpha is wired into the DiskANN config, but events are not implemented yet
```
QuerthDP: What events?

README.md
QuerthDP, line 1: This commit could be introduced sooner when the env variables are introduced.

crates/vector-store/tests/integration/info.rs
QuerthDP, line 1: This could be introduced earlier as well, when the version is introduced.

crates/vector-store/src/vs_index/diskann.rs
Akvear: These are required by the `VsIndexFactory` trait.

crates/vector-store/src/vs_index/diskann.rs
Akvear: I don't know where to get the number of threads from? Config or something like usearch with `perf::num_workers().into()`

crates/vector-store/src/vs_index/diskann.rs
Akvear: Pruned degree is the `R` in VECTOR-717, it has nothing to do with MAX_POINTS. MAX_POINTS is a limit of how big the dataset the index is built around is, so it can preallocate resources - i am not sure where to get this value from.

crates/vector-store/src/vs_index/diskann.rs
ewienik: We should use async I/O functions

crates/vector-store/src/vs_index/diskann.rs
ewienik: perf::num_workers() should be used to get the number  of tokio workers, that is the number of thread for tokio runtime. But this NUM_THREADS is the parameter for diskann - does it mean diskann provides its own thread pool? Could it be configured to use our thread pool?

ewienik: WIth the first look at IndexConfiguration and DiskAnnIndex this value is a hint for the number of threads, where index will be used - it provides hint for the Queue. So it seems DiskAnnIndex doesn't store its own thread pool. This value can be replaced with perf::num_workers()

crates/vector-store/src/vs_index/diskann.rs
ewienik: I see there is still TryFrom implemented for a tuple of three types - IMHO it is a sign that there is a need for a ctor method instead of TryFrom.

crates/vector-store/src/lib.rs
```diff
@@ -157,6 +157,23 @@ impl std::fmt::Display for TableIdentifier {
     }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, PartialOrd)]
+pub struct PositiveFiniteF32(f32);
```
ewienik: This is used for Diskann alpha parameter. Consider rename it to describe this, maybe something similar to `DiskannAlfa`?

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -76,3 +88,140 @@ fn new(index_key: crate::IndexKey) -> anyhow::Result<mpsc::Sender<VsIndex>> {
...
+                .ok_or(anyhow::anyhow!("expansion_search must be > 0"))?,
+        })
+    }
+}
+
+impl From<Dimensions> for usize {
+    fn from(dim: Dimensions) -> Self {
+        usize::from(dim.0)
+    }
+}
```
ewienik: Do we need this at all? If so, consider moving it close to the Dimensions definition.

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -49,19 +59,48 @@ impl VsIndexFactory for DiskannIndexFactory {
...
+    let index_dir = diskann_index_path.join(index_key.as_ref());
+
     let (tx, mut rx) = mpsc::channel(perf::channel_size().into());
 
     tokio::spawn(perf::hotpath_async(
         {
             async move {
                 debug!("starting");
 
+                async {
```
ewienik: This `async {` block seems to be not needed.

crates/vector-store/src/vs_index/diskann.rs
ewienik: We need to wipe out index directory while we create or delete index.

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -49,19 +59,48 @@ impl VsIndexFactory for DiskannIndexFactory {
...
+                    if tokio::fs::try_exists(&index_dir).await.unwrap_or(false) {
+                        let mut dir = tokio::fs::read_dir(&index_dir).await?;
+                        if dir.next_entry().await?.is_some() {
+                            anyhow::bail!("DiskANN index directory is non-empty: {index_dir:?}");
+                        }
+                    }
+
+                    tokio::fs::create_dir_all(&index_dir)
```
ewienik: nit: consider `use tokio::fs`

ewienik (changes requested):

Akvear:
Changes:
- removed parameters mapping from the scope as they are not needed yet
- fixed commit history to read better
- removed unused crates
- added directory wipe both in new_diskann and on index creation
- added unit tests for dir wipe

I am not sure whether it's needed to have all functions be async in the last commit, or if tokio has API that allows for something different.

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -37,19 +49,67 @@ impl VsIndexFactory for DiskannIndexFactory {
...
+        if path.is_dir() {
+            fs::remove_dir_all(&path)
+                .await
+                .context(format!("failed to remove DiskANN directory: {:?}", path))?;
+        } else {
+            fs::remove_file(&path)
+                .await
+                .context(format!("failed to remove DiskANN file: {:?}", path))?;
+        }
+    }
```
Akvear: Is it better/allowed to remove the directory entirely and recreate it?

ewienik: We could assume that the parent directory provided by env variable exists and the content we can remove/create how we want.

crates/vector-store/src/vs_index/diskann.rs
Akvear: I don't have a clear understanding of that yet.

crates/vector-store/src/vs_index/diskann.rs
Akvear: What about when the directory we want to wipe is the diskann root path from the env var?

ewienik: IMHO we shouldn't remove directory provided by the env - it is our parent. We can modify all its content, but not directory itself. I think we shouldn't create this directory also - it should be created for vector-store before.

Akvear: Ok, that's the way it is now.

Akvear:
Changes: 
- Removed passing the path env var
- Deleted mention of this env var in README
- Adjusted tests

ewienik (dismissed):

Akvear:
Changes:
- Added and document new env variable `VECTOR_STORE_USE_DISKANN` since alpha parameter is optional even when you want to use diskann (see main.rs engine selection)

Akvear:
Changes:
- revised error path behavior when creating a `DiskannAlpha` struct.

README.md
```diff
@@ -45,14 +47,14 @@ components run `rustup install` in the main directory of the repository.
...
-$ cargo b [-r]
-$ cargo r [-r]
+cargo b [-r]
+cargo r [-r]
 ```
 
 To install all cargo tools used in the CI:
 
 ```
-$ scripts/install-cargo-tools
```
QuerthDP, line 55: Why remove the `$`?

QuerthDP (commented):
Looks fine. Commit messages are still missing, but the titles are somewhat descriptive so let it be. One more question, let's resolve it and I'll approve.

README.md
Akvear: Auto-formatter must have changed this

Akvear: fixed

QuerthDP (approved):

PR #506, opened 2026-07-07
vector-store: expose index build progress in status endpoint

nuivall:
The GET /api/v1/indexes/{keyspace}/{index}/status handler now reports the
build_progress percentage taken from the index entry's full-scan progress
(Progress::Done maps to 100%). This lets clients observe how far the
initial backfill scan has advanced while an index is BOOTSTRAPPING.

Refs: https://scylladb.atlassian.net/browse/SCYLLADB-2327

### DIFF
--- a/api/openapi.json
+++ b/api/openapi.json
@@ -6,7 +6,7 @@
     "license": {
       "name": "LicenseRef-ScyllaDB-Source-Available-1.0"
     },
-    "version": "2.0.0"
+    "version": "2.1.0"
   },
   "paths": {
     "/api/v1/indexes": {
@@ -439,6 +439,13 @@
           "count"
         ],
         "properties": {
+          "build_progress": {
+            "type": "number",
+            "format": "double",
+            "description": "Progress of the initial full table scan that backfills the index,\nexpressed as a percentage in the range `0.0..=100.0`. This reflects\nonly how far the full scan has advanced; it is not a source of truth\nfor the index state. The `status` field is authoritative: an index\nreporting `SERVING` is serving regardless of `build_progress`, which\nmay not reach `100.0` (e.g. if some scan ranges failed). Defaults to\n`100.0` when omitted, so a new client talking to an older server that\ndoes not report the field falls back to relying solely on `status`.",
+            "maximum": 100,
+            "minimum": 0
+          },
           "count": {
             "type": "integer",
             "minimum": 0
--- a/crates/httpapi/src/lib.rs
+++ b/crates/httpapi/src/lib.rs
@@ -149,6 +149,21 @@ pub enum IndexStatus {
 pub struct IndexStatusResponse {
     pub status: IndexStatus,
     pub count: usize,
+    /// Progress of the initial full table scan that backfills the index,
+    /// expressed as a percentage in the range `0.0..=100.0`. This reflects
+    /// only how far the full scan has advanced; it is not a source of truth
+    /// for the index state. The `status` field is authoritative: an index
+    /// reporting `SERVING` is serving regardless of `build_progress`, which
+    /// may not reach `100.0` (e.g. if some scan ranges failed). Defaults to
+    /// `100.0` when omitted, so a new client talking to an older server that
+    /// does not report the field falls back to relying solely on `status`.
+    #[serde(default = "default_build_progress")]
+    #[schema(minimum = 0, maximum = 100)]
+    pub build_progress: f64,
+}
+
+fn default_build_progress() -> f64 {
+    100.0
 }
 
 #[derive(Debug, PartialEq, serde::Deserialize, serde::Serialize, utoipa::ToSchema)]
--- a/crates/vector-store/src/httproutes.rs
+++ b/crates/vector-store/src/httproutes.rs
@@ -91,7 +91,7 @@ use utoipa_swagger_ui::SwaggerUi;
             name = "LicenseRef-ScyllaDB-Source-Available-1.0"
         ),
         // version should be updated manually when there are changes in API
-        version = "2.0.0"
+        version = "2.1.0"
     ),
     tags(
         (
@@ -372,6 +372,10 @@ async fn get_index_status(
             response::Json(httpapi::IndexStatusResponse {
                 status: status.into(),
                 count,
+                // Placeholder; the real full-scan progress is wired up in a
+                // follow-up commit. Defaults to 100.0 so the field is always
+                // populated and `status` remains authoritative.
+                build_progress: 100.0,
             }),
         )
             .into_response(),
--- a/crates/vector-store/src/httproutes.rs
+++ b/crates/vector-store/src/httproutes.rs
@@ -344,12 +344,20 @@ async fn get_index_status(
         Fts(Sender<crate::fts_index::FtsIndex>),
     }
 
-    let (index, status) = {
+    let (index, status, progress) = {
         let indexes = state.indexes.read().unwrap();
         if let Some(entry) = indexes.get_vs(&index_key) {
-            (IndexSender::Vs(entry.index().clone()), entry.status())
+            (
+                IndexSender::Vs(entry.index().clone()),
+                entry.status(),
+                entry.progress(),
+            )
         } else if let Some(entry) = indexes.get_fts(&index_key) {
-            (IndexSender::Fts(entry.index().clone()), entry.status())
+            (
+                IndexSender::Fts(entry.index().clone()),
+                entry.status(),
+                entry.progress(),
+            )
         } else {
             let msg = format!("missing index: {keyspace_name}.{index_name}");
             debug!("get_index_status: {msg}");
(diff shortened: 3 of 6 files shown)


nuivall:
@DarioMirovic could you pre-check if this rust code makes sense?

crates/vector-store/src/httproutes.rs
```diff
@@ -89,7 +89,7 @@ use utoipa_swagger_ui::SwaggerUi;
             name = "LicenseRef-ScyllaDB-Source-Available-1.0"
         ),
         // version should be updated manually when there are changes in API
-        version = "2.0.0"
+        version = "2.1.0"
```
knowack1, line 94: nit: this change should be in first commit. Along the actual API change.

api/openapi.json
```diff
@@ -417,6 +417,11 @@
           "count"
         ],
         "properties": {
+          "build_progress": {
+            "type": "number",
+            "format": "double",
+            "description": "Progress of the initial full table scan that backfills the index,\nexpressed as a percentage in the range `0.0..=100.0`. It is `100.0`\nonce the index is `SERVING`. Defaults to `0.0` for backward\ncompatibility with older clients that omit the field."
+          },
```
knowack1, line 448: nit: This file (auto-generated) should be in fist commit alongside with the actual API change.

crates/httpapi/src/lib.rs
```diff
@@ -149,6 +149,12 @@ pub enum IndexStatus {
 pub struct IndexStatusResponse {
     pub status: IndexStatus,
     pub count: usize,
+    /// Progress of the initial full table scan that backfills the index,
+    /// expressed as a percentage in the range `0.0..=100.0`. It is `100.0`
+    /// once the index is `SERVING`. Defaults to `0.0` for backward
+    /// compatibility with older clients that omit the field.
```
knowack1: "Defaults to `0.0` for backward compatibility with older clients that omit the field." - I think "defaults to 0.0" is anyhow related with backward compatibility. I mean that this is backward compatible change regardless if the default value.

api/openapi.json
```diff
@@ -417,6 +417,11 @@
           "count"
         ],
         "properties": {
+          "build_progress": {
```
knowack1, line 442: nit: could specify:
`minimum: 0` 
`maximum: 100`

crates/vector-store/tests/integration/routing.rs
```diff
@@ -269,6 +271,97 @@ async fn ann_routes_to_serving_index_while_replacement_is_bootstrapping() {
...
 
+// While the initial full table scan that backfills an index is paused (held
+// open via blocking_scan_fn, simulating a paused/slow scan), the index status
+// endpoint must report it as BOOTSTRAPPING and surface the build progress
+// percentage reported by the database. Once the scan finishes the index
+// becomes SERVING with build_progress == 100.
+#[tokio::test]
+#[ntest::timeout(10_000)]
+#[cfg_attr(not(feature = "slow-test-hooks"), ignore = "requires slow-test-hooks")]
+async fn index_status_reports_build_progress_while_bootstrapping() {
```
knowack1: These tests are in the wrong file. `routing.rs` is meant for search query routing to the correct index. We could move them to `status.rs` or create a new `index_status.rs` file, as `status.rs` currently only covers node status.

knowack1 (dismissed):
A few comments on this. @nuivall  / @swasik , should I continue working on this PR since @nuivall  is now busy with new duties?

crates/httpapi/src/lib.rs
DarioMirovic: I'd say that we need default value of `100.0`.

The caller using the new client would be checking both `status` and `build_progress`.
New client + old server = use default value.
Current code: the caller never progresses, because `build_progress` hasn't reached 100%.
100.0 default: the caller effectively just waits for `status`, as do the current callers that don't have `build_status` exposed - they only look at `status`.

crates/httpapi/src/lib.rs
```diff
@@ -149,6 +149,12 @@ pub enum IndexStatus {
 pub struct IndexStatusResponse {
     pub status: IndexStatus,
     pub count: usize,
+    /// Progress of the initial full table scan that backfills the index,
+    /// expressed as a percentage in the range `0.0..=100.0`. It is `100.0`
+    /// once the index is `SERVING`. Defaults to `0.0` for backward
```
DarioMirovic: In `initial_scan`, if at least one `preform_range_scan` call fails, it drops permit, as it should, but never increments `completed_scan_length` for that range. The counter can never reach `u64::MAX`, so we'll never reach 100% build progress, even when we transition to the serving state:
```rust
impl From<u64> for Percentage {
    fn from(value: u64) -> Self {
        Percentage::try_from((value as f64 / u64::MAX as f64) * 100.0).unwrap()
    }
}

impl From<u64> for Progress {
    fn from(value: u64) -> Self {
        if value == u64::MAX {
            Progress::Done
        } else {
            Progress::InProgress(Percentage::from(value))
        }
    }
}
```

This is a pre-existing issue. The focus of this comment is compliance of the doc comment with the actual behavior.

crates/httpapi/src/lib.rs
ewienik: Currently build_progress is only information about the progress of fullscan - it shouldn't be taken as a source of truth for the index state. The valid information is the state of the index - if this is SERVING it means the index is serving regardless of build_progress. If build_progress doesn't reach 100% it might be a trigger for investigation, what happened.

crates/httpapi/src/lib.rs
ewienik: You're right, currently we are skipping errors without increasing fullscan progress. There is an issue to design better handling of errors during fullscan and cdc: https://scylladb.atlassian.net/browse/VECTOR-726

nuivall:
v2:
- Default build_progress 0.0 → 100.0 (fall back to status); doc reworded
- Added OpenAPI min/max 0–100
- Moved index_status tests to new index_status.rs
- Split commits: openapi + version bump into commit 1

crates/vector-store/tests/integration/routing.rs
```diff
@@ -276,6 +278,11 @@ async fn ann_routes_to_serving_index_while_replacement_is_bootstrapping() {
     assert_eq!(response.status(), StatusCode::OK);
 }
 
+// While the initial full table scan that backfills an index is paused (held
+// open via blocking_scan_fn, simulating a paused/slow scan), the index status
+// endpoint must report it as BOOTSTRAPPING and surface the build progress
+// percentage reported by the database. Once the scan finishes the index
+// becomes SERVING with build_progress == 100.
```
ewienik: nit: leftover? There are no changes to the ann_routes_to_newest_serving_index

crates/vector-store/tests/integration/routing.rs
```diff
@@ -42,11 +42,13 @@ fn ordered_timeuuid(time: u32) -> Uuid {
...
 
 /// A scan function that never completes, keeping the index bootstrapping.
-fn blocking_scan_fn() -> ScanFn {
+pub(crate) fn blocking_scan_fn() -> ScanFn {
     Box::new(|_tx| std::future::pending::<()>().boxed())
 }
 
-fn single_row_scan(pks: impl IntoIterator<Item = CqlValue> + Send + Sync + 'static) -> ScanFn {
+pub(crate) fn single_row_scan(
+    pks: impl IntoIterator<Item = CqlValue> + Send + Sync + 'static,
```
ewienik, line 50: nit: ',' leftover?

crates/vector-store/tests/integration/index_status.rs
```diff
@@ -0,0 +1,109 @@
+/*
+ * Copyright 2026-present ScyllaDB
+ * SPDX-License-Identifier: LicenseRef-ScyllaDB-Source-Available-1.0
```
ewienik: nit: LicenseRef-ScyllaDB-Source-Available-1.1

ewienik (dismissed):

crates/vector-store/tests/integration/routing.rs
nuivall: Making it pub(crate) pushes the signature to 106 chars, so rustfmt wraps the parameter list and adds the trailing comma; cargo fmt --check fails without it

nuivall:
v3:
- removed leftovers

ewienik (approved):

ewienik:
@DarioMirovic Do you have more comments?

nuivall:
> @DarioMirovic Do you have more comments?

Dario is also on holidays.

ewienik:
@nuivall Can you make rebase to master? Probably there are changes that are not visible in the new integration test module.

nuivall:
Rebased

ewienik (dismissed):

ewienik (approved):

PR #508, opened 2026-07-08
vector-store: fix node stuck in bootstrapping on index recreate

swasik:
## Problem

The Vector Store node status (`GET /api/v1/status`) never transitions to `SERVING` if a vector index is dropped and recreated (or any index is added/dropped) **while the node is still performing its initial bootstrap** — even though every individual index reports `SERVING`. The node stays in `BOOTSTRAPPING` indefinitely.

Reproduced on 1.8.0 and 1.5.0 (Docker, single-node ScyllaDB 2026.1.7).

## Root cause

In `node_state.rs`, the node reaches `Serving` only when the bootstrap-gate set `initial_idxs` empties (drained by `FullScanFinished`). `monitor_indexes` re-emits `DiscoveringIndexes` + `IndexesDiscovered` on every schema change; the discovery handler then **overwrote** `initial_idxs` with the currently-present indexes whenever status was `DiscoveringIndexes`. Indexes that had already finished their initial full scan got re-inserted and never emit `FullScanFinished` again, so the gate could never empty.

## Fix

- Key `initial_idxs` by `IndexKey` and add a `bootstrap_started` flag.
- Capture the initial index set once; on later mid-bootstrap discoveries, only **prune** indexes that no longer exist — never re-add.
- Never move a node that already reached `Serving` back to bootstrapping on a later schema change.
- Remove from the gate by key in `FullScanFinished` (robust to an index being recreated with a new version).

## Testing

- New regression test `node_reaches_serving_after_index_recreated_mid_bootstrap` replays the drop/recreate sequence — fails before, passes after.
- All `node_state` unit tests pass; `cargo fmt --check` and `cargo clippy` clean.
- End-to-end revalidated (100k×1532-dim vectors, drop+recreate mid-bootstrap): the node now reaches `SERVING` at ~51s (previously it never did).

Fixes VECTOR-730.

### DIFF
--- a/crates/vector-store/src/node_state.rs
+++ b/crates/vector-store/src/node_state.rs
@@ -104,7 +104,8 @@ pub(crate) async fn new() -> mpsc::Sender<NodeState> {
             debug!("starting");
 
             let mut status = NodeStatus::Initializing;
-            let mut initial_idxs = HashSet::new();
+            let mut bootstrap_started = false;
+            let mut initial_idxs = HashSet::<IndexKey>::new();
             let mut idxs = HashMap::<IndexKey, IndexStatus>::new();
             while let Some(msg) = rx.recv().await {
                 match msg {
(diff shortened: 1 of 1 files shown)


crates/vector-store/src/node_state.rs
```diff
@@ -125,28 +126,56 @@ pub(crate) async fn new() -> mpsc::Sender<NodeState> {
                                 continue;
                             }
 
-                            update_indexes(
-                                &mut idxs,
-                                indexes.iter().map(|meta| meta.key()).collect(),
-                            );
+                            let keys: HashSet<IndexKey> =
```
knowack1, line 129: To make this state machine more resilient to transition errors, we could consider adding a timer to periodically check the status of the indices. If all are serving, then the node changes its state to serving. I think this would be a more robust way to keep the node state in sync with the index states.

@ewienik @swasik what do you think?

swasik:
It was AI attempt to fix this issue that was diagnosed correctly but solved using alternative solution in #509.

crates/vector-store/src/node_state.rs
swasik: I like the idea but we can make it in the follow up to keep this fix simple and less risky.

PR #509, opened 2026-07-08
vector-store: fix overwritting NodeStatus::IndexingEmbeddings

ewienik:
The root cause for a bug was that initial_idx got overwritten in several paths of events. initial_idx stores the set of initial indexed, which was read during NodeStatus::DiscoveringIndexes state. After FullScanFinished the built index is removed from the set. When initial_idx is overwritten with the indexes which were already built, we wouldn't receive further events that the index is built, and we would stack forever in node bootstrapping.

Scenarios which are fixed with this commit and tested:
- discovering_indexes_while_bootstrapping: during bootstrapping scylladb schema change happens and node_state receives DiscoveringIndexes and IndexesDiscovered events
- db_reconnect_while_bootstrapping: during bootstrapping scylladb connection resets and node_state receives ConnectingToDb and ConnectedToDb events
- removing_index_while_bootstrapping: during bootstrapping one of the initial indexes is removed
- adding_index_while_bootstrapping: during bootstrapping a new index is added

Fixes: VECTOR-730
Fixes: VECTOR-731

### DIFF
--- a/crates/vector-store/src/node_state.rs
+++ b/crates/vector-store/src/node_state.rs
@@ -14,6 +14,7 @@ use tokio::sync::oneshot;
 use tracing::Instrument;
 use tracing::debug;
 use tracing::debug_span;
+use tracing::error;
 use tracing::info;
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
@@ -104,7 +105,7 @@ pub(crate) async fn new() -> mpsc::Sender<NodeState> {
             debug!("starting");
 
             let mut status = NodeStatus::Initializing;
-            let mut initial_idxs = HashSet::new();
+            let mut initial_idxs: Option<HashSet<_>> = None;
             let mut idxs = HashMap::<IndexKey, IndexStatus>::new();
             while let Some(msg) = rx.recv().await {
                 match msg {
@@ -112,17 +113,32 @@ pub(crate) async fn new() -> mpsc::Sender<NodeState> {
                         Event::ConnectingToDb => {
                             status = NodeStatus::ConnectingToDb;
                         }
+
                         Event::ConnectedToDb => {}
-                        Event::DiscoveringIndexes => {
-                            if status != NodeStatus::Serving {
-                                status = NodeStatus::DiscoveringIndexes;
+
+                        Event::DiscoveringIndexes => match status {
+                            NodeStatus::Initializing | NodeStatus::ConnectingToDb => {
+                                if let Some(initial_idxs) = &initial_idxs {
+                                    if initial_idxs.is_empty() {
+                                        // Initial indexes already built
+                                        status = NodeStatus::Serving;
+                                    } else {
+                                        // Still in the process of building initial indexes
+                                        status = NodeStatus::IndexingEmbeddings
+                                    }
+                                } else {
+                                    // Need to discover indexes first
+                                    status = NodeStatus::DiscoveringIndexes;
+                                }
                             }
-                        }
+                            _ => {}
+                        },
+
                         Event::IndexesDiscovered(indexes) => {
-                            if indexes.is_empty() && status != NodeStatus::Serving {
-                                status = NodeStatus::Serving;
-                                info!("Service is running, no indexes to build");
-                                continue;
+                            let mut initial = false;
+                            if initial_idxs.is_none() {
+                                initial = true;
+                                initial_idxs = Some(indexes.clone());
                             }
 
                             update_indexes(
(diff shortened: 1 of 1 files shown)


crates/vector-store/src/node_state.rs
```diff
@@ -113,11 +113,14 @@ pub(crate) async fn new() -> mpsc::Sender<NodeState> {
                             status = NodeStatus::ConnectingToDb;
```
ewienik, line 114: This lines also changes node status unconditionally when new session is created and it can lead to similar stuck in bootstrapping. Let's use some flag that initial indexes are not overwrite in this case.

crates/vector-store/src/node_state.rs
QuerthDP, line 1: > If any index finishes scanning (send
Event::FullScanFinished) initial_idxs got overwritten and we lost information
than the index is built - we stuck in NodeStatus::IndexingEmbeddings forever.

typo: than -> that

QuerthDP, line 1: > If any index finishes scanning (send
Event::FullScanFinished) initial_idxs got overwritten and we lost information
than the index is built - we stuck in NodeStatus::IndexingEmbeddings forever.

Could you please explain that more clearly? 
`initial_idxs` got overwritten to what? Why was it problematic?
We lost information necessary for which index, the built or the being build one?

QuerthDP (commented):
Please provide a clear explanaition. A path of states that happen to be problematic

swasik (dismissed):
I validated that it resolves the issue using the scenario described in VECTOR-730 so approving the change. Just resolve @QuerthDP comments.

swasik:
> Fixes: [VECTOR-731](https://scylladb.atlassian.net/browse/VECTOR-731)

Nit: also VECTOR-730.

[VECTOR-731]: https://scylladb.atlassian.net/browse/VECTOR-731?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

knowack1 (dismissed):

ewienik:
Changelog for 300dc7977
- fix refactored

<details>
<summary>range diff</summary>

```
1:  769171704 < -:  --------- vector-store: fix overwritting NodeStatus::IndexingEmbeddings
-:  --------- > 1:  300dc7977 vector-store: fix overwritting NodeStatus::IndexingEmbeddings
```

crates/vector-store/src/node_state.rs
ewienik: fixed in 300dc7977

crates/vector-store/src/node_state.rs
ewienik: fixed in 300dc7977

ewienik: fixed in 300dc7977

ewienik:
Changelog for 22ecaa6fc
- fix IndexesDisovered process

<details>
<summary>range diff</summary>

```
1:  300dc7977 ! 1:  22ecaa6fc vector-store: fix overwritting NodeStatus::IndexingEmbeddings
    @@ crates/vector-store/src/node_state.rs: pub(crate) async fn new() -> mpsc::Sender
     +                                    );
     +                                }
     +                                continue;
    -+                            }
    -+
    -+                            match status {
    -+                                NodeStatus::DiscoveringIndexes => {
    -+                                    status = NodeStatus::IndexingEmbeddings
    -+                                }
    -+                                NodeStatus::IndexingEmbeddings => {}
    -+                                NodeStatus::Serving => {}
    -+                                _ => status = NodeStatus::Serving,
                                  }
    ++
    ++                            // we have initial indexes to build, so we need to be in the
    ++                            // IndexingEmbeddings state
    ++                            status = NodeStatus::IndexingEmbeddings;
                              }
     +
                              Event::FullScanStarted(metadata) => {
```

swasik (dismissed):
Now it correctly handles both scenarios I tested - removing large index mid-scan and replacing small index. Verified experimentally.

ewienik:
Changelog for d6bd26c1e
- fix typo

<details>
<summary>range diff</summary>

```
1:  22ecaa6fc ! 1:  d6bd26c1e vector-store: fix overwritting NodeStatus::IndexingEmbeddings
    @@ crates/vector-store/src/node_state.rs: pub(crate) async fn new() -> mpsc::Sender
                                  if initial_idxs.is_empty() && status != NodeStatus::Serving {
                                      status = NodeStatus::Serving;
     -                                info!("Service is running, finished building indexes");
    -+                                info!("Service is running, finished building intitial indexes");
    ++                                info!("Service is running, finished building initial indexes");
                                  }
                              }
                          },
```

swasik (approved):

crates/vector-store/src/node_state.rs
```diff
@@ -104,51 +105,93 @@ pub(crate) async fn new() -> mpsc::Sender<NodeState> {
...
-                        Event::DiscoveringIndexes => {
-                            if status != NodeStatus::Serving {
-                                status = NodeStatus::DiscoveringIndexes;
+
+                        Event::DiscoveringIndexes => match status {
+                            NodeStatus::Initializing | NodeStatus::ConnectingToDb => {
+                                if let Some(initial_idxs) = &initial_idxs {
+                                    if initial_idxs.is_empty() {
+                                        // Initial indexes already built
+                                        status = NodeStatus::Serving;
```
QuerthDP, line 124: Do we print any logs "Service is running" in that case? 

If so, it is the `IndexesDiscovered` branch right? Then we're setting the `NodeStatus::Serving` there anyway. Maybe we should have done this once.

ewienik: When DiscoveringIndexes is received while node is connecting to the db, and initial indexes were built, it means we have connection to the database, so the status should be now Serving.

PR #510, opened 2026-07-08
Add standard text analyzer to FTS index

knowack1:
Introduces a custom "standard" analyzer for the Tantivy full-text search index. Text in the `body` field is now tokenized by splitting on punctuation and whitespace, lowercased for case-insensitive matching, filtered for English stop words, and capped to drop excessively long tokens, while preserving Unicode alphanumeric characters.

Fixes VECTOR-618

### DIFF
--- a/crates/vector-store/src/fts_index/tantivy.rs
+++ b/crates/vector-store/src/fts_index/tantivy.rs
@@ -15,10 +15,17 @@ use tantivy::collector::TopDocs;
 use tantivy::indexer::IndexWriterOptions;
 use tantivy::query::QueryParser;
 use tantivy::schema::INDEXED;
+use tantivy::schema::IndexRecordOption;
 use tantivy::schema::STORED;
 use tantivy::schema::Schema;
-use tantivy::schema::TEXT;
+use tantivy::schema::TextFieldIndexing;
+use tantivy::schema::TextOptions;
 use tantivy::schema::Value;
+use tantivy::tokenizer::Language;
+use tantivy::tokenizer::LowerCaser;
+use tantivy::tokenizer::SimpleTokenizer;
+use tantivy::tokenizer::StopWordFilter;
+use tantivy::tokenizer::TextAnalyzer;
 use tokio::sync::mpsc;
 use tokio::sync::watch;
 use tracing::debug;
@@ -72,10 +79,15 @@ struct IndexState {
     schema: Schema,
 }
 
+const TOKENIZER_NAME: &str = "standard";
+
 impl IndexState {
     fn new() -> anyhow::Result<Self> {
         let schema = build_schema();
         let index = tantivy::Index::create_in_ram(schema.clone());
+        index
+            .tokenizers()
+            .register(TOKENIZER_NAME, build_standard_analyzer()?);
         let options = IndexWriterOptions::builder()
             .num_worker_threads(perf::num_workers().into())
             .build();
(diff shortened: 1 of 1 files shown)


crates/vector-store/src/fts_index/tantivy.rs
```diff
@@ -433,6 +462,16 @@ mod tests {
         tx
     }
 
+    fn tokenize_with_standard_analyzer(text: &str) -> Vec<String> {
+        let mut analyzer = build_standard_analyzer();
```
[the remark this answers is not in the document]
knowack1: Skipped — this is a false positive. `TextAnalyzer::token_stream` in tantivy 0.26.1 has the signature `fn token_stream<'a>(&'a mut self, text: &'a str) -> BoxTokenStream<'a>`, i.e. it borrows `self` mutably. The `mut` binding is therefore required; removing it fails to compile with "cannot borrow `analyzer` as mutable". No `unused_mut` warning is produced (verified with `cargo clippy -p vector-store --all-targets`, which finished clean).

crates/vector-store/src/fts_index/tantivy.rs
```diff
@@ -94,10 +107,26 @@ impl IndexState {
     }
 }
 
+fn build_standard_analyzer() -> TextAnalyzer {
+    let stop_words = StopWordFilter::new(Language::English).expect("english stop words available");
+    TextAnalyzer::builder(SimpleTokenizer::default())
```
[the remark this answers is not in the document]
knowack1: Fixed.

crates/vector-store/src/fts_index/tantivy.rs
```diff
@@ -94,10 +107,27 @@ impl IndexState {
     }
 }
 
+fn build_standard_analyzer() -> anyhow::Result<TextAnalyzer> {
+    let stop_words = StopWordFilter::new(Language::English)
+        .ok_or_else(|| anyhow!("fts: english stop words unavailable"))?;
+    Ok(TextAnalyzer::builder(SimpleTokenizer::default())
+        .filter(RemoveLongFilter::limit(40))
```
ewienik: Consider providing named const with some explanation why `40` is the limit.

crates/vector-store/src/fts_index/tantivy.rs
```diff
@@ -94,10 +107,27 @@ impl IndexState {
     }
 }
 
+fn build_standard_analyzer() -> anyhow::Result<TextAnalyzer> {
+    let stop_words = StopWordFilter::new(Language::English)
```
ewienik, line 110: Will we have in the future other languages?

ewienik (dismissed):

crates/vector-store/src/fts_index/tantivy.rs
knowack1: Yes. There is plan to introduce other languages analyzers. This is the standard analyzer as defined in https://scylladb.atlassian.net/browse/VECTOR-618.

crates/vector-store/src/fts_index/tantivy.rs
knowack1: I took this from the Tantivy examples:
 ```rust
 use tantivy::tokenizer::*;

 let en_stem = TextAnalyzer::builder(SimpleTokenizer::default())
     .filter(RemoveLongFilter::limit(40))
     .filter(LowerCaser)
     .filter(Stemmer::new(Language::English))
     .build();
```

 but since there is no requirement for it and I'm not sure we really need it, I'm removing this option for now.

ewienik (approved):

PR #515, opened 2026-07-10
vector-store: add CDC reader health metrics

QuerthDP:
Adds Prometheus metrics for CDC reader health, as a follow-up to a recent
incident where a CDC reader silently stopped processing changes without
any visible signal.

- `cdc_reader_up` - whether each index's CDC reader (wide/fine) is  currently running or stopped.
- `cdc_handler_errors_total` - CDC handler errors per index and reader.
- `cdc_reader_restarts_total` - restart-after-backoff attempts per index  and reader.

Also fixes two issues found while testing this work:
- a pre-existing race in the CDC handler shutdown path that could  orphan the driver's per-stream fetch tasks indefinitely instead of  stopping them.
- a pre-existing leak of `indexing_lag_seconds` series on index  deletion.

Split from: https://github.com/scylladb/vector-store/pull/503 as it kept getting bigger.
This patch will introduce the straightforward metrics that could be supported out of the box.
The latter will include the `cdc_last_processed_timestamp_seconds` metric and a potential refactor of `db_cdc.rs`.

Refs: VECTOR-722
Fixes: VECTOR-739

### DIFF
--- a/crates/vector-store/src/db.rs
+++ b/crates/vector-store/src/db.rs
@@ -19,6 +19,7 @@ use crate::IndexMetadata;
 use crate::IndexName;
 use crate::IndexVersion;
 use crate::KeyspaceName;
+use crate::Metrics;
 use crate::NonemptyArc;
 use crate::NonemptyIteratorExt;
 use crate::Quantization;
@@ -257,6 +258,7 @@ pub(crate) async fn new(
     node_state: Sender<NodeState>,
     internals: Sender<Internals>,
     mut config_rx: watch::Receiver<Arc<Config>>,
+    metrics: Arc<Metrics>,
 ) -> anyhow::Result<mpsc::Sender<Db>> {
     let (tx, mut rx) = mpsc::channel(perf::channel_size().into());
     tokio::spawn(
@@ -286,7 +288,7 @@ pub(crate) async fn new(
                                     internals.create_session(Some(session.clone())).await;
                                     session_tx.send(Some(session)).ok();
                                     if statements.is_none() {
-                                        statements = Some(Arc::new(Statements::new(config_rx.clone(), session_rx.clone()).await.unwrap()));
+                                        statements = Some(Arc::new(Statements::new(config_rx.clone(), session_rx.clone(), metrics.clone()).await.unwrap()));
                                     }
                                     info!("Connected to ScyllaDB at {}", config.scylladb_uri);
                                 }
@@ -505,6 +507,7 @@ fn credentials_changed(
 struct Statements {
     config_rx: watch::Receiver<Arc<Config>>,
     session_rx: watch::Receiver<Option<Arc<Session>>>,
+    metrics: Arc<Metrics>,
     st_latest_schema_version: PreparedStatement,
     st_get_indexes: PreparedStatement,
     st_get_index_target_type: PreparedStatement,
@@ -641,6 +644,7 @@ impl Statements {
     async fn new(
         config_rx: watch::Receiver<Arc<Config>>,
         session_rx: watch::Receiver<Option<Arc<Session>>>,
+        metrics: Arc<Metrics>,
     ) -> anyhow::Result<Self> {
         let session = session_rx.borrow().clone().ok_or_else(|| {
             anyhow::anyhow!("No session available during Statements initialization")
@@ -648,6 +652,7 @@ impl Statements {
 
         Ok(Self {
             config_rx,
+            metrics,
 
             st_latest_schema_version: session
                 .prepare(Self::ST_LATEST_SCHEMA_VERSION)
@@ -689,6 +694,7 @@ impl Statements {
             metadata,
             node_state,
             internals,
+            self.metrics.clone(),
             cdc_error_notify,
         )
         .await
--- a/crates/vector-store/src/db_cdc.rs
+++ b/crates/vector-store/src/db_cdc.rs
@@ -384,6 +384,7 @@ impl CdcReaderState {
     ) {
         let session = session_rx.borrow().clone();
         if let Some(session) = session {
+            record_reader_restart(&self.metrics, &self.keyspace, &self.index_name, self.name);
             let config = config_rx.borrow().clone();
             let params = (self.params_fn)(&config);
             self.restart(params, &session, metadata, tx_embeddings, internals)
@@ -401,6 +402,19 @@ fn record_handler_error(
 ) {
     metrics
         .cdc_handler_errors_total
+        .with_label_values(&[keyspace.as_ref(), index_name.as_ref(), reader])
+        .inc();
+}
+
+/// Increments the `cdc_reader_restarts_total` counter for the given index and reader.
+fn record_reader_restart(
+    metrics: &Metrics,
+    keyspace: &KeyspaceName,
+    index_name: &IndexName,
+    reader: &str,
+) {
+    metrics
+        .cdc_reader_restarts_total
         .with_label_values(&[keyspace.as_ref(), index_name.as_ref(), reader])
         .inc();
 }
--- a/crates/vector-store/src/db_cdc.rs
+++ b/crates/vector-store/src/db_cdc.rs
@@ -10,6 +10,7 @@ use crate::DbIndexedRow;
 use crate::DbIndexedValue;
 use crate::IndexKind;
 use crate::IndexMetadata;
+use crate::Metrics;
 use crate::NonemptyArc;
 use crate::NonemptyIteratorExt;
 use crate::db_index_backend::CdcValueStatus;
@@ -117,6 +118,7 @@ pub(crate) fn new(
     mut session_rx: watch::Receiver<Option<Arc<Session>>>,
     metadata: IndexMetadata,
     internals: Sender<Internals>,
+    _metrics: Arc<Metrics>,
     tx_embeddings: mpsc::Sender<(DbIndexedRow, AsyncInProgress)>,
     config: CdcReaderConfig,
 ) -> mpsc::Sender<DbCdc> {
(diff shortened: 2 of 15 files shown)


crates/vector-store/src/db_cdc.rs
```diff
@@ -96,27 +97,44 @@ impl CdcReaderParams {
...
 }
 
-impl From<CdcReaderConfig> for CdcReaderState {
+impl From<CdcReaderConfig> for (&'static str, fn(&Config) -> CdcReaderParams) {
     fn from(config: CdcReaderConfig) -> Self {
         match config {
-            CdcReaderConfig::Wide => Self::new("wide", CdcReaderParams::wide),
-            CdcReaderConfig::Fine => Self::new("fine", CdcReaderParams::fine),
+            CdcReaderConfig::Wide => (READER_WIDE, CdcReaderParams::wide),
+            CdcReaderConfig::Fine => (READER_FINE, CdcReaderParams::fine),
```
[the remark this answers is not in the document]
QuerthDP: It does compile

crates/vector-store/src/db_cdc.rs
```diff
@@ -402,16 +402,20 @@ fn spawn_handler_task(
...
+                    result = &mut handler => {
+                        if let Err(err) = result {
+                            warn!("CDC handler error: {err}");
+                            internals.increment_counter(errors_counter_name).await;
+                            cdc_error_notify.notify_one();
+                        }
+                        break;
+                    }
+                    _ = shutdown_notify.notified() => {
+                        debug!("CDC handler: shutdown requested, waiting for the reader to stop");
```
ewienik: > `spawn_handler_task` raced the driver's supervisor future against
a shutdown `Notify` in a `tokio::select!`. The Notify branch almost
always won, dropping the supervisor future before it could propagate
the new stop timestamp to its internally spawned per-stream fetch tasks.
Those tasks are plain `tokio::spawn` + `JoinHandle` with no abort mechanism,
so they were orphaned and kept running (and checkpointing) forever.

Is it true? when shutdown_notify is notified it means also cdc was stopped - so handler spawned task ought to finish. Is there a really need for this loop over select?

crates/vector-store/src/db.rs
```diff
@@ -140,7 +142,7 @@ pub enum Db {
 }
 
 pub(crate) trait DbExt {
-    async fn get_db_index(&self, metadata: IndexMetadata) -> GetDbIndexR;
+    async fn get_db_index(&self, metadata: IndexMetadata, metrics: Arc<Metrics>) -> GetDbIndexR;
```
ewienik: Consider passing metrics into db::new once instead of every get_db_index.

crates/vector-store/src/db_cdc.rs
```diff
@@ -115,10 +115,16 @@ impl From<CdcReaderConfig> for (&'static str, fn(&Config) -> CdcReaderParams) {
...
+    fn from((config, metadata, metrics): (CdcReaderConfig, &IndexMetadata, Arc<Metrics>)) -> Self {
         let (name, params_fn) = config.into();
-        Self::new(name, params_fn)
+        Self::new(
+            name,
+            params_fn,
+            metrics,
+            metadata.keyspace_name.as_ref().to_string(),
+            metadata.index_name.as_ref().to_string(),
+        )
```
ewienik: It seems better to introduce CdcReaderState::new ctor instead of this tuple conversion

crates/vector-store/src/db_cdc.rs
```diff
@@ -213,19 +221,54 @@ struct CdcReaderState {
     start: Duration,
     name: &'static str,
     params_fn: fn(&Config) -> CdcReaderParams,
+    metrics: Arc<Metrics>,
+    keyspace: String,
```
ewienik: try to use KespaceName type

crates/vector-store/src/db_cdc.rs
```diff
@@ -213,19 +221,54 @@ struct CdcReaderState {
...
+    index_name: String,
```
ewienik: try to use IndexName type

crates/vector-store/src/db_cdc.rs
```diff
@@ -213,19 +221,54 @@ struct CdcReaderState {
...
+    /// - `true` -> 1 (running),
+    /// - `false` -> 0 (stopped).
+    fn set_reader_up(&self, up: bool) {
+        set_reader_up(
+            &self.metrics,
+            &self.keyspace,
+            &self.index_name,
+            self.name,
+            up,
+        );
```
ewienik: This function is called once here as a forward, isn't it? If so, consider moving the body directly here.

crates/vector-store/src/db_cdc.rs
```diff
@@ -213,19 +221,54 @@ struct CdcReaderState {
...
+            index_name,
+        };
+        state.set_reader_up(false);
+        state
+    }
+
+    /// Sets the `cdc_reader_up` gauge for this reader:
+    /// - `true` -> 1 (running),
+    /// - `false` -> 0 (stopped).
+    fn set_reader_up(&self, up: bool) {
```
ewienik: Nit: consider providing two methods: set_reader_metric_up and set_reader_metric_down(). set_reader_up can suggest that you setup reader up.

ewienik (changes requested):

crates/vector-store/src/db_cdc.rs
QuerthDP: `reader.stop()` does correctly signal the driver - that's not in question. 

`shutdown_notify` firing means "shutdown was requested," not "shutdown finished." The driver needs a moment to notice the new stop timestamp and wind down its internal per-stream tasks, which it can only do if we keep polling it. 

The old single-shot `select!` treated the two as the same instant: since the notify was already signaled before we started waiting, it won the race immediately and `select!` dropped the driver future before it could finish that cleanup - orphaning those tasks.

The loop doesn't need multiple notifications; it just makes sure a "shutdown requested" signal only logs and keeps waiting, rather than cancelling the driver mid-cleanup. The only way to actually exit is the driver finishing on its own.

QuerthDP: Moved to https://github.com/scylladb/vector-store/pull/518 for further consideration.

QuerthDP:
Changelog:
- thread Metrics through `db::new` instead of `get_db_index`
- use `KeyspaceName` and `IndexName`
- remove unnecessary helper function
- change `set_reader_up(bool)` to `set_reader_metric_up()` and `set_reader_metric_down()`
- move the CDC reader lifetime commit to https://github.com/scylladb/vector-store/pull/518
- use `CdcReaderState::new` instead of tuple `From` impl

QuerthDP:
Changelog:
- fix `cdc_reader_restarts_total` metric helper to mention it counts attempts not restarts as requested https://github.com/scylladb/vector-store/pull/503#discussion_r3577514615

crates/vector-store/src/lib.rs
```diff
@@ -693,7 +693,25 @@ pub async fn run(
...
+}
+
+pub async fn run_with_metrics(
+    node_state: Sender<NodeState>,
+    db_actor: Sender<Db>,
+    internals: Sender<Internals>,
+    index_factory: Box<dyn VsIndexFactory + Send + Sync>,
+    receivers: ConfigReceivers,
+    metrics: Arc<Metrics>,
+) -> anyhow::Result<(Sender<HttpServer>, Sender<HttpServer>)> {
```
ewienik: As we require now Metrics for `new_db` and we have `new_metrics` function we don't need two separate `run*` functions - IMHO it could lead into errors in the future. What do you think about adding metrics arg to the single `run` function?

crates/vector-store/tests/integration/db_basic.rs
```diff
@@ -259,7 +259,7 @@ impl DbBasic {
 
 fn process_db(db: &DbBasic, msg: Db, node_state: Sender<NodeState>) {
     match msg {
-        Db::GetDbIndex { metadata, tx } => tx
+        Db::GetDbIndex { metadata, tx, .. } => tx
```
ewienik: nit: leftover?

crates/vector-store/src/db.rs
```diff
@@ -1188,7 +1194,7 @@ pub(crate) mod tests {
 
                 while let Some(msg) = rx.recv().await {
                     match msg {
-                        Db::GetDbIndex { metadata, tx } => sim.get_db_index(metadata, tx).await,
+                        Db::GetDbIndex { metadata, tx, .. } => sim.get_db_index(metadata, tx).await,
```
ewienik: nit: leftover?

crates/vector-store/src/db_cdc.rs
```diff
@@ -391,6 +392,14 @@ impl CdcReaderState {
     }
 }
 
+/// Increments the `cdc_handler_errors_total` counter for the given index and reader.
+fn record_handler_error(metrics: &Metrics, keyspace: &str, index_name: &str, reader: &str) {
```
ewienik: Consider using here also KeyspaceName and IndexName types

ewienik (changes requested):

crates/vector-store/src/lib.rs
QuerthDP: I didn't want to add a new arg to `run` function as it would require to change all the `run` calls in tests which was unnecessary. I may change that and leave us with only one `run_*` function, it's up for preference IMO.

QuerthDP: Changed to use `run`

QuerthDP:
Changelog:
- remove `run_with_metrics`, instead add `Metrics` as an argument to `run`
- remove leftovers
- use `KeyspaceName` and `IndexName` in `record_*` functions

ewienik (approved):

PR #518, opened 2026-07-13
vector-store: stop dropping the CDC driver's supervisor future on shutdown

QuerthDP:
`spawn_handler_task` raced the driver's supervisor future against a shutdown `Notify` in a `tokio::select!`. The Notify branch almost always won, dropping the supervisor future before it could propagate the new stop timestamp to its internally spawned per-stream fetch tasks. Those tasks are plain `tokio::spawn` + `JoinHandle` with no abort mechanism, so they were orphaned and kept running (and checkpointing) forever.

Fixed by pinning the handler future and looping until it resolves on its own once shutdown is requested, rather than dropping it.

### DIFF
--- a/crates/vector-store/src/db_cdc.rs
+++ b/crates/vector-store/src/db_cdc.rs
@@ -402,16 +402,20 @@ fn spawn_handler_task(
     tokio::spawn(
         async move {
             internals.increment_counter(started_counter_name).await;
-            tokio::select! {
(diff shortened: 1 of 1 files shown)


wprzytula:
This only hides the real problem. The correct fix: https://github.com/scylladb/scylla-cdc-rust/pull/165.

PR #519, opened 2026-07-14
vector-store: add FTS metrics

QuerthDP:
Adds Prometheus metrics for full-text search (FTS) indexes, exposed via the existing `/metrics` endpoint.

New gauges, refreshed on scrape for indexes with pending writes:
- `fts_index_size_bytes` - in-memory size of a Tantivy index.
- `fts_segment_count` - number of tantivy segments in a Tantivy index.

Reused existing, index-type-agnostic instruments instead of duplicating them under an `fts_` prefix:
- `fts_query_latency_seconds` -> `request_latency_seconds` (now also timed around the `bm25` query handler)
- `fts_query_count_total` -> `request_latency_seconds_count` (comes for free with the histogram)
- `fts_index_doc_count` -> `index_size`
- `fts_indexing_rate` -> `index_modified` (raw counter; apply `rate()` in PromQL)
- `fts_indexing_lag_seconds` -> `indexing_lag_seconds` (only recorded for CDC-sourced writes, not full-scan indexing)

Deferred to a follow-up: `fts_merge_duration_seconds`. Tantivy has no completion hook for automatic background merges, so measuring it requires disabling auto-merge and driving/timing merges manually - a bigger, separate
change.

Fixes: VECTOR-629
Follow-up: VECTOR-743

### DIFF
--- a/crates/vector-store/src/engine.rs
+++ b/crates/vector-store/src/engine.rs
@@ -14,6 +14,7 @@ use crate::db::Db;
 use crate::db::DbExt;
 use crate::db_index::DbIndex;
 use crate::db_index::DbIndexExt;
+use crate::fts_index::FtsIndex;
 use crate::fts_index::FtsIndexFactory;
 use crate::indexes::Indexes;
 use crate::memory;
@@ -44,6 +45,7 @@ use tracing::trace;
 type GetVsIndexKeysR = Vec<(IndexKey, crate::IndexOptionsVs)>;
 type AddIndexR = anyhow::Result<()>;
 type GetVsIndexR = Option<(mpsc::Sender<VsIndex>, mpsc::Sender<DbIndex>)>;
+type GetFtsIndexR = Option<(mpsc::Sender<FtsIndex>, mpsc::Sender<DbIndex>)>;
 
 pub(crate) enum Engine {
     GetVsIndexKeys {
@@ -60,6 +62,10 @@ pub(crate) enum Engine {
         key: IndexKey,
         tx: oneshot::Sender<GetVsIndexR>,
     },
+    GetFtsIndex {
+        key: IndexKey,
+        tx: oneshot::Sender<GetFtsIndexR>,
+    },
 }
 
 pub(crate) trait EngineExt {
@@ -67,6 +73,7 @@ pub(crate) trait EngineExt {
     async fn add_index(&self, metadata: IndexMetadata) -> AddIndexR;
     async fn del_index(&self, key: IndexKey);
     async fn get_vs_index(&self, key: IndexKey) -> GetVsIndexR;
+    async fn get_fts_index(&self, key: IndexKey) -> GetFtsIndexR;
 }
 
 impl EngineExt for mpsc::Sender<Engine> {
@@ -101,6 +108,15 @@ impl EngineExt for mpsc::Sender<Engine> {
             .expect("EngineExt::get_vs_index: internal actor should receive request");
         rx.await
             .expect("EngineExt::get_vs_index: internal actor should send response")
+    }
+
+    async fn get_fts_index(&self, key: IndexKey) -> GetFtsIndexR {
+        let (tx, rx) = oneshot::channel();
+        self.send(Engine::GetFtsIndex { key, tx })
+            .await
+            .expect("EngineExt::get_fts_index: internal actor should receive request");
+        rx.await
+            .expect("EngineExt::get_fts_index: internal actor should send response")
     }
 }
 
@@ -163,6 +179,10 @@ pub(crate) async fn new(
                             Engine::DelIndex { key } => del_index(key, &indexes, &metrics).await,
 
                             Engine::GetVsIndex { key, tx } => get_vs_index(key, tx, &indexes).await,
+
+                            Engine::GetFtsIndex { key, tx } => {
+                                get_fts_index(key, tx, &indexes).await
+                            }
 
                         }
                     }
@@ -358,6 +378,20 @@ async fn get_vs_index(key: IndexKey, tx: oneshot::Sender<GetVsIndexR>, indexes: 
     );
 }
 
+async fn get_fts_index(
+    key: IndexKey,
+    tx: oneshot::Sender<GetFtsIndexR>,
+    indexes: &RwLock<Indexes>,
+) {
+    _ = tx.send(
+        indexes
+            .read()
+            .unwrap()
+            .get_fts(&key)
+            .map(|entry| (entry.index().clone(), entry.db_index())),
+    );
+}
+
 async fn update_indexes(node_state: &Sender<NodeState>, indexes: &RwLock<Indexes>) {
     let actual_indexes: Vec<_> = {
         let indexes = indexes.read().unwrap();
@@ -428,6 +462,12 @@ pub(crate) mod tests {
             key: IndexKey,
             tx: oneshot::Sender<GetVsIndexR>,
         ) -> impl Future<Output = ()> + Send + 'static;
+
+        fn get_fts_index(
+            &self,
+            key: IndexKey,
+            tx: oneshot::Sender<GetFtsIndexR>,
+        ) -> impl Future<Output = ()> + Send + 'static;
     }
 
     pub(crate) fn new(sim: impl SimEngine + Send + 'static) -> mpsc::Sender<Engine> {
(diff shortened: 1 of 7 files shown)


QuerthDP:
This PR is ready for review. Waiting for https://github.com/scylladb/vector-store/pull/515 and https://github.com/scylladb/vector-store/pull/503 to get in to rebase on them and open.

crates/vector-store/src/metrics.rs
```diff
@@ -21,6 +21,12 @@ pub struct Metrics {
     pub size: GaugeVec,
     pub modified: CounterVec,
     pub indexing_lag: HistogramVec,
+    pub cdc_reader_up: GaugeVec,
+    pub cdc_handler_errors_total: CounterVec,
+    pub cdc_reader_restarts_total: CounterVec,
+    pub cdc_last_processed_timestamp_seconds: GaugeVec,
+    pub fts_index_size_bytes: GaugeVec,
+    pub fts_segment_count: GaugeVec,
```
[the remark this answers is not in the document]
QuerthDP: Not related. Those are from PR's this PR is based on.

crates/vector-store/src/fts_index/tantivy.rs
```diff
@@ -592,6 +622,35 @@ mod tests {
...
+
+        add_doc(&sender, 1, "hello world").await;
+        add_doc(&sender, 2, "foo bar").await;
+
+        let key = make_index_key();
+        let stats = sender.stats(key).await.unwrap();
+
+        assert_eq!(stats.num_docs, 2);
+        assert_eq!(stats.segment_count, 2);
+        assert!(stats.size_bytes > 0);
```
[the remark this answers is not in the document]
QuerthDP: Changed to > 0

QuerthDP:
@ewienik please review

crates/vector-store/src/httproutes.rs
```diff
@@ -380,14 +380,44 @@ async fn get_metrics(
...
+                        .set(count as f64);
+                }
+            }
+            None => {
+                let index = state
+                    .indexes
+                    .read()
+                    .unwrap()
+                    .get_fts(&key)
+                    .map(|entry| entry.index().clone());
```
ewienik: consider using state.engine.get_fts_index to follow pattern with get_vs_index

ewienik (dismissed):
@knowack1 Do you have any comments?

crates/vector-store/src/httproutes.rs
QuerthDP: There is no `state.engine.get_fts_index`

crates/vector-store/src/metrics.rs
```diff
@@ -139,6 +141,24 @@ impl Metrics {
         )
         .unwrap();
 
+        let fts_index_size_bytes = GaugeVec::new(
+            prometheus::Opts::new(
+                "fts_index_size_bytes",
+                "Total in-memory size of a full-text search index (bytes)",
```
knowack1: Its in-memory today but could be on this in the future. Consider removing "in-memory" if we want this metric to also work for potential future on disk indexes.

crates/vector-store/src/metrics.rs
```diff
@@ -139,6 +141,24 @@ impl Metrics {
...
+            ),
+            &["keyspace", "index_name"],
+        )
+        .unwrap();
+
+        let fts_segment_count = GaugeVec::new(
+            prometheus::Opts::new(
+                "fts_segment_count",
+                "Number of tantivy segments in a full-text search index",
```
knowack1: I would not expose tantivy here. I think we should treat this as impl detail.

crates/vector-store/src/httproutes.rs
ewienik: It could be good to add this

crates/vector-store/src/httproutes.rs
```diff
@@ -761,18 +761,28 @@ async fn post_index_bm25(
...
+            timer.observe_duration();
+
             let msg = format!("missing index: {keyspace}.{index_name}");
             debug!("post_index_bm25: {msg}");
             return (StatusCode::NOT_FOUND, msg).into_response();
         };
         if entry.status() != crate::node_state::IndexStatus::Serving {
             match entry.progress() {
                 Progress::InProgress(percentage) => {
+                    timer.observe_duration();
```
knowack1, line 817: nit: I wonder if it makes a difference to measure latency explicitly here instead of depending on measured latency in timer drop. So that observe does not need to be called explicitly

crates/vector-store/src/httproutes.rs
```diff
@@ -380,14 +380,44 @@ async fn get_metrics(
...
+                let index = state
+                    .indexes
+                    .read()
+                    .unwrap()
+                    .get_fts(&key)
+                    .map(|entry| entry.index().clone());
+                if let Some(index) = index
+                    && let Ok(stats) = index.stats(key).await
+                {
+                    let labels = [keyspace.as_ref(), index_name.as_ref()];
```
knowack1: this var can be moved on top of this function as is also used in vs branch.

crates/vector-store/src/httproutes.rs
```diff
@@ -380,14 +380,44 @@ async fn get_metrics(
...
         let key = IndexKey::new(&keyspace, &index_name);
-        if let Some((index, _)) = state.engine.get_vs_index(key.clone()).await
-            && let Ok(count) = index.count(key).await
-        {
-            state
-                .metrics
-                .size
-                .with_label_values(&[keyspace.as_ref(), index_name.as_ref()])
-                .set(count as f64);
+        match state.engine.get_vs_index(key.clone()).await {
```
knowack1: There are metrics.rs integration tests for this endpoint. Consider adding verification of the new metrics there.

crates/vector-store/src/httproutes.rs
```diff
@@ -380,14 +380,44 @@ async fn get_metrics(
...
+                        .set(stats.size_bytes as f64);
+                    state
+                        .metrics
+                        .fts_segment_count
+                        .with_label_values(&labels)
+                        .set(stats.segment_count as f64);
+                }
+            }
         }
     }
```
knowack1, line 425: Consider moving to dedicated function:
```
    match state.engine.get_vs_index(key.clone()).await {  
            Some((index, _)) => {
                if let Ok(count) = index.count(key).await {
                    state
                        .metrics
                        .size
                        .with_label_values(&[keyspace.as_ref(), index_name.as_ref()])
                        .set(count as f64);
                }
            }
            None => {
                let index = state
                    .indexes
                    .read()
                    .unwrap()
                    .get_fts(&key)
                    .map(|entry| entry.index().clone());
                if let Some(index) = index
                    && let Ok(stats) = index.stats(key).await
                {
                    let labels = [keyspace.as_ref(), index_name.as_ref()]; 
                    state
                        .metrics
                        .size
                        .with_label_values(&labels)
                        .set(stats.num_docs as f64);
                    state
                        .metrics
                        .fts_index_size_bytes
                        .with_label_values(&labels)
                        .set(stats.size_bytes as f64);
                    state
                        .metrics
                        .fts_segment_count
                        .with_label_values(&labels)
                        .set(stats.segment_count as f64);
                }
            }
        }
    }
```

crates/vector-store/src/httproutes.rs
QuerthDP: That's true. I'd prefer the dropping instead as it makes sure we won't forget to add any explicit `observe_duration` call, but the explicit call was already done for the Vector Search queries so I sticked to the same implementation.

QuerthDP:
Changelog:
- moved refreshing index metrics to a separate function
- added fts metrics integration test
- changed metrics description to not expose the Tantivy backend internals
- added `get_fts_index`

crates/vector-store/src/engine.rs
```diff
@@ -60,13 +62,18 @@ pub(crate) enum Engine {
         key: IndexKey,
         tx: oneshot::Sender<GetVsIndexR>,
     },
+    GetFtsIndex {
+        key: IndexKey,
+        tx: oneshot::Sender<GetFtsIndexR>,
+    },
```
QuerthDP, line 68: This should be added in a separate commit. I'll fix that

QuerthDP:
Changelog:
- split `get_fts_index` addition into separate commit

ewienik (approved):

PR #524, opened 2026-07-16
vector-store: distinguish 503 from ann/bm25 when node is bootstrapping

knowack1:
When an ANN or BM25 request is made while the index is not yet serving, the 503 response body now contains a JSON object with a 'reason' field that lets clients differentiate between two distinct cases without making a separate /api/v1/status call (which would introduce a TOCTOU race condition):

  {"reason":"NODE_BOOTSTRAPPING"}
    The whole node has not finished its startup sequence and is not
    ready to serve any traffic. Clients should treat this as a
    node-level outage and trigger high-availability failover.

  {"reason":"INDEX_BUILDING","progress":P}
    The node is healthy (SERVING) but this specific index is still
    being constructed after creation. Clients should surface this
    error to the caller rather than failing over.

Reference: SCYLLADB-3209

### DIFF
--- a/crates/vector-store/src/httproutes.rs
+++ b/crates/vector-store/src/httproutes.rs
@@ -107,7 +107,11 @@ use utoipa_swagger_ui::SwaggerUi;
     components(
         schemas(
             httpapi::KeyspaceName,
-            httpapi::IndexName
+            httpapi::IndexName,
+            httpapi::IndexNotReadyReason
+        ),
+        responses(
+            httpapi::IndexNotReadyResponse
         )
     ),
 )]
@@ -544,9 +548,7 @@ If TLS is enabled on the server, clients must connect using a HTTPS protocol.",
         ),
         (
             status = 503,
-            description = "Service Unavailable. Indicates that a full scan of the index is in progress and the search cannot be performed at this time.",
-            content_type = "application/json",
-            body = ErrorMessage
+            response = httpapi::IndexNotReadyResponse
         )
     )
 )]
@@ -574,12 +576,14 @@ async fn post_index_ann(
         let index_key = IndexKey::new(&keyspace, &index_name);
         let (equality_cols, range_cols) = restriction_columns(&request.filter);
         let allow_filtering = request.filter.as_ref().is_some_and(|f| f.allow_filtering);
-        let (routed_key, index, primary_key_columns, filtering_columns, table_columns) = match state
-            .indexes
-            .read()
-            .unwrap()
-            .best_index(&index_key, &equality_cols, &range_cols)
-        {
+        let best_index_state =
+            state
+                .indexes
+                .read()
+                .unwrap()
+                .best_index(&index_key, &equality_cols, &range_cols);
+        let (routed_key, index, primary_key_columns, filtering_columns, table_columns) =
+            match best_index_state {
             indexes::BestIndexState::Serving {
                 key: routed_key,
                 index,
@@ -620,13 +624,18 @@ async fn post_index_ann(
 
                 match progress {
                     Progress::InProgress(percentage) => {
-                        let msg = format!(
-                            "Index {keyspace}.{index_name} is not available yet \
-                            as it is still being constructed, progress: {:.3}%",
-                            percentage.get()
+                        let reason = index_not_ready_reason(
+                            &state.node_state,
+                            &keyspace,
+                            &index_name,
+                            percentage,
+                        )
+                        .await;
+                        debug!(
+                            "post_index_ann: index {keyspace}.{index_name} not ready: {reason:?}"
                         );
-                        debug!("post_index_ann: {msg}");
-                        return (StatusCode::SERVICE_UNAVAILABLE, msg).into_response();
+                        return (StatusCode::SERVICE_UNAVAILABLE, response::Json(reason))
+                            .into_response();
                     }
                     Progress::Done => {
                         let msg = format!(
@@ -783,9 +792,7 @@ If TLS is enabled on the server, clients must connect using a HTTPS protocol.",
         ),
         (
             status = 503,
-            description = "Service Unavailable. Indicates that a full scan of the index is in progress and the search cannot be performed at this time.",
-            content_type = "application/json",
-            body = ErrorMessage
+            response = httpapi::IndexNotReadyResponse
         )
     )
 )]
@@ -809,7 +816,7 @@ async fn post_index_bm25(
 
     let index_key = IndexKey::new(&keyspace, &index_name);
 
-    let (fts_sender, primary_key_columns) = {
+    let serving_or_progress = {
         let indexes = state.indexes.read().unwrap();
         let Some(entry) = indexes.get_fts(&index_key) else {
             timer.observe_duration();
@@ -818,30 +825,31 @@ async fn post_index_bm25(
             debug!("post_index_bm25: {msg}");
             return (StatusCode::NOT_FOUND, msg).into_response();
         };
-        if entry.status() != crate::node_state::IndexStatus::Serving {
-            match entry.progress() {
-                Progress::InProgress(percentage) => {
-                    timer.observe_duration();
-
-                    let msg = format!(
-                        "Index {keyspace}.{index_name} is not available yet as it is still being constructed, progress: {:.3}%",
-                        percentage.get()
-                    );
-                    debug!("post_index_bm25: {msg}");
-                    return (StatusCode::SERVICE_UNAVAILABLE, msg).into_response();
-                }
-                Progress::Done => {
-                    timer.observe_duration();
-
-                    let msg = format!(
-                        "Index {keyspace}.{index_name} is not serving, but full scan did finish."
-                    );
-                    debug!("post_index_bm25: {msg}");
-                    return (StatusCode::INTERNAL_SERVER_ERROR, msg).into_response();
-                }
-            }
+        if entry.status() == crate::node_state::IndexStatus::Serving {
+            Ok((entry.index().clone(), entry.primary_key_columns().clone()))
+        } else {
+            Err(entry.progress())
         }
-        (entry.index().clone(), entry.primary_key_columns().clone())
+    };
+
+    let (fts_sender, primary_key_columns) = match serving_or_progress {
+        Ok(serving) => serving,
+        Err(Progress::InProgress(percentage)) => {
+            timer.observe_duration();
+
+            let reason =
+                index_not_ready_reason(&state.node_state, &keyspace, &index_name, percentage).await;
+            debug!("post_index_bm25: index {keyspace}.{index_name} not ready: {reason:?}");
+            return (StatusCode::SERVICE_UNAVAILABLE, response::Json(reason)).into_response();
+        }
+        Err(Progress::Done) => {
+            timer.observe_duration();
+
+            let msg =
+                format!("Index {keyspace}.{index_name} is not serving, but full scan did finish.");
+            debug!("post_index_bm25: {msg}");
+            return (StatusCode::INTERNAL_SERVER_ERROR, msg).into_response();
+        }
     };
 
     let search_result = fts_sender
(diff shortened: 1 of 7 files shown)


crates/vector-store/src/httproutes.rs
```diff
@@ -743,9 +747,12 @@ If TLS is enabled on the server, clients must connect using a HTTPS protocol.",
...
         (
             status = 503,
-            description = "Service Unavailable. Indicates that a full scan of the index is in progress and the search cannot be performed at this time.",
+            description = "Service Unavailable. The index is not ready to serve requests. \
+The body is a JSON object with a 'reason' field: \
+'NODE_BOOTSTRAPPING' when the node has not yet finished its startup sequence; \
+'INDEX_BUILDING' (with a 'progress' field) when the node is healthy but this index is still being constructed.",
             content_type = "application/json",
-            body = ErrorMessage
+            body = httpapi::IndexNotReadyReason
```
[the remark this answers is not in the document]
knowack1: Fixed

crates/vector-store/src/httproutes.rs
```diff
@@ -763,33 +770,37 @@ async fn post_index_bm25(
...
-                }
-            }
+        (entry.status() != crate::node_state::IndexStatus::Serving).then(|| entry.progress())
+    };
+
+    let (fts_sender, primary_key_columns) = match not_ready_progress {
+        Some(Progress::InProgress(percentage)) => {
+            let reason = index_not_ready_reason(&state.node_state, percentage).await;
+            debug!("post_index_bm25: index {keyspace}.{index_name} not ready: {reason:?}");
+            return (StatusCode::SERVICE_UNAVAILABLE, response::Json(reason)).into_response();
```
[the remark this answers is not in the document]
knowack1: There is not backward transition from Serving

crates/vector-store/src/httproutes.rs
```diff
@@ -776,9 +780,12 @@ If TLS is enabled on the server, clients must connect using a HTTPS protocol.",
         ),
         (
             status = 503,
-            description = "Service Unavailable. Indicates that a full scan of the index is in progress and the search cannot be performed at this time.",
+            description = "Service Unavailable. The index is not ready to serve requests. \
+The body is a JSON object with a 'reason' field: \
+'NODE_BOOTSTRAPPING' when the node has not yet finished its startup sequence; \
+'INDEX_BUILDING' (with a 'progress' field) when the node is healthy but this index is still being constructed.",
```
QuerthDP: Not related to this PR, but genuinely curious should we keep the API definitions duplicated for both ANN/BM25? Couldn't the error codes be defined ones and reused?

crates/vector-store/src/httproutes.rs
```diff
@@ -574,12 +578,13 @@ async fn post_index_ann(
...
-            .unwrap()
-            .best_index(&index_key, &equality_cols, &range_cols)
-        {
+        let best_index_state =
+            state
+                .indexes
+                .read()
+                .unwrap()
+                .best_index(&index_key, &equality_cols, &range_cols);
+        let (routed_key, index, primary_key_columns, table_columns) = match best_index_state {
```
QuerthDP: Just a cosmetic change right?

crates/vector-store/tests/integration/fts.rs
```diff
@@ -348,6 +351,122 @@ async fn fts_bm25_search_not_serving_returns_503() {
...
+        internals,
+        index_factory,
+        receivers,
+        vector_store::new_metrics(),
+    )
+    .await
+    .unwrap();
+    let addr = (*server.address().await.borrow()).unwrap();
+    let client = HttpClient::new(addr);
+    let _keep = (server, senders);
```
QuerthDP: Why is the setup so huge? Don't we have any common utility for that?

crates/vector-store/tests/integration/usearch.rs
```diff
@@ -554,9 +555,124 @@ async fn ann_fail_while_building() {
...
+        internals,
+        index_factory,
+        receivers,
+        vector_store::new_metrics(),
+    )
+    .await
+    .unwrap();
+    let addr = (*server.address().await.borrow()).unwrap();
+    let client = HttpClient::new(addr);
+    let _keep = (server, senders);
```
QuerthDP: ditto

crates/vector-store/src/httproutes.rs
```diff
@@ -811,30 +818,36 @@ async fn post_index_bm25(
...
+        }
+        None => {
+            let indexes = state.indexes.read().unwrap();
+            let Some(entry) = indexes.get_fts(&index_key) else {
+                timer.observe_duration();
+
+                let msg = format!("missing index: {keyspace}.{index_name}");
+                debug!("post_index_bm25: {msg}");
+                return (StatusCode::NOT_FOUND, msg).into_response();
+            };
```
QuerthDP: It is duplicated above. Why do we need this 2nd time?

QuerthDP (commented):
I really like this change. Feels helpful and smart. Leaving some comments before I approve.

crates/vector-store/src/httproutes.rs
```diff
@@ -802,7 +809,7 @@ async fn post_index_bm25(
 
     let index_key = IndexKey::new(&keyspace, &index_name);
 
-    let (fts_sender, primary_key_columns) = {
+    let not_ready_progress = {
```
QuerthDP: One more thing, I think the `not_ready_progress` is a bit counterintuitive. I see from the code what we want to achieve and that `None` means we're all good, but it's not easy to think about what exactly "None not ready progress" or "Some not ready progress" means. I feel like the if was slightly better in this case.

crates/vector-store/src/httproutes.rs
knowack1: Not purely cosmetic. Extracting `best_index(...)` into the named `best_index_state`  drops the `RwLock` read guard at the end of the `let` statement, before the `match`. The match arms now contain `.await` points (`index_not_ready_reason(...).await`), and a `std` `RwLockReadGuard` is not `Send`, so holding it across an `.await` would not compile.

crates/vector-store/tests/integration/fts.rs
```diff
@@ -304,28 +312,70 @@ async fn fts_bm25_search_not_found_returns_404() {
 }
 
 #[tokio::test]
-async fn fts_bm25_search_not_serving_returns_503() {
+async fn fts_bm25_search_returns_503_and_node_bootstrapping_reason_when_index_is_building_during_node_bootstrapping()
+ {
```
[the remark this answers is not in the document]
knowack1: fmt passes

crates/vector-store/tests/integration/fts.rs
```diff
@@ -304,28 +312,70 @@ async fn fts_bm25_search_not_found_returns_404() {
...
+        .await;
+
+    assert_eq!(response.status(), StatusCode::SERVICE_UNAVAILABLE);
+    let reason: IndexNotReadyReason = response.json().await.unwrap();
+    assert_eq!(reason, IndexNotReadyReason::NodeBootstrapping);
+}
+
+#[tokio::test]
+async fn fts_bm25_search_returns_503_and_index_building_reason_when_index_is_building_while_node_is_serving()
+ {
```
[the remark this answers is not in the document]
knowack1: fmt passes

crates/vector-store/src/httproutes.rs
knowack1: I put this into IndexNotReadyResponse. So now there is no duplicate. Unfortunately I put this into same commit. I hope now its not too much in the single commit.

crates/vector-store/tests/integration/fts.rs
knowack1: I've made some improvements, so the setup is now a bit more compact. Instead of doing all these initialization steps manually, I use a generic test utility that creates the table/index and waits until it's ready. After that, I create a second index for bootstrapping and run queries against it. This makes the test setup a little smaller without significantly changing or refactoring the setup utilities.

crates/vector-store/src/httproutes.rs
knowack1: Fixed.

crates/vector-store/tests/integration/usearch.rs
knowack1: Just like with FTS, I need to create a new column for the second index here. Otherwise, the best index routing will always pick up the serving index.

crates/vector-store/src/httproutes.rs
knowack1: Fixed. I renamed this serving_or_progress and removed the None.

knowack1:
Changelog:
- 503 response now contains 'message' instead of 'progress'. The message will be forwarded to the CQL client in case of INDEX_BUILDING, keeping the exact behavior as it was before this change.

crates/vector-store/src/httproutes.rs
```diff
@@ -280,6 +284,18 @@ async fn get_indexes(State(state): State<RoutesInnerState>) -> Response {
...
 struct ErrorMessage(#[allow(dead_code)] String);
 
+/// The 503 response shared by the ANN and BM25 search endpoints when the index
+/// is not ready to serve requests.
+#[derive(utoipa::ToResponse)]
+#[response(
+    description = "Service Unavailable. The index is not ready to serve requests. \
+The body is a JSON object with a 'reason' field: \
+'NODE_BOOTSTRAPPING' when the node has not yet finished its startup sequence; \
+'INDEX_BUILDING' when the node is healthy but this index is still being constructed.",
```
QuerthDP: Should we mention the "message" field as well?

crates/vector-store/src/httproutes.rs
```diff
@@ -280,6 +284,18 @@ async fn get_indexes(State(state): State<RoutesInnerState>) -> Response {
...
+    content_type = "application/json"
+)]
+struct IndexNotReadyResponse(#[allow(dead_code)] httpapi::IndexNotReadyReason);
```
QuerthDP: Why `#[allow(dead_code)]`?

crates/vector-store/tests/integration/fts.rs
```diff
@@ -348,13 +350,73 @@ async fn fts_bm25_search_not_serving_returns_503() {
...
+    crate::enable_tracing();
+
+    let (client, _keyspace_name, _index_name, db, _hold) =
+        setup_fts_and_wait([(vec![CqlValue::Int(1)], "hello world", 10)], 1).await;
```
QuerthDP, line 363: Maybe we should use non-FTS index so that adding routing does not break the test?

QuerthDP: Or at least on a different column?

QuerthDP (dismissed):
Looks really good. Thanks for adressing my comments

crates/vector-store/src/httproutes.rs
```diff
@@ -280,6 +284,18 @@ async fn get_indexes(State(state): State<RoutesInnerState>) -> Response {
...
+/// is not ready to serve requests.
+#[derive(utoipa::ToResponse)]
+#[response(
+    description = "Service Unavailable. The index is not ready to serve requests. \
+The body is a JSON object with a 'reason' field: \
+'NODE_BOOTSTRAPPING' when the node has not yet finished its startup sequence; \
+'INDEX_BUILDING' (with a 'progress' field) when the node is healthy but this index is still being constructed.",
+    content_type = "application/json"
+)]
+struct IndexNotReadyResponse(#[allow(dead_code)] httpapi::IndexNotReadyReason);
```
ewienik: Consider moving it inito httpapi crate

ewienik (changes requested):

crates/vector-store/src/httproutes.rs
knowack1: I can mention if you think this is essential.  I think it is not mandatory since it is specified in the openapi.json.

crates/vector-store/src/httproutes.rs
knowack1: Fixed. Moved into `httpapi` crate.

crates/vector-store/src/httproutes.rs
knowack1: Without this the code does not compile.  Not sure why. Maybe because this type is used for schema doc only and this is some utoipa issue. I took this pattern from `ErrorMessage` that has the same case.

crates/vector-store/tests/integration/fts.rs
knowack1: This is fts specific test, so I think fts index and bm25 query is essential here. I think adding this extra column now will be confusing. Also most likely more tests will require adjustments after adding routing.

QuerthDP (dismissed):

ewienik (dismissed):

QuerthDP:
@knowack1 please resolve conflicts

ewienik (approved):

PR #525, opened 2026-07-16
vector-store: create an actual index and materialize index files

Akvear:
This is a follow-up to #504 
It's a work in progress rn


Akvear:
closed due to approach change

PR #526, opened 2026-07-20
vector-store: create an in memory diskann provider

Akvear:
In this PR a DiskANNIndex is created in a in-memory `DefaultProvider` mode using the parameters mapped from `VsIndexConfiguration`

Fixes: VECTOR-717

### DIFF
--- a/crates/vector-store/Cargo.toml
+++ b/crates/vector-store/Cargo.toml
@@ -41,6 +41,8 @@ console-subscriber = {workspace = true, optional = true}
 dashmap.workspace = true
 derive_more.workspace = true
 diskann.workspace = true
+diskann-providers.workspace = true
+diskann-vector.workspace = true
 dotenvy.workspace = true
 futures.workspace = true
 hotpath.workspace = true
--- a/crates/vector-store/src/vs_index/diskann.rs
+++ b/crates/vector-store/src/vs_index/diskann.rs
@@ -4,12 +4,30 @@
  */
 
 use crate::Config;
+use crate::Dimensions;
+use crate::DiskannAlpha;
+use crate::SpaceType;
 use crate::VsIndexFactory;
 use crate::memory::Memory;
 use crate::perf;
 use crate::table::Table;
 use crate::vs_index::actor::VsIndex;
 use crate::vs_index::factory::VsIndexConfiguration;
+use anyhow::Context;
+use diskann::graph::Config as DiskannConfig;
+use diskann::graph::DiskANNIndex;
+use diskann::graph::config::Builder;
+use diskann::graph::config::MaxDegree;
+use diskann::graph::config::defaults::ALPHA as DISKANN_DEFAULT_ALPHA;
+use diskann_providers::model::graph::provider::async_::FastMemoryVectorProviderAsync;
+use diskann_providers::model::graph::provider::async_::TableDeleteProviderAsync;
+use diskann_providers::model::graph::provider::async_::common::NoStore;
+use diskann_providers::model::graph::provider::async_::common::TableBasedDeletes;
+use diskann_providers::model::graph::provider::async_::inmem::CreateFullPrecision;
+use diskann_providers::model::graph::provider::async_::inmem::DefaultProvider;
+use diskann_providers::model::graph::provider::async_::inmem::DefaultProviderParameters;
+use diskann_vector::distance::Metric;
+use std::num::NonZeroUsize;
 use std::sync::Arc;
 use std::sync::RwLock;
 use tokio::sync::mpsc;
@@ -19,7 +37,14 @@ use tracing::debug;
 use tracing::debug_span;
 use tracing::warn;
 
-pub struct DiskannIndexFactory;
+const MAX_POINTS: NonZeroUsize = NonZeroUsize::new(1_000_000).unwrap();
+
+type DiskannProvider =
+    DefaultProvider<FastMemoryVectorProviderAsync<f32>, NoStore, TableDeleteProviderAsync>;
+
+pub struct DiskannIndexFactory {
+    alpha: DiskannAlpha,
+}
 
 impl VsIndexFactory for DiskannIndexFactory {
     fn create_index(
@@ -28,7 +53,25 @@ impl VsIndexFactory for DiskannIndexFactory {
         _table: Arc<RwLock<Table>>,
         _memory: mpsc::Sender<Memory>,
     ) -> anyhow::Result<mpsc::Sender<VsIndex>> {
-        new(index.key)
+        let params = DiskannParams::new(&index, self.alpha, MAX_POINTS)?;
+        let provider_params = DefaultProviderParameters::simple(
+            usize::from(params.max_points),
+            usize::from(params.dim.0),
+            params.metric,
+            u32::from(params.config.max_degree_u32()),
+        );
+
+        let provider: DiskannProvider = DefaultProvider::new_empty(
+            provider_params,
+            CreateFullPrecision::<f32>::new(usize::from(params.dim.0), None),
+            NoStore,
+            TableBasedDeletes,
+        )
+        .context("failed to create DiskANN provider")?;
+
+        let diskann_index = DiskANNIndex::new(params.config, provider, None);
+
+        new(index.key, diskann_index)
     }
 
     fn index_engine_version(&self) -> String {
@@ -37,12 +80,21 @@ impl VsIndexFactory for DiskannIndexFactory {
 }
 
 pub fn new_diskann(
-    _config_rx: watch::Receiver<Arc<Config>>,
+    mut config_rx: watch::Receiver<Arc<Config>>,
 ) -> anyhow::Result<DiskannIndexFactory> {
-    Ok(DiskannIndexFactory)
+    let config = config_rx.borrow_and_update();
+
+    Ok(DiskannIndexFactory {
+        alpha: config
+            .diskann_alpha
+            .unwrap_or(DiskannAlpha::new(DISKANN_DEFAULT_ALPHA).unwrap()),
+    })
 }
 
-fn new(index_key: crate::IndexKey) -> anyhow::Result<mpsc::Sender<VsIndex>> {
+fn new(
+    index_key: crate::IndexKey,
+    index: DiskANNIndex<DiskannProvider>,
+) -> anyhow::Result<mpsc::Sender<VsIndex>> {
     let (tx, mut rx) = mpsc::channel(perf::channel_size().into());
 
     tokio::spawn(perf::hotpath_async(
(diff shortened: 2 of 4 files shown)


crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -0,0 +1,437 @@
...
+use crate::PartitionId;
+use crate::SpaceType;
+use crate::Vector;
+use crate::VsIndexFactory;
+use crate::memory::Memory;
+use crate::perf;
+use crate::table::PrimaryId;
+use crate::table::Table;
+use crate::table::TableSearch;
+use crate::vs_index::actor::AnnR;
```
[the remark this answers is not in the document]
Akvear: not true

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -76,3 +134,129 @@ fn new(index_key: crate::IndexKey) -> anyhow::Result<mpsc::Sender<VsIndex>> {
...
+mod tests {
+    use super::*;
+    use crate::Connectivity;
+    use crate::ExpansionAdd;
+    use crate::ExpansionSearch;
+    use crate::IndexKey;
+    use crate::IndexName;
+    use crate::KeyspaceName;
+    use crate::Quantization;
+    use diskann::graph::config::defaults::ALPHA as DISKANN_DEFAULT_ALPHA;
```
knowack1: You already have this imported in production code - no need to repeat here.

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -76,3 +134,129 @@ fn new(index_key: crate::IndexKey) -> anyhow::Result<mpsc::Sender<VsIndex>> {
...
+    use std::num::NonZeroUsize;
```
knowack1: same here

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -76,3 +134,129 @@ fn new(index_key: crate::IndexKey) -> anyhow::Result<mpsc::Sender<VsIndex>> {
...
+
+    const MAX_POINTS: NonZeroUsize = NonZeroUsize::new(1_000_000).unwrap();
```
knowack1: Already defined in production code.

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -76,3 +134,129 @@ fn new(index_key: crate::IndexKey) -> anyhow::Result<mpsc::Sender<VsIndex>> {
...
+
+    fn test_index() -> VsIndexConfiguration {
```
knowack1: Not sure why this is extracted to function as is used in a single place. Also makes the diskann_params_try_from_index_configuration harder to read as numbers to compare in assertions are in this function . Consider putting this inline to diskann_params_try_from_index_configuration.

I you want to keep that way consider renaming to `test_index_config`, as this is actually index config not index itself.

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -76,3 +134,129 @@ fn new(index_key: crate::IndexKey) -> anyhow::Result<mpsc::Sender<VsIndex>> {
...
+        Ok(Self {
+            config,
+            metric,
+            dim: cfg.dimensions,
+            max_points,
+        })
+    }
+}
+
+impl From<DiskannParams> for IndexConfiguration {
```
knowack1: I this used anywhere?

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -28,6 +63,23 @@ impl VsIndexFactory for DiskannIndexFactory {
...
+        );
+
+        let provider: DiskannProvider = DefaultProvider::new_empty(
+            provider_params,
+            CreateFullPrecision::<f32>::new(usize::from(params.dim.0), None),
+            NoStore,
+            TableBasedDeletes,
+        )?;
+
+        let _diskann_index = DiskANNIndex::new(params.config, provider, None);
```
knowack1: This index will be dropped at the end of this function. Even if it's not utilized yet, I would keep it alive, as this object is the essence of this implementation.

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -76,3 +87,132 @@ fn new(index_key: crate::IndexKey) -> anyhow::Result<mpsc::Sender<VsIndex>> {
...
+#[derive(Clone)]
+struct DiskannParams {
+    config: DiskannConfig,
+    metric: Metric,
+    dim: Dimensions,
+    max_points: NonZeroUsize,
+}
+
+impl DiskannParams {
+    #[allow(dead_code)]
```
knowack1: I think the order of the introduced changes is wrong. It doesn't make sense to create commit with `dead_code` markers and comments that are just removed in the next commit - it creates unnecessary noise.

Imagine first creating the index with a fixed configuration, and then adding the configuration mapping in the following commit.

That said, you don't need to fix this now; it's just something for future consideration.

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -39,6 +63,23 @@ impl VsIndexFactory for DiskannIndexFactory {
...
+            params.metric,
+            u32::from(params.config.max_degree_u32()),
+        );
+
+        let provider: DiskannProvider = DefaultProvider::new_empty(
+            provider_params,
+            CreateFullPrecision::<f32>::new(usize::from(params.dim.0), None),
+            NoStore,
+            TableBasedDeletes,
+        )?;
```
knowack1: nit; consider adding context here to keep the pattern already started in:
```
        let config = builder
            .build()
            .context("failed to build DiskANN configuration")?;
````

File: crates/vector-store/src/vs_index/diskann.rs (hunk 6 above)
Comment (Akvear) [reply crates/vector-store/src/vs_index/diskann.rs]: No, it must have been a remanent after the on-disk index creation attempts.

File: crates/vector-store/src/vs_index/diskann.rs (hunk 7 above)
Comment (Akvear) [reply crates/vector-store/src/vs_index/diskann.rs]: How do you propose we do that, while keeping clippy happy?

**Comment by Akvear** [comment]
Changes:
- Changed the order of the last 2 commits
- Removed test imports already in production
- Removed test helper, since it's only used in 1 place
- Removed unused TryFrom impl

File: crates/vector-store/src/vs_index/diskann.rs (hunk 7 above)
Comment (knowack1) [reply crates/vector-store/src/vs_index/diskann.rs]: I do not know the exact challenge, but at worst, clippy can be suppressed.

File: crates/vector-store/Cargo.toml
Comment (QuerthDP) [inline crates/vector-store/Cargo.toml:1]: Could you add brief explanation what are those crates needed for?

File: crates/vector-store/src/vs_index/diskann.rs
Comment (QuerthDP) [inline crates/vector-store/src/vs_index/diskann.rs:1]: > - DefaultContext - DefaultProvider docs state that this should be used in most cases

Why?

File: crates/vector-store/src/vs_index/diskann.rs [hunk 10]
```diff
@@ -19,6 +32,16 @@ use tracing::debug;
 use tracing::debug_span;
 use tracing::warn;
 
+/// Concrete in-memory DiskANN provider used for the PoC: full-precision `f32`
```
QuerthDP: I would not mention it's used only for PoC. We can introduce this now, and remove later, no need to argument that it's for PoC only. We don't want this comment in the future if this gets out of PoC.

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -19,6 +32,16 @@ use tracing::debug;
...
+/// vectors, no auxiliary store, deletes enabled, default context.
+/// Aliased to avoid repeating the full type at every use site.
```
QuerthDP: Actually, the whole comment is not needed imo

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -28,6 +51,22 @@ impl VsIndexFactory for DiskannIndexFactory {
         _table: Arc<RwLock<Table>>,
         _memory: mpsc::Sender<Memory>,
     ) -> anyhow::Result<mpsc::Sender<VsIndex>> {
+        let provider_params = DefaultProviderParameters::simple(1_000, 3, Metric::L2, 20);
```
QuerthDP: This needs a comment what those magic numbers are.

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -28,6 +51,22 @@ impl VsIndexFactory for DiskannIndexFactory {
...
+
+        let provider: DiskannProvider = DefaultProvider::new_empty(
+            provider_params,
+            CreateFullPrecision::<f32>::new(3, None),
+            NoStore,
+            TableBasedDeletes,
+        )
+        .context("failed to create DiskANN provider")?;
+
+        let config = Builder::new(3, MaxDegree::default_slack(), 3, Metric::L2.into())
```
QuerthDP: Is 3 a default dimension or what is that? Why is it statically set here? Would we support only vector of dim 3?

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -51,21 +61,23 @@ impl VsIndexFactory for DiskannIndexFactory {
...
             NoStore,
             TableBasedDeletes,
         )
         .context("failed to create DiskANN provider")?;
 
-        let config = Builder::new(3, MaxDegree::default_slack(), 3, Metric::L2.into())
-            .build()
-            .context("Failed to build DiskANN configuration")?;
-
-        let _diskann_index = DiskANNIndex::new(config, provider, None);
```
QuerthDP: Ok I see it now. Why did we introduce this stub at all? I think it's not needed and confusing a lot.

QuerthDP (changes requested):

Akvear:
Changes:
- joined the 2 troublesome commits
- added explanation to why crates are neede

crates/vector-store/src/vs_index/diskann.rs
Akvear: It specifies this - https://docs.rs/diskann/0.55.0/diskann/provider/trait.ExecutionContext.html, should this be linked in the commit message?

crates/vector-store/src/vs_index/diskann.rs
QuerthDP: Wouldn't removing the underscore be exact the same and the index will be dropped at the end anyway?
I would leave it as is than to omit the clippy explicitly.
We will remove the underscore when we'll use the index.

QuerthDP (dismissed):

ewienik (dismissed):

crates/vector-store/src/vs_index/diskann.rs
ewienik: You can pass this into new function, then before end of scope of the actor task in new use `drop(index)`

Akvear:
Changes: 
- moved the index into the async block in the `new()` function

QuerthDP (approved):

ewienik (approved):

PR #537, opened 2026-07-29
scripts: pre-create cargo cache dirs before container bind-mount

Annamikhlin:
On a fresh GitHub Actions runner, ~/.cargo/registry already exists but ~/.cargo/git does not, since it is only created on the first git-dependency fetch. Docker auto-creates a missing bind-mount source path as root, so the build container (running as the non-root host user) gets Permission denied when cargo tries to write into it.

This caused the build-docker job to fail in the release.yaml run for 1.9.1 https://github.com/scylladb/vector-store/actions/runs/30000488118/job/89184255041

Ref: https://scylladb.atlassian.net/browse/RELENG-715

### DIFF
--- a/scripts/run-with-release-toolchain
+++ b/scripts/run-with-release-toolchain
@@ -15,7 +15,10 @@ user_cargo_home=$CARGO_HOME
 
 docker_cargo_home=$(docker run --rm $image sh -c 'echo $CARGO_HOME')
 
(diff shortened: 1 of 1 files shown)


yaronkaikov (approved):
LGTM

ewienik (approved):

PR #538, opened 2026-07-29
.github/workflows/release.yaml: variable should be VECTOR_

amnonh:


swasik (changes requested):
Could you add commit and PR description?

swasik:
Duplicate of #539

PR #536, opened 2026-07-28
vector-store: refactor passing memory & worker actors

ewienik:
We need to have single instance of worker actor - all indexes ought to use it. To be able to provide single instance of worker we need to refactor run function to not provide index factory function.

Additionally this PR adds *.log files to gitignore list.

Fixes: VECTOR-778

PR #540, opened 2026-07-30
vector-store: using new inmem provider in diskann

Akvear:
This PR is an alternative to #535 with the newly merged Diskann In Mem 2.0 provider.
It is more convenient to use because of arbitrary ids but it is still experimental and likely to require fixes from the DiskANN team. I think it's better to try and incorporate it rather than relying on the old provider. 

Fixes: VECTOR-714

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -903,15 +903,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "bincode"
-version = "1.3.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b1f45e9417d87227c7a56d22e471c6206462cba514c7590c09aff4cf6d1ddcad"
-dependencies = [
- "serde",
-]
-
-[[package]]
 name = "bitflags"
 version = "2.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1828,51 +1819,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "diskann-providers"
-version = "0.55.0"
-source = "git+https://github.com/microsoft/DiskANN.git?rev=90f49467a7558bd53e01f61617c294611a96158d#90f49467a7558bd53e01f61617c294611a96158d"
-dependencies = [
- "arc-swap",
- "bincode",
- "bytemuck",
- "byteorder",
- "cfg-if",
- "diskann",
- "diskann-quantization",
- "diskann-utils",
- "diskann-vector",
- "diskann-wide",
- "futures-util",
- "half",
- "hashbrown 0.16.0",
- "num-traits",
- "prost",
- "rand 0.9.4",
- "rand_distr",
- "rayon",
- "serde",
- "thiserror 2.0.18",
- "tokio",
- "tracing",
-]
-
-[[package]]
-name = "diskann-quantization"
-version = "0.55.0"
-source = "git+https://github.com/microsoft/DiskANN.git?rev=90f49467a7558bd53e01f61617c294611a96158d#90f49467a7558bd53e01f61617c294611a96158d"
-dependencies = [
- "bytemuck",
- "cfg-if",
- "diskann-utils",
- "diskann-vector",
- "diskann-wide",
- "half",
- "rand 0.9.4",
- "rayon",
- "thiserror 2.0.18",
-]
-
-[[package]]
 name = "diskann-utils"
 version = "0.55.0"
 source = "git+https://github.com/microsoft/DiskANN.git?rev=90f49467a7558bd53e01f61617c294611a96158d#90f49467a7558bd53e01f61617c294611a96158d"
@@ -1884,7 +1830,6 @@ dependencies = [
  "half",
  "rand 0.9.4",
  "rand_distr",
- "rayon",
  "thiserror 2.0.18",
 ]
 
@@ -6161,7 +6106,6 @@ dependencies = [
  "derive_more",
  "diskann",
  "diskann-inmem",
- "diskann-providers",
  "diskann-vector",
  "dotenvy",
  "futures",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1477,6 +1477,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "crossbeam-queue"
+version = "0.3.13"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "803d13fb3b09d88be9f4dbc29062c66b19bf7170867ceb746d2a8689bf6c7a26"
+dependencies = [
+ "crossbeam-utils",
+]
+
+[[package]]
 name = "crossbeam-utils"
 version = "0.8.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1784,7 +1793,7 @@ dependencies = [
 [[package]]
 name = "diskann"
 version = "0.55.0"
-source = "git+https://github.com/microsoft/DiskANN.git?rev=541472a8d0bbfa4d458ec1b909e1d41a18adaabc#541472a8d0bbfa4d458ec1b909e1d41a18adaabc"
+source = "git+https://github.com/microsoft/DiskANN.git?rev=90f49467a7558bd53e01f61617c294611a96158d#90f49467a7558bd53e01f61617c294611a96158d"
 dependencies = [
  "anyhow",
  "bytemuck",
@@ -1802,9 +1811,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "diskann-inmem"
+version = "0.55.0"
+source = "git+https://github.com/microsoft/DiskANN.git?rev=90f49467a7558bd53e01f61617c294611a96158d#90f49467a7558bd53e01f61617c294611a96158d"
+dependencies = [
+ "bytemuck",
+ "crossbeam-queue",
+ "dashmap 6.1.0",
+ "diskann",
+ "diskann-utils",
+ "diskann-vector",
+ "diskann-wide",
+ "half",
+ "parking_lot",
+ "thiserror 2.0.18",
+]
+
+[[package]]
 name = "diskann-providers"
 version = "0.55.0"
-source = "git+https://github.com/microsoft/DiskANN.git?rev=541472a8d0bbfa4d458ec1b909e1d41a18adaabc#541472a8d0bbfa4d458ec1b909e1d41a18adaabc"
+source = "git+https://github.com/microsoft/DiskANN.git?rev=90f49467a7558bd53e01f61617c294611a96158d#90f49467a7558bd53e01f61617c294611a96158d"
 dependencies = [
  "arc-swap",
  "bincode",
@@ -1833,7 +1859,7 @@ dependencies = [
 [[package]]
 name = "diskann-quantization"
 version = "0.55.0"
-source = "git+https://github.com/microsoft/DiskANN.git?rev=541472a8d0bbfa4d458ec1b909e1d41a18adaabc#541472a8d0bbfa4d458ec1b909e1d41a18adaabc"
+source = "git+https://github.com/microsoft/DiskANN.git?rev=90f49467a7558bd53e01f61617c294611a96158d#90f49467a7558bd53e01f61617c294611a96158d"
 dependencies = [
  "bytemuck",
  "cfg-if",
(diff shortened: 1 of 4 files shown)


crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -90,11 +94,26 @@ fn new(
             async move {
                 debug!("starting");
 
+                let mut state = State::new(index);
```
ewienik: Try to use BTreeMap<PartitionId, State> since we can have different partitions stored in a single index actor

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -90,11 +94,26 @@ fn new(
...
-                        VsIndex::AddVector { .. }
-                        | VsIndex::RemoveVector { .. }
-                        | VsIndex::RemovePartition { .. } => {
+                        VsIndex::AddVector {
+                            primary_id,
+                            embedding,
+                            in_progress: _in_progress,
+                            ..
+                        } => {
+                            state.add_vector(primary_id, embedding).await;
```
ewienik: Use partition_id to get the correct state

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -90,11 +94,26 @@ fn new(
...
+                        }
+                        VsIndex::RemoveVector {
+                            primary_id,
+                            in_progress: _in_progress,
+                            ..
+                        } => {
+                            state.remove_vector(primary_id).await;
```
ewienik: Use partition_id to get the correct state

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -116,9 +133,21 @@ fn new(
...
-                                .send(Err(anyhow::anyhow!("DiskANN index is not implemented yet")));
+                        VsIndex::Ann {
+                            embedding,
+                            limit,
+                            tx,
+                            ..
+                        } => {
+                            if let Some(tx) = validate_dimensions(tx, &embedding, state.params.dim)
+                            {
+                                _ = tx.send(state.ann(embedding, limit).await);
```
ewienik: Convert index_key into partition_id to get the correct state

crates/vector-store/src/table/primary_id.rs
```diff
@@ -15,6 +15,7 @@ use tracing::info;
     Copy,
     Clone,
     Debug,
+    Default,
```
ewienik: We shouldn't provide default PrimaryId -> it needs to be created by try_new to hold invariants

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -171,14 +213,107 @@ impl State {
...
+    ) -> anyhow::Result<impl Iterator<Item = anyhow::Result<(PrimaryId, Distance)>>> {
+        let space_type = self.params.space_type;
+        let dimensions = embedding.dim();
+
+        let l_value = self.params.l_default.get().max(k);
+        let k = k.min(self.params.max_points.get());
+        let params = Knn::new(k, l_value, Some(self.params.beam_width.get()))
+            .context("failed to build DiskANN search parameters")?;
+
+        let mut neighbors = vec![Neighbor::<PrimaryId>::default(); params.k_value().get()];
```
ewienik: We don't have default value for PrimaryId. Is it possible to use preallocated but not filled vector or other storage?

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -171,14 +213,107 @@ impl State {
...
+            };
+            Distance::try_from((raw_distance, space_type, dimensions))
+                .map(|distance| (*neighbor.id(), distance))
+        }))
+    }
+
+    async fn ann(&self, embedding: Vector, limit: Limit) -> AnnR {
+        let partition_id: PartitionId = {
+            let table = self.table.read().unwrap();
+            if let Some((partition_id, _)) = table.partition_id(&self.index_key, None) {
```
ewienik: IndexKey comes with ann message, it shouldn't be stored in the State.

ewienik (changes requested):
When #536 is merged diskann will need to use Worker to spawn work

Akvear:
Changes:
- removed unused diskann-providers
- removed Default from PrimaryId derives
- added partitioning straight away
- added memory allocation subsciber

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -50,28 +70,30 @@ impl VsIndexFactory for DiskannIndexFactory {
...
+        let index_fn = {
+            move || -> anyhow::Result<DiskannIndex> {
+                let params = params.clone();
+                let layer = Full::<f32>::new(usize::from(params.dim.0), params.metric);
+                let cfg = InmemProviderConfig::new(
+                    usize::from(params.max_points),
+                    params.config.max_degree().get(),
+                );
+                let provider = InmemProvider::<_, PrimaryId>::new(layer, cfg, vec![])
+                    .context("failed to create InmemProvider")?;
```
Akvear: The new inmem provider requires at least 1 starting point to work (right now `vec![]` - so not working). What should this be initialized to? @ewienik

ewienik: What is the meaning of start_points? Are these real vectors from the dataset?

Cargo.lock
ewienik, line 1: Can Cargo.lock changes go to the first commit?

Cargo.toml
ewienik, line 1: Can Cargo.toml changes go to the first commit?

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -53,17 +61,25 @@ impl VsIndexFactory for DiskannIndexFactory {
...
+                let diskann_index = DiskANNIndex::new(params.config.clone(), provider, None);
+                Ok(diskann_index)
+            }
```
ewienik: Since this lambda is used as a field in State, consider define this function as a State's method.

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -107,9 +125,24 @@ fn new(
...
                     match msg {
-                        VsIndex::AddVector { .. }
-                        | VsIndex::RemoveVector { .. }
-                        | VsIndex::RemovePartition { .. } => {
+                        VsIndex::AddVector {
+                            partition_id,
+                            primary_id,
+                            embedding,
+                            in_progress: _in_progress,
+                            ..
```
ewienik: nit: not needed anymore

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -107,9 +125,24 @@ fn new(
...
+                        } => {
+                            state.add_vector(partition_id, primary_id, embedding).await;
+                        }
+                        VsIndex::RemoveVector {
+                            partition_id,
+                            primary_id,
+                            in_progress: _in_progress,
+                            ..
```
ewienik: nit: not needed anymore

crates/vector-store/src/vs_index/diskann.rs
Akvear: Copilot:
They are serialized into an internal matrix and loaded into the diskann store as frozen slots. They are the graph traversal entry anchors. Search starts from them and computes their query distances first, but they are not returned to caller as they do not have external ids.

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -133,6 +165,77 @@ fn new(
...
+        }
+    }
+
+    fn get_or_create_partition(&mut self, partition_id: &PartitionId) -> Option<&mut Partition> {
+        if !self.partitions.contains_key(partition_id) {
+            let index = self.index_fn.clone()()
+                .inspect_err(|err| {
+                    error!("failed to create index for partition {partition_id:?}: {err}")
+                })
+                .ok()?;
```
ewienik: Consider return anyhow::Result

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -133,6 +165,77 @@ fn new(
...
+
+    async fn add_vector(
+        &mut self,
+        partition_id: PartitionId,
+        primary_id: PrimaryId,
+        embedding: Vector,
+    ) {
+        if let Err(err) = self
+            .get_or_create_partition(&partition_id)
+            .expect("failed to get or create partition")
```
ewienik: Consider return anyhow::Result

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -133,6 +165,77 @@ fn new(
...
+            .await
+        {
+            warn!("add_vector: failed to insert vector: {err}");
+        }
+    }
+
+    async fn remove_vector(&mut self, partition_id: PartitionId, primary_id: PrimaryId) {
+        if let Err(err) = self
+            .get_or_create_partition(&partition_id)
+            .expect("failed to get or create partition")
```
ewienik: Consider return anyhow::Result

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -177,9 +178,8 @@ fn new(
                                 "DiskANN index does not support filtered search"
                             )));
                         }
-                        VsIndex::Count { tx, .. } => {
-                            _ = tx
-                                .send(Err(anyhow::anyhow!("DiskANN index is not implemented yet")));
+                        VsIndex::Count { index_key, tx, .. } => {
```
ewienik: nit: .. not needed anymore

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -240,9 +242,10 @@ where
         primary_id: PrimaryId,
         embedding: Vector,
     ) {
-        if let Err(err) = self
+        let partition = self
             .get_or_create_partition(&partition_id)
-            .expect("failed to get or create partition")
+            .expect("failed to get or create partition");
```
ewienik: consider return an error

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -347,6 +355,23 @@ where
...
+            };
+            index_id
+        };
+
+        Ok(self
+            .partitions
+            .iter()
+            .filter(|(partition_id, _)| partition_id.index_id() == index_id)
+            .map(|(_, partition)| partition.size)
+            .sum())
```
ewienik: nit: get would be enough?

ewienik (changes requested):

Cargo.toml
Akvear: If we remove the diskann-providers crate in the first commit then diskann.rs wouldn't compile

crates/vector-store/src/vs_index/diskann.rs
ewienik: Diskann creates a new provider for new partition - it is triggered for new add vector - so you will have a single vector with primary_id.
I doubt that we will need to prepare somehow these starting points.

Akvear:
no need for review yet

Akvear:
Changes:
- addressed nits
- changed counter to be per IndexId
- used inplace_delete() instead of data_provider.delete() in remove_vector()
- added remove partition
- moved index creation function to impl State
- used the first incoming vector as the starting point for the index

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -133,6 +148,101 @@ fn new(
...
+        start_point: Option<&[f32]>,
+    ) -> anyhow::Result<&mut Partition> {
+        if !self.partitions.contains_key(partition_id) {
+            let index = self.create_diskann_index(start_point).context(format!(
+                "failed to create index for partition {partition_id:?}",
+            ))?;
+            self.partitions.insert(*partition_id, Partition { index });
+        }
+        self.partitions
+            .get_mut(partition_id)
```
ewienik: nit: consider using Entry from BTreeMap to lookup only once

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -133,6 +148,101 @@ fn new(
...
+        embedding: Vector,
+    ) {
+        let partition =
+            match self.get_or_create_partition(&partition_id, Some(embedding.as_slice())) {
+                Ok(partition) => partition,
+                Err(err) => {
+                    warn!("add_vector failed: {err}");
+                    return;
+                }
+            };
```
ewienik, line 257: nit: if you want you can use `?` and inspect_err for logging warning

Akvear:
rebased on master

ewienik (changes requested):
The PR needs to be rebased with master

Edit (already done :-) )

crates/vector-store/src/vs_index/diskann.rs
Akvear: similarly to usearch, the add_vector fn doesnt have a return type (so no `?`) so I think it's just better and easier to keep it this way.

Akvear:
changes:
- reduced number of lookups in `get_or_create_partition()`
- moved `create_diskann_index()` outside from `impl State` block to avoid issues with borrow checker

ewienik (approved):

PR #541, opened 2026-07-31
benches,benchmark: refactor for testing search and delete

ewienik:
We need performance tests for benchmarking parallel search and delete operations.

This PR refactors benches and introduces new tests: cdc-delete and search-while-deleting.

Additionally this PR adds new benchmark command delete-rows to be able to simulate massive deletion from table.

For faster benchmark testing this PR refactors loading data from parquet file by running reading from disk parallel to the uploading data to scylladb.

Fixes: VECTOR-781

### DIFF
--- a/crates/benchmark/src/data/fbin.rs
+++ b/crates/benchmark/src/data/fbin.rs
@@ -6,7 +6,6 @@
 use crate::data::Query;
 use futures::StreamExt;
 use futures::stream;
-use futures::stream::BoxStream;
 use serde::Deserialize;
 use std::collections::HashSet;
 use std::io::SeekFrom;
@@ -51,9 +50,20 @@ pub(crate) async fn dimension(path: Arc<PathBuf>, config: Arc<Config>) -> usize 
         .dimension as usize
 }
 
-pub(crate) async fn ids_stream(path: Arc<PathBuf>, config: Arc<Config>) -> BoxStream<'static, i64> {
+pub(crate) async fn ids_stream(path: Arc<PathBuf>, config: Arc<Config>) -> mpsc::Receiver<i64> {
     let header = Header::header(&path.join(&config.data_fbin)).await;
-    stream::iter(0..header.count as i64).boxed()
+
+    let workers = Handle::current().metrics().num_workers();
+    const OVERLOAD_FACTOR: usize = 3;
+    let (tx, rx) = mpsc::channel(workers * OVERLOAD_FACTOR);
+    tokio::spawn(async move {
+        for id in 0..header.count as i64 {
+            if tx.send(id).await.is_err() {
+                break;
+            }
+        }
+    });
+    rx
 }
 
 pub(crate) async fn vector_stream(
--- a/crates/benchmark/src/data/parquet.rs
+++ b/crates/benchmark/src/data/parquet.rs
@@ -12,7 +12,6 @@ use arrow_array::types::Int64Type;
 use futures::Stream;
 use futures::StreamExt;
 use futures::stream;
-use futures::stream::BoxStream;
 use itertools::Itertools;
 use parquet::arrow::ProjectionMask;
 use parquet::arrow::async_reader::ParquetRecordBatchStreamBuilder;
@@ -167,8 +166,8 @@ fn extract_embedding(
     ids.zip(embs).collect_vec()
 }
 
-pub(crate) async fn ids_stream(path: Arc<PathBuf>, config: Arc<Config>) -> BoxStream<'static, i64> {
-    train_files(path, Arc::clone(&config))
+pub(crate) async fn ids_stream(path: Arc<PathBuf>, config: Arc<Config>) -> mpsc::Receiver<i64> {
+    let mut stream = train_files(path, Arc::clone(&config))
         .await
         .then({
             let config = Arc::clone(&config);
@@ -202,7 +201,19 @@ pub(crate) async fn ids_stream(path: Arc<PathBuf>, config: Arc<Config>) -> BoxSt
             stream::iter(ids)
         })
         .flatten()
-        .boxed()
+        .boxed();
+
+    let workers = Handle::current().metrics().num_workers();
+    const OVERLOAD_FACTOR: usize = 3;
+    let (tx, rx) = mpsc::channel(workers * OVERLOAD_FACTOR);
+    tokio::spawn(async move {
+        while let Some(id) = stream.next().await {
+            if tx.send(id).await.is_err() {
+                return;
+            }
+        }
+    });
+    rx
 }
 
 pub(crate) async fn vector_stream(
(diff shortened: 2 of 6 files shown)


ewienik:
Changelog for 760246027
- rebase to master

<details>
<summary>range diff</summary>

```
1:  3c0b2affd = 1:  a31cb4dcf benches: setup usearch simulator with 1us time
2:  ae1d96cfe = 2:  27d66c4de benches: increase timestamp as 0 is the initial timestamp in table mod
3:  855352762 = 3:  7115554ee benches: rename add into insert
4:  3791f4e4a = 4:  ccda0ec44 benches: refactor benches background task
5:  50ecb437d = 5:  48073ba1b benches: add cdc-delete and search-while-deleting tests
6:  c8820f34e = 6:  5bfd75734 benchmark: add delete-rows command
7:  01cab0c8a = 7:  760246027 benchmark: refactor reading parquet dataset
```

crates/vector-store/benches/pipeline.rs
```diff
@@ -143,7 +143,12 @@ fn default_index_metadata(
 
 async fn default_config() -> Arc<Config> {
     fn env(key: &str) -> anyhow::Result<String> {
-        Ok(dotenvy::var(key)?)
+        Ok(match key {
+            "VECTOR_STORE_USEARCH_SIMULATOR" => {
+                dotenvy::var(key).unwrap_or("1us:1us:1us".to_string())
```
QuerthDP, line 148: What was the default before? I assume, we did run the original usearch instead of the simulator then?

QuerthDP:
> beches,benchmark: refactor for testing search and delete

typo: beches -> benches

crates/vector-store/benches/pipeline.rs
```diff
@@ -1012,6 +1138,158 @@ fn search_while_inserting(c: &mut Criterion) {
...
+                        {
+                            break;
+                        }
+                    }
+                    // wait until all in-progress markers are dropped
+                    drop(tx_in_progress);
+                    while rx_in_progress.recv().await.is_some() {}
+                });
+                let (bg_task, bg_stop) =
+                    run_search_in_background(db.clone(), client.clone(), bg_concurrency).await;
```
QuerthDP, line 1233: So we're searching twice here? Once in bg and once benchmarking? 
Why couldn't we just run the searches while benchmarking?

crates/benchmark/src/db.rs
```diff
@@ -219,6 +219,50 @@ impl Scylla {
...
+        let mut count = 0;
+        while let Some(vector_id) = rx.recv().await {
+            let permit = Arc::clone(&semaphore).acquire_owned().await.unwrap();
+            let scylla = Arc::clone(&self.0);
+            let st_delete = st_delete.clone();
+
+            count += 1;
+            if count % 1_000_000 == 0 {
+                info!("Deleting vector {}M", count / 1_000_000);
+            }
```
QuerthDP, line 251: Do I understand correctly those are the infos printed per 1M vectors?
So that once "Deleting vector 1M" is printed, we know that 1M vectors were removed, not that 1Mth vector was removed?

crates/benchmark/src/data/parquet.rs
```diff
@@ -220,66 +224,94 @@ pub(crate) async fn vector_stream(
...
+            for row_group_idx in 0..metadata.row_groups().len() {
+                tracing::info!("Processing {row_group_idx} row_group for {path:?}");
                 let config = Arc::clone(&config);
-                async move {
-                    ParquetRecordBatchStreamBuilder::new(File::open(path).await.unwrap())
-                        .await
-                        .unwrap()
+                let tx = tx.clone();
+                let reader_metadata = reader_metadata.clone();
+                let file = file.try_clone().await.unwrap();
```
QuerthDP, line 254: According to Claude this creates shared file offset, which leads to data corruption.
Could you verify that?

QuerthDP (changes requested):
Leaving some concerns. Although the patch looks quite fine.

crates/vector-store/benches/pipeline.rs
ewienik: Previously the simulator settings was done by environment. These benches tests pipeline, not usearch performance, so this commit sets this for default.

crates/vector-store/benches/pipeline.rs
ewienik: Background search is needed to simulate busy worker by other indexes.

crates/benchmark/src/db.rs
ewienik: It means that we are now in the process of deleting 1Mth vector. It also means that 1M vectors were removed. Do you want to improve wording?

crates/benchmark/src/data/parquet.rs
ewienik: It shouldn't lead to data coruption. See documentations: https://arrow.apache.org/rust/parquet/arrow/arrow_reader/struct.ArrowReaderBuilder.html#example-reading-row-groups-in-parallel

crates/vector-store/benches/pipeline.rs
QuerthDP: Meaning that, before the change if environment was not set then the pipeline used usearch instead of the simulator right?

crates/benchmark/src/data/parquet.rs
QuerthDP: I can see they used it really similarly so I think we're safe to assume it's ok.

crates/benchmark/src/db.rs
QuerthDP: It's already the same for Inserting vectors I think, so let it be.

crates/vector-store/benches/pipeline.rs
ewienik: Yes, that's right.

crates/benchmark/src/data/fbin.rs
```diff
@@ -51,9 +50,20 @@ pub(crate) async fn dimension(path: Arc<PathBuf>, config: Arc<Config>) -> usize
...
+    const OVERLOAD_FACTOR: usize = 3;
+    let (tx, rx) = mpsc::channel(workers * OVERLOAD_FACTOR);
+    tokio::spawn(async move {
+        for id in 0..header.count as i64 {
+            if tx.send(id).await.is_err() {
+                break;
+            }
+        }
+    });
+    rx
```
QuerthDP, line 66: One more question: does this change really belong to the add delete-rows command commit?
I cannot understand why was it needed here, and it is not described in the commit message at all.

ewienik: ids_stream changed its definition: it returns now mpsc::Receiver instead of stream. We need to pass this receiver to the delete_rows. ids_stream must be defined for both dataset types.

QuerthDP: Ok, so as I understand it could be the BoxStream before as it was used only to create the buckets (fully consumed) and now we need to read the data continuously so that rx is needed.
Is that correct?

ewienik: Passing stream object is less convenient than passing Receiver. Consuming stream in a for loop with semaphore is less readable IMHO. So for this reason I refactored API.

QuerthDP (approved):

PR #543, opened 2026-08-03
vector-store: diskann integration tests

Akvear:
This PR refactors vs_index integrations tests. Tests are moved from `usearch.rs` to `vs_index.rs` and parametrized by config using the `rstest` crate which replaces the existing `ntest` crate fully.

Fixes: VECTOR-719


crates/vector-store/tests/integration/vs_index.rs
```diff
@@ -204,540 +205,551 @@ pub(crate) async fn setup_store_and_wait_for_index(
...
 
+#[rstest]
+#[case::diskann(diskann_test_config())]
+#[case::usearch(usearch_test_config())]
 #[tokio::test]
-async fn simple_create_search_delete_index() {
+async fn simple_create_search_delete_index(#[case] config: Config) {
     crate::enable_tracing();
 
-    for config in [usearch_test_config(), diskann_test_config()] {
```
Akvear: This is an alternative to just having a loop in the test - it's probably better if we're OK with having a new dependency. The last commit shall be squashed intro the previous or dropped depending on the decision.

ewienik:
Consider moving tests into vs_index in two steps - first commit with single rename usearch.rs into vs_index.rs, and next commit to modify vs_index and add a new usearch.rs file. In this way we can preserve git history of tests functions.

Akvear:
> Consider moving tests into vs_index in two steps - first commit with single rename usearch.rs into vs_index.rs, and next commit to modify vs_index and add a new usearch.rs file. In this way we can preserve git history of tests functions.

If we're ok with using rstest, we can just have a single vs_index (renamed from usearch) file with added test cases, only usearch for filtered ann, and both usearch and diskann for others. WDYT @ewienik

ewienik:
> > Consider moving tests into vs_index in two steps - first commit with single rename usearch.rs into vs_index.rs, and next commit to modify vs_index and add a new usearch.rs file. In this way we can preserve git history of tests functions.
> 
> If we're ok with using rstest, we can just have a single vs_index (renamed from usearch) file with added test cases, only usearch for filtered ann, and both usearch and diskann for others. WDYT @ewienik

I think we can replace ntest with rstest as rstest also has timeout but seems to have better support for fixtures. It could be done with a separate commit.

crates/vector-store/src/fts_index/tantivy.rs
```diff
@@ -615,8 +616,9 @@ mod tests {
         tokens
     }
 
+    #[rstest]
     #[tokio::test]
-    #[ntest::timeout(10_000)]
+    #[timeout(Duration::from_secs(10))]
```
QuerthDP, line 620: It's a bit confusing whether it's rstest::timeout or tokio::timeout (if exists).
I assume it's the first, but I'm not sure by just looking at the code.

crates/vector-store/tests/integration/vs_index.rs
```diff
@@ -54,6 +54,18 @@ pub(crate) fn test_config() -> Config {
     }
 }
 
+fn usearch_test_config() -> Config {
+    test_config()
```
QuerthDP: Is test_config() used anywhere else? If not then maybe we could inline it here, like the diskann_test_config does.

Akvear: Yes, it's used across all integration tests. In the first commit moving from usearch to vs_index requires changing a lot of imports from `use crate::usearch::test_config` to `use crate::vs_index::test_config`.

crates/vector-store/src/fts_index/tantivy.rs
Akvear: Yes, it's from rstest. Unfortunately there is no `rstest::timeout`, using the `rstest` attribute allows us to add an additional `timeout` attribute. I can move it to be before `tokio::test` though.

Akvear:
Changes:
- moved `#[timeout...]` to before `#[tokio::test]`

crates/vector-store/tests/integration/metrics.rs
```diff
@@ -24,7 +23,7 @@ async fn setup_single_vector_index() -> (
     impl Sized,
     Sender<NodeState>,
 ) {
-    usearch::setup_store_and_wait_for_index(
+    crate::vs_index::setup_store_and_wait_for_index(
```
ewienik: nit: `use crate::vs_index`

crates/vector-store/tests/integration/status.rs
```diff
@@ -13,19 +12,20 @@ use vector_store::DbIndexPartitioning;
...
-            ("pk".to_string().into(), NativeType::Int),
-            ("ck".to_string().into(), NativeType::Text),
-        ],
-        None,
-        None,
-        None,
-    )
-    .await;
+    let (_index, client, _db, _server, _node_state) =
+        crate::vs_index::setup_store_and_wait_for_index(
```
ewienik: nit: `use crate::vs_index`

crates/vector-store/tests/integration/status.rs
```diff
@@ -34,7 +34,7 @@ async fn status_is_serving_after_creation() {
 #[tokio::test]
 async fn status_is_bootstrapping_while_discovering_indexes() {
     crate::enable_tracing();
-    let (run, _index, db, _node_state) = usearch::setup_store(
+    let (run, _index, db, _node_state) = crate::vs_index::setup_store(
```
ewienik: nit: `use crate::vs_index`

crates/vector-store/tests/integration/status.rs
```diff
@@ -14,6 +14,7 @@ async fn status_is_serving_after_creation() {
     crate::enable_tracing();
     let (_index, client, _db, _server, _node_state) =
         crate::vs_index::setup_store_and_wait_for_index(
+            crate::vs_index::test_config(),
```
ewienik: nit: `use crate::vs_index`

ewienik (dismissed):

Akvear:
Changes:
- added `use vs_index` instead of calling directly `vs_index::setup_store` etc.

crates/vector-store/tests/integration/metrics.rs
QuerthDP: The use crate::vs_index should be introduced in this commit.

crates/vector-store/tests/integration/status.rs
QuerthDP: Ditto

crates/vector-store/tests/integration/vs_index.rs
QuerthDP: I would like it that the `usearch_test_config` has `test_config` inlined, and `usearch_test_config` being called all the place around instead.

QuerthDP (dismissed):
Looks good. Please consider my comments.

Akvear:
Changes:
- moved changes mentioned above to the proper commit
- renamed test_config() to usearch_test_config() and used that in various places

QuerthDP (approved):

ewienik (approved):

PR #545, opened 2026-08-05
httpapi: vector-store: expose index creation options via API endpoints

QuerthDP:
Today the only way to see the options an index was created with is to
query ScyllaDB directly via CQL (`system_schema.indexes`). This PR
surfaces those options through the HTTP API, using only what Vector Store
already holds in memory - no extra ScyllaDB queries.

- **`GET /indexes`** now includes each index's creation options, alongside its existing keyspace/name/type fields.
- **`GET /indexes/{keyspace}/{index}/status`** is unchanged - it stays
  focused on operational status (state, count, build progress).
- **`GET /indexes/{keyspace}/{index}`** is new - the same
  type and options information as the `/indexes` listing, scoped to one
  index, so a caller who already knows which index they care about
  doesn't have to fetch and filter the whole list.

Fixes: VECTOR-723

### DIFF
--- a/crates/httpapi/src/lib.rs
+++ b/crates/httpapi/src/lib.rs
@@ -56,16 +56,6 @@ pub enum DataType {
     B1,
 }
 
-#[derive(Debug, PartialEq, serde::Deserialize, serde::Serialize, utoipa::ToSchema)]
-#[serde(tag = "type", rename_all = "lowercase")]
-/// Type of index, distinguishing between vector search and fulltext search indexes.
-pub enum IndexType {
-    /// Vector search index with its associated data type.
-    Vector { data_type: DataType },
-    /// Fulltext search index.
-    Fulltext,
-}
-
 #[derive(
     Copy,
     Clone,
@@ -92,12 +82,11 @@ impl Serialize for Distance {
 }
 
 #[derive(Debug, PartialEq, serde::Deserialize, serde::Serialize, utoipa::ToSchema)]
-/// Information about an index, such as keyspace, name and type.
+/// Information about an index, such as keyspace, name, and creation options.
 pub struct IndexInfo {
     pub keyspace: KeyspaceName,
     pub index: IndexName,
-    #[serde(flatten)]
-    pub index_type: IndexType,
+    pub options: IndexOptions,
 }
 
 impl IndexInfo {
@@ -105,9 +94,14 @@ impl IndexInfo {
         IndexInfo {
             keyspace: String::from(keyspace).into(),
             index: String::from(index).into(),
-            index_type: IndexType::Vector {
-                data_type: DataType::F32,
-            },
+            options: IndexOptions::Vector(VectorIndexOptions {
+                dimensions: 3,
+                maximum_node_connections: 16,
+                construction_beam_width: 128,
+                search_beam_width: 64,
+                similarity_function: SimilarityFunction::Euclidean,
+                quantization: DataType::F32,
+            }),
         }
     }
 }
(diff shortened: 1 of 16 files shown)


crates/httpapi/src/lib.rs
```diff
@@ -60,8 +60,8 @@ pub enum DataType {
 #[serde(tag = "type", rename_all = "lowercase")]
 /// Type of index, distinguishing between vector search and fulltext search indexes.
 pub enum IndexType {
-    /// Vector search index with its associated data type.
-    Vector { data_type: DataType },
+    /// Vector search index.
+    Vector,
```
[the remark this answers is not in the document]
QuerthDP: Leaving `data_type` here would add a duplication of quantization option. Maybe we should bump the version 3.0.0?

crates/vector-store/src/httproutes.rs
```diff
@@ -387,6 +457,77 @@ async fn get_index_status(
...
+        IndexInfo {
+            keyspace: keyspace_name.into(),
+            index: index_name.into(),
+            index_type: IndexType::Fulltext,
+            options: IndexOptions::Fulltext(entry.options().into()),
+        }
+    } else {
+        let msg = format!("missing index: {keyspace_name}.{index_name}");
+        debug!("get_index_info: {msg}");
+        return (StatusCode::NOT_FOUND, msg).into_response();
```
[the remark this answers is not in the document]
QuerthDP: It's used the same all over the code.
If it is a real bug, we should fix that in separate PR in all places.

crates/vector-store/src/engine.rs
```diff
@@ -39,15 +39,12 @@ use tracing::debug_span;
 use tracing::info;
 use tracing::trace;
 
-type GetVsIndexKeysR = Vec<(IndexKey, crate::IndexOptionsVs)>;
 type AddIndexR = anyhow::Result<()>;
 type GetVsIndexR = Option<(mpsc::Sender<VsIndex>, mpsc::Sender<DbIndex>)>;
 type GetFtsIndexR = Option<(mpsc::Sender<FtsIndex>, mpsc::Sender<DbIndex>)>;
 
+#[allow(clippy::enum_variant_names)]
```
ewienik, line 46: clippy allowing not needed

crates/vector-store/src/httproutes.rs
```diff
@@ -91,7 +98,7 @@ use utoipa_swagger_ui::SwaggerUi;
             name = "LicenseRef-ScyllaDB-Source-Available-1.0"
         ),
         // version should be updated manually when there are changes in API
-        version = "2.1.0"
+        version = "2.2.0"
```
ewienik: As this is a breaking change it should be 3.0.0

crates/httpapi/src/lib.rs
ewienik: What about putting options directly here? Then we don't need separate type for options.
Then we can rename IndexType into IndexOptions with tag = "type". Then we should get something along the lines of:
```
{
    "keyspace": "my_keyspace",
    "index": "my_vector_index",
    "options": {
        "type": "vector",
        "dimensions": 384,
        "maximum_node_connections": 16,
        "construction_beam_width": 128,
        "search_beam_width": 64,
        "similarity_function": "COSINE",
        "quantization": "F32"
    }
},

```

crates/httpapi/src/lib.rs
```diff
@@ -92,22 +92,29 @@ impl Serialize for Distance {
 }
 
 #[derive(Debug, PartialEq, serde::Deserialize, serde::Serialize, utoipa::ToSchema)]
-/// Information about an index, such as keyspace, name and type.
+/// Information about an index, such as keyspace, name, type and creation options.
 pub struct IndexInfo {
     pub keyspace: KeyspaceName,
     pub index: IndexName,
     #[serde(flatten)]
```
ewienik: If we combine options withing index_type we won't need flattening here.

crates/vector-store/tests/integration/common.rs
```diff
@@ -0,0 +1,121 @@
+/*
+ * Copyright 2026-present ScyllaDB
+ * SPDX-License-Identifier: LicenseRef-ScyllaDB-Source-Available-1.0
```
ewienik: nit: 1.1

crates/httpapi/src/lib.rs
```diff
@@ -92,22 +92,29 @@ impl Serialize for Distance {
...
     pub index_type: IndexType,
+    pub options: IndexOptions,
```
ewienik, line 89: Consider combining IndexType with IndexOptions.
Additionally consider adding IndexStatus a a `"status": {}` object. here also - to get all data for the indexes in one call.

crates/vector-store/src/httproutes.rs
```diff
@@ -456,6 +457,77 @@ async fn get_index_status(
     }
 }
 
+#[utoipa::path(
+    get,
+    path = "/api/v1/indexes/{keyspace}/{index}/info",
```
ewienik: Consider using GET "/api/v1/indexes/{keyspace}/{index}" path

ewienik (changes requested):

crates/vector-store/tests/integration/common.rs
QuerthDP: It's copied from `routing.rs` which is based on 1.0 license. Should I update it?

ewienik: IIUC git history is preserved only when moving files, not copying parts of file. Therefore, this copy seems like file creation, so you can update license.

crates/vector-store/src/engine.rs
QuerthDP: It is indeed needed since all enums share the common suffix `Index`. Without it clippy fails.
We could remove the `Index` suffix, but the names look accurate now and removing the `Index` part would make them less clear.

crates/httpapi/src/lib.rs
QuerthDP: The problem with `status` is that it needs additional call to the `count` endpoint. It's not straight forward how should we behave in case of error. Fail whole request? Omit only those indexes that count failed? Fill it with none? 

Let's leave it for a follow-up.

crates/vector-store/tests/integration/common.rs
QuerthDP: Changed

crates/httpapi/src/lib.rs
QuerthDP: Changed according to the proposal.

crates/vector-store/src/httproutes.rs
QuerthDP: Changed

crates/httpapi/src/lib.rs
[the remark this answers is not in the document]
QuerthDP: Changed to 3.0.0

QuerthDP:
Changelog:
- bump API version to 3.0.0
- bump license version in `common.rs` to 1.1
- include index `type` in `options`
- rename the `GET /indexes/{keyspace}/{index}/info` endpoint to `GET /indexes/{keyspace}/{index}`

ewienik (approved):

crates/vector-store/src/httproutes.rs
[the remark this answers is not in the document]
ewienik: Need separate PR

PR #546, opened 2026-08-06
vector-store: add CQL preferred datacenter/rack config

QuerthDP:
The Rust driver (scylla 1.7.0) added `SessionBuilder::prefer_datacenter` and `prefer_datacenter_and_rack`. Those allow for setting soft preferences that route reads to local nodes first, falling back to remote datacenters/racks if unavailable.

Vector Store is a read-only client of ScyllaDB sharing a single Session per process, so deployments that colocate a Vector Store instance with a specific ScyllaDB datacenter/rack can now reduce cross-rack read traffic by setting `VECTOR_STORE_CQL_PREFERRED_DATACENTER` and `VECTOR_STORE_CQL_PREFERRED_RACK`.

Vector Store has no way to detect its own datacenter/rack placement, so this is opt-in and defaults to unset, leaving existing deployments unaffected. Setting the rack without the datacenter is rejected at config load time, since the driver requires both.

Fixes: VECTOR-725


crates/vector-store/src/db.rs
Lorak-mmk, line 639: This turns out to not be true! It is possible to disable client sharding in scylla. Then, you are connected to Scylla but Sharder is None.

QuerthDP: Thanks, but not related to this PR. Let's consider that in separate issue.

crates/vector-store/src/db.rs
```diff
@@ -480,6 +480,12 @@ fn reconnect_reasons(old_config: &Config, new_config: &Config) -> Vec<&'static s
     if new_config.cql_uri_translation_map != old_config.cql_uri_translation_map {
         reasons.push("CQL URI translation map");
     }
+    if new_config.cql_preferred_datacenter != old_config.cql_preferred_datacenter {
+        reasons.push("CQL preferred datacenter");
+    }
+    if new_config.cql_preferred_rack != old_config.cql_preferred_rack {
+        reasons.push("CQL preferred rack");
```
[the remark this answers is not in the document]
QuerthDP: Added

QuerthDP:
Changelog:
- add tests for reconnecting when dc/rack config changes

crates/vector-store/src/db.rs
QuerthDP: VECTOR-839

swasik:
> Vector Store has no way to detect its own datacenter/rack placement, so this is opt-in and defaults to unset

@QuerthDP Could you also add a task for CLOUD team to make sure the configuration is set correctly?

QuerthDP:
> > Vector Store has no way to detect its own datacenter/rack placement, so this is opt-in and defaults to unset
> 
> @QuerthDP Could you also add a task for CLOUD team to make sure the configuration is set correctly?

[CLOUD-3377](https://scylladb.atlassian.net/browse/CLOUD-3377)

[CLOUD-3377]: https://scylladb.atlassian.net/browse/CLOUD-3377?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

ewienik (approved):

PR #549, opened 2026-08-07
validator: enable authorization by default

QuerthDP:
Enable ScyllaDB authorization (PasswordAuthenticator + CassandraAuthorizer) by default across the validator harness, so the e2e suite exercises the same reduced-privilege setup used in production (Cloud) deployments
instead of full/unauthenticated access.

The default ScyllaDB cluster config now enables auth, and the Vector Store connects using a dedicated role granted only `VECTOR_SEARCH_INDEXING` and `TEXT_SEARCH_INDEXING` (mirroring the fine-grained indexing permissions used in Cloud rather than full `SELECT`). The validator's own test-driver CQL sessions keep connecting as an internal superuser so existing test setup/assertions are unaffected.

Dedicated auth tests already manage their own isolated superuser/role setup and are unaffected, since the default bootstrap only runs when a cluster was started with the shared default auth config.

Fixes: VECTOR-758

### DIFF
--- a/crates/validator/src/common.rs
+++ b/crates/validator/src/common.rs
@@ -29,6 +29,7 @@ use std::collections::HashMap;
 use std::iter;
 use std::net::Ipv4Addr;
 use std::sync::Arc;
+use std::sync::LazyLock;
 use std::sync::atomic::AtomicUsize;
 use std::sync::atomic::Ordering;
 use std::time::Duration;
@@ -36,6 +37,7 @@ use std::time::Instant;
 use tap::Pipe;
 use tokio::time;
 use tracing::info;
+use uuid::Uuid;
 
 pub const DEFAULT_TEST_TIMEOUT: Duration = Duration::from_secs(10 * 60); // 10 minutes
 pub const DEFAULT_OPERATION_TIMEOUT: Duration = Duration::from_secs(20);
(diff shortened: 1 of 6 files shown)


crates/validator/src/common.rs
```diff
@@ -323,9 +400,20 @@ pub async fn init_with_config(
 ) {
     init_dns(actors).await;
 
+    // Bootstrap the default role whenever any VS node is configured to log in as `DEFAULT_DB_USER`.
+    // Callers that bring their own auth setup opt out by connecting VS with a different user.
+    let use_default_auth = vs_configs
+        .iter()
+        .any(|config| config.user.as_deref() == Some(DEFAULT_DB_USER));
+
```
ewienik: Consider adding a parameter `use_default_auth: bool` or adding new `init_with_config_without_auth` or other explicit way to avoid hidden logic to applying default auth.

crates/validator/src/common.rs
```diff
@@ -55,6 +57,75 @@ pub const VS_OCTET_1: u8 = 21;
...
+    LazyLock::new(|| Uuid::new_v4().simple().to_string());
+pub(crate) static SUPERUSER_PASSWORD: LazyLock<String> =
+    LazyLock::new(|| Uuid::new_v4().simple().to_string());
+static SUPERUSER_SALTED_PASSWORD: LazyLock<String> = LazyLock::new(|| {
+    bcrypt::hash(&*SUPERUSER_PASSWORD, bcrypt::DEFAULT_COST)
+        .expect("failed to hash superuser password")
+});
+
+/// Builds the ScyllaDB config YAML with authentication enabled and superuser credentials set.
+pub fn scylla_auth_config() -> Vec<u8> {
```
ewienik, line 79: consider use the same setup for alternator/auth

ewienik (dismissed):

QuerthDP:
Changelog:
- added `use_default_auth` bool param to the `init_with_config`
- deduplicated the SUPERUSER setup from `auth.rs` and `alternator/auth.rs`

ewienik (approved):

PR #550, opened 2026-08-07
packer: keep only journald logging, cap size, add systemd unit stats

d-helios:
## Summary
- Purge `rsyslog` during image build so logs live only in journald instead of being duplicated into `/var/log` files.
- Add a `journald.conf.d` drop-in capping total journal disk usage at 10G (journald rotates/vacuums automatically once the cap is hit).
- Enable node_exporter's systemd collector, scoped to `vector-store` and `systemd-coredump.*` units, so their state is exposed as metrics.

Refs: CLOUDEVOPS-2151

### DIFF
--- a/packer/files/vector_store_install_image
+++ b/packer/files/vector_store_install_image
@@ -103,6 +103,17 @@ def install_service():
         print(e.output)
     run('systemctl daemon-reload')
 
+def configure_journald():
+    conf_dir = '/etc/systemd/journald.conf.d'
+    os.makedirs(conf_dir, exist_ok=True)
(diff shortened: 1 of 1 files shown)


swasik (dismissed):

swasik:
Please resolve conflicts, @d-helios.

packer/files/node-exporter.service
QuerthDP, line 1: Rebase instead of merging master into here

swasik (approved):

QuerthDP:
@swasik any idea why didn't the checks run here?

QuerthDP:
Rebased on master to re-run the checks.

PR #551, opened 2026-08-10
ci: build release per architecture on native runners

swasik:
## Motivation

The release workflow built both `amd64` and `arm64` images in a single `build-docker` job on `ubuntu-latest`, emulating arm64 with QEMU, which is slow. Building and pushing also happened in the same job, so nothing (e.g. validator tests) could be inserted between building a release and publishing it.

## Changes

- `scripts/build-release` and `scripts/build-dockers` accept an optional list of architectures as arguments, defaulting to the full `amd64 arm64` release list so the local release flow described in `docs/releasing.md` is unchanged. With a single architecture, `build-dockers` skips the combined multi-arch image — the only part that needed QEMU.
- The release workflow's `build-docker` job is split in two:
  - a matrix **`build`** job (`amd64` on `ubuntu-latest`, `arm64` on `ubuntu-24.04-arm`, same runner mapping as `daily.yml`) builds each architecture natively via `./scripts/build-release <arch>` and uploads the docker image, release tarball and SBOMs as per-arch artifacts;
  - a **`push-docker`** job downloads the per-arch images, pushes them, assembles the multi-arch manifest, and archives the combined `vector-store-sboms-<version>` artifact under the same name as before.
- `build-cloud-images` is untouched (Packer downloads the tarball from the GitHub release assets, not from these jobs); `announce` now depends on `push-docker` instead of `build-docker`.

vector-store is now built exactly once per architecture, natively, and the resulting artifacts are reusable by downstream jobs. A follow-up (VECTOR-790) will run the validator test suite against these artifacts and gate publishing on them.

## Test plan

- [x] `actionlint` clean on `.github/workflows/release.yaml`
- [x] `shellcheck -S warning` clean on `scripts/build-release` and `scripts/build-dockers`
- [ ] Dry-run the workflow via `workflow_dispatch` against a dev tag to verify the arm64 native build and the load/push/manifest path end to end

Refs: VECTOR-848

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -13,10 +13,21 @@ env:
   VECTOR_VERSION: ${{ github.event_name == 'release' && github.event.release.tag_name || github.event.inputs.VECTOR_VERSION }}
 
 jobs:
-  build-docker:
-    runs-on: ubuntu-latest
+  # Build vector-store once per architecture, natively (no QEMU emulation),
+  # and expose the results (binary tarball, docker image, SBOMs) as workflow
+  # artifacts for the downstream jobs.
+  build:
+    runs-on: ${{ matrix.os }}
     permissions:
       contents: read
+    strategy:
+      matrix:
+        os: [ubuntu-latest, ubuntu-24.04-arm]
+        include:
+          - os: ubuntu-latest
+            arch: amd64
+          - os: ubuntu-24.04-arm
+            arch: arm64
     steps:
       - name: Checkout repository
         uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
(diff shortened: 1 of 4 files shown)


docs/releasing.md
```diff
@@ -29,7 +29,12 @@ version tag. Then from the root of the repository run:
 ```
 
 This will create a tar.gz release in the `releases/{amd64/arm64}` directory,
-per-architecture docker images, and a multi-platform docker manifest. You can
+per-architecture docker images, and a multi-platform docker manifest. The
+script optionally accepts a subset of architectures to build as arguments
+(e.g. `./scripts/build-release arm64`); a single-architecture build runs
+natively and doesn't create the multi-platform docker image, so it doesn't
+need qemu (this is how the release workflow builds each architecture on its
+own native runner). You can
```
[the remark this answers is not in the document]
swasik: Good catch — reworded in b15d91a: a single-arch build is qemu-free only when it matches the host; building a non-native architecture still requires qemu.

.github/workflows/release.yaml
```diff
@@ -57,13 +107,20 @@ jobs:
...
         with:
           name: vector-store-sboms-${{ env.VECTOR_VERSION }}
           path: |
-            vector-store-${{ env.VECTOR_VERSION }}.cdx.json
-            vector-store-docker-${{ env.VECTOR_VERSION }}-*.cdx.json
+            sboms/vector-store-${{ env.VECTOR_VERSION }}.cdx.json
+            sboms/vector-store-docker-${{ env.VECTOR_VERSION }}-*.cdx.json
           if-no-files-found: warn
 
   build-cloud-images:
```
ewienik, line 126: build-cloud-images needs push-docker

.github/workflows/release.yaml
```diff
@@ -13,38 +13,88 @@ env:
...
+
       - name: Log in to Docker Hub
         uses: docker/login-action@9780b0c442fbb1117ed29e0efdff1e18412f7567 # v3.3.0
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
 
-      - name: Build images, SBOMs and manifest
-        run: ./scripts/build-release
-
```
ewienik, line 47: When we remove build-release here there is no place to create multi-arch docker. Additionally multi-arch docker needs qemu for arch and docker support for multi-platform support. Consider calling scripts/prepare-docker-qemu and creating multi-platform docker here.

scripts/build-dockers
```diff
@@ -28,13 +31,18 @@ $dockerfile
...
+        [[ -z $platform ]] && platform="linux/$arch" || platform="$platform,linux/$arch"
+    done
 
-echo Build the multi-arch docker image for $platform
-docker build --platform $platform --tag $docker_tag -f- . <<EOF
+    echo Build the multi-arch docker image for $platform
+    docker build --platform $platform --tag $docker_tag -f- . <<EOF
 $dockerfile
 EOF
+fi
```
ewienik, line 46: We need a way to build multi-platform docker from existing dockers.

ewienik (changes requested):

scripts/build-dockers
swasik: Multi-platform dockers are not working at the moment anyway. I plan to add them in a separate PR (VECTOR-851) to solve issues one by one.

.github/workflows/release.yaml
swasik: Multi-platform dockers are not working at the moment anyway. I plan to add them in a separate PR (VECTOR-851) to solve issues one by one.

.github/workflows/release.yaml
swasik: Added in 1cc0e6e — build-cloud-images now has needs: [push-docker].

ewienik (approved):

PR #552, opened 2026-08-11
scripts: assemble multi-arch docker image as manifest at upload time

swasik:
## Motivation

The 1.10.0 release run failed in the `build-docker` job:

https://github.com/scylladb/vector-store/actions/runs/31197502662/job/92929245847

```
Build the multi-arch docker image for linux/amd64,linux/arm64
ERROR: failed to build: Multi-platform build is not supported for the docker driver.
```

`scripts/build-dockers` built the per-arch images successfully and then tried to build a combined multi-arch image with `docker build --platform linux/amd64,linux/arm64`, which the default docker driver does not support — a multi-platform image cannot be stored in the classic docker image store.

Since #551 the release workflow builds each architecture natively on its own runner and never hits this path, but any multi-arch invocation of the script (the default, e.g. a local `./scripts/build-release`) still fails the same way.

## Changes

The locally built multi-arch image was redundant: the release workflow already assembles the multi-arch tag with `docker manifest` from the pushed per-arch images. This PR aligns the scripts with that approach:

- `scripts/build-dockers`: drop the combined multi-arch `docker build`; only the per-arch images are built.
- `scripts/upload-dockers`: instead of pushing a local combined image, push the per-arch images and then create and push the multi-arch manifest (`docker manifest create --amend` + `docker manifest push --purge`), matching the release workflow's `push-docker` job.
- `docs/releasing.md`: document the manifest-at-upload flow; qemu is only needed for building non-native architectures.

## Test plan

- [x] `bash -n` on both modified scripts
- [ ] Manual release dry-run: `./scripts/build-release` on a machine with the default docker driver (previously failed, now builds per-arch images only)
- [ ] Verify `./scripts/upload-dockers` pushes per-arch tags and the multi-arch manifest resolves for both architectures

Fixes: VECTOR-851


docs/releasing.md
```diff
@@ -28,20 +28,25 @@ version tag. Then from the root of the repository run:
...
-This will create a tar.gz release in the `releases/{amd64/arm64}` directory,
-per-architecture docker images, and a multi-platform docker manifest. The
-script optionally accepts a subset of architectures to build as arguments
-(e.g. `./scripts/build-release arm64`). A single-architecture build doesn't
-create the multi-platform docker image, so it doesn't need qemu as long as
-the requested architecture matches the host — this is how the release
-workflow builds each architecture natively on its own per-arch runner.
-Building a non-native architecture still requires qemu. You can
-push the docker images and manifest to docker hub:
+This will create a tar.gz release in the `releases/{amd64/arm64}` directory
```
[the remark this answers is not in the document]
swasik: Fixed in e2175a9 — the docs now point to `target/{amd64,arm64}/release`, which is where `build-release` writes the archives and `upload-release` reads them from.

docs/releasing.md
```diff
@@ -28,20 +28,25 @@ version tag. Then from the root of the repository run:
...
+and per-architecture docker images. The script optionally accepts a subset of
+architectures to build as arguments (e.g. `./scripts/build-release arm64`).
+Building an architecture matching the host doesn't need qemu — this is how
+the release workflow builds each architecture natively on its own per-arch
+runner. Building a non-native architecture still requires qemu.
```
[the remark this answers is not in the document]
swasik: Fixed in e2175a9 — the opening paragraph and Prerequisites now state that qemu is needed only when building a non-native architecture.

scripts/upload-dockers
```diff
@@ -12,7 +12,13 @@ vector_store_version=$(./scripts/run-with-release-toolchain cargo info vector-st
...
 for arch in $archs ; do
     docker push $docker_tag-$arch
+    arch_tags="$arch_tags $docker_tag-$arch"
 done
-docker push $docker_tag
+
+# combine the pushed per-arch images into a multi-arch manifest, the classic
+# docker image store cannot hold a multi-platform image locally
+docker manifest create --amend $docker_tag $arch_tags
+docker manifest push --purge $docker_tag
```
ewienik: Are you sure it is working? I have error output:
```
docker.io/scylladb/vector-store:1.10.0-amd64 is a manifest list
```

ewienik (dismissed):

swasik:
You're right, thanks for catching it — `docker manifest create` does not work here. Fixed in 69d0467.

**What happens**

`docker manifest create` refuses a source image that is itself an index. A single-platform `docker build --platform linux/amd64` produces an index whenever buildx attaches provenance/SBOM attestations to it, which the containerd image store does by default. So the per-arch tags are indexes, not plain image manifests.

The tags already on Docker Hub confirm it — `scylladb/vector-store:1.10.0-amd64` is an OCI index holding the image plus an `unknown/unknown` attestation manifest:

```
1.10.0-amd64 -> application/vnd.oci.image.index.v1+json
                 [ {amd64, linux}, {unknown, unknown} ]
```

**Reproduced and verified** against a local registry (containerd image store, docker 29.5.2 / buildx 0.34.0), building two single-arch images exactly the way `build-dockers` does:

```
docker manifest create --amend localhost:5555/vs-test:1.0.0 ...-amd64 ...-arm64
  -> localhost:5555/vs-test:1.0.0-amd64 is a manifest list      (exit 1)

docker buildx imagetools create --tag localhost:5555/vs-test:1.0.0 ...-amd64 ...-arm64
  -> DONE                                                        (exit 0)
     result: index [ {amd64,linux}, {unknown,unknown}, {arm64,linux}, {unknown,unknown} ]
```

**The fix** is `docker buildx imagetools create`, which accepts index sources and copies all their child manifests, so the attestations are preserved. The resulting structure matches the existing `scylladb/vector-store:1.10.0` tag on Docker Hub exactly (amd64 + arm64 + both attestations), so this produces the same thing the tag already has.

I applied it to the `push-docker` job in the release workflow too — it had the same `docker manifest create --amend` and would have hit this as soon as the runners switch to the containerd image store.

swasik:
Follow-up in de86441: the per-arch images are now built with `--provenance=false --sbom=false`, so the combined tag lists exactly the two real platforms instead of carrying buildx's `unknown/unknown` attestation manifests. The release SBOMs come from syft and cargo-cyclonedx anyway, so nothing is lost.

Verified end-to-end against a local registry, running the same command shapes the scripts now use:

```
per-arch tag  -> application/vnd.docker.distribution.manifest.v2+json   (plain image, no index)

combined tag  -> application/vnd.docker.distribution.manifest.list.v2+json
                   linux/amd64
                   linux/arm64
```

Compare with what is published for 1.10.0 today, which has four entries (amd64, arm64 and two `unknown/unknown`).

I kept `imagetools create` for the combining step rather than going back to `docker manifest create` — both work now that the sources are plain manifests, but imagetools also works if attestations ever get re-enabled, so the upload step no longer depends on how the images were built.

swasik:
Switched to keeping the attestations, in 81dc3aa. `--provenance=false --sbom=false` is reverted; instead the per-arch images are combined **by the digest of their `linux/<arch>` manifest** rather than by their tags, so the attestation manifests are simply not carried into the combined tag.

That keeps the provenance — including the resolved `redhat/ubi9-minimal` base image digest, which nothing else in the release records — discoverable on the per-arch tags, while the multi-arch tag stays clean.

Verified end-to-end against a local registry with attestations on (buildx default), running the new logic verbatim:

```
resolved sources: vs-test@sha256:d2e4a53d…  vs-test@sha256:d33233fd…

combined tag   ->  linux/amd64
                   linux/arm64                       (exactly two entries)

per-arch tag   ->  linux/amd64
                   unknown/unknown                   (attestation retained)

imagetools inspect --format '{{json .Provenance}}' <tag>-amd64
               ->  {"SLSA": {"buildDefinition": … }}   (still retrievable)
```

The fallback branch is covered too: if a per-arch image is ever built without attestations the digest lookup yields nothing and the tag is used directly, so the upload works either way.

`scripts/upload-dockers` now uses `jq` — already a dependency of the daily and validator workflows, and added to the Prerequisites list in `docs/releasing.md`. The same logic is inlined in the `push-docker` job since that job has no checkout.

scripts/upload-dockers
swasik: You are right - I tested it incorrectly. Now it should be fixed.

docs/releasing.md
```diff
@@ -63,8 +89,9 @@ authenticated.
 
 ### Prerequisites
 
-- Docker with qemu support for multi-platform builds (see
-  `scripts/prepare-docker-qemu`)
+- Docker; qemu support is needed only when building a non-native
```
ewienik, line 92: nit: do you think it is needed to mention that docker buildx is needed? I doubt if it is installed with every docker package.

ewienik (dismissed):

docs/releasing.md
swasik: Good point, and it is more than a nit — this PR is what introduced the dependency. Before it, `upload-dockers` used only `docker push` and `docker manifest`, both built into the docker CLI; `docker buildx imagetools` needs the plugin.

And it is genuinely packaged separately, not bundled with the engine:

```
docker-buildx-plugin   Docker's own apt/yum repositories
docker-buildx          Debian/Ubuntu universe
```

Added to Prerequisites in 3fe8f41.

swasik: Added a preflight check as well, in 1c70e49 — docs alone would not have stopped anyone, and the tooling is only needed *after* the per-arch images are pushed, so a late failure would leave them published without the multi-arch tag pointing at them. Now it refuses to start:

```
error: the docker buildx plugin is required to assemble the multi-arch tag.
It is packaged separately from the docker engine, as docker-buildx-plugin
in the docker repositories and as docker-buildx in Debian/Ubuntu.
```

Same for `jq`. Verified all three paths with stubbed binaries — buildx missing, jq missing, and both present — the first two exit 1 before any push, the third proceeds.

ewienik (approved):

PR #557, opened 2026-08-17
vector-store: parametrize max points for diskann

Akvear:
### Motivations
Unlike in usearch, where the index size can be changed, in diskann the index size is determined on index creation and can't be changed later. For the sake of the PoC we specified a constant 1M vectors limit for the diskann index (per partition). The current setup would not work for indexes larger than 1M vectors (vectors added after 1M would not be added to the diskann index with a warning logged). This means it is necessary to either lift the cap (not possible due to diskann crate) or make it parametrizable. 

### Why not set MAX_POINTS to some stupidly large constant?
Because diskann preallocates memory for the index, a large constant would be a massive waste of resources for majority of use cases.

Fixes: VECTOR-828


Akvear:
Should the variable be required for diskann (then it could replace USE_DISKANN) or is it OK to have a default MAX_POINTS value?

crates/vector-store/src/vs_index/diskann.rs
```diff
@@ -637,4 +637,47 @@ mod tests {
...
+            index,
+        };
+        let size = AtomicUsize::new(0);
+
+        // Try to insert more vectors than the cap allows.
+        for (i, embedding) in embeddings.into_iter().enumerate() {
+            add_vector(&partition, &size, PrimaryId::from(i as u64), embedding).await;
+        }
+
+        assert_eq!(size.load(Ordering::Relaxed), max_points.get());
```
QuerthDP, line 681: So we just ignore the fact? Is it logged anywhere that the maximum was reached?

QuerthDP (commented):
The code is really fine, but I'm not sure about the motivation for this change.
Why do we need to make it configurable? Should we really just ignore the fact when we reach maximum and just stop adding vectors like nothing happened?

QuerthDP:
> Should the variable be required for diskann (then it could replace USE_DISKANN) or is it OK to have a default MAX_POINTS value?

I think it shouldn't be required. 1M seems like a fine default, especially as I understand it's for one partition.
Tell me if I'm wrong

Akvear:
>Why do we need to make it configurable? 

Unlike usearch, the diskann indexes have a constant size determined on index creation - so if we encounter an index with size let's say 100M, then the 1M cap wouldn't work. On the other hand, we can't just set MAX_POINTS to 100M because it preallocates memory and it 100M would likely be too much for most cases. 

>Should we really just ignore the fact when we reach maximum and just stop adding vectors like nothing happened?

There is a warning logged, the alternative is to fully abort.

crates/vector-store/src/vs_index/diskann.rs
Akvear: It is logged in `add_vector()` with an error message returned from the diskann crate.

QuerthDP:
> > Why do we need to make it configurable?
> 
> Unlike usearch, the diskann indexes have a constant size determined on index creation - so if we encounter an index with size let's say 100M, then the 1M cap wouldn't work. On the other hand, we can't just set MAX_POINTS to 100M because it preallocates memory and it 100M would likely be too much for most cases.
> 
> >

Thanks, really good explanaition. It should be **at least** mentioned in PR cover.

crates/vector-store/src/config_manager.rs
```diff
@@ -448,6 +449,15 @@ pub async fn load_config(env: impl Fn(&str) -> anyhow::Result<String>) -> anyhow
         );
     }
 
+    if let Ok(diskann_max_points) = env("VECTOR_STORE_DISKANN_MAX_POINTS") {
+        let max_points = diskann_max_points.trim().parse::<usize>().map_err(|_| {
```
QuerthDP: Could we parse it as NonZeroUsize?

README.md
```diff
@@ -38,6 +38,7 @@ service supports also `.env` files.
 | `VECTOR_STORE_USEARCH_SIMULATOR`           | Enable simulator for USearch. Provides human readable delays for simulated operations (`search:add-remove:reserve`).                                                                 |                          |
 | `VECTOR_STORE_USE_DISKANN`                 | Use DiskANN as the indexing engine instead of USearch.                                                | `false`                  |
 | `VECTOR_STORE_DISKANN_ALPHA`               | DiskANN parameter that controls the trade-off between index quality and build time. | (DiskANN default)                    |
+| `VECTOR_STORE_DISKANN_MAX_POINTS`          | Maximum number of vectors a single DiskANN partition index can hold. | `1000000`                    |
```
QuerthDP: Maybe some spacing here to make it straightforward looking like 1M?

Akvear:
Changes:
- update readme to be more readable
- parse max_points as a NonZeroUsize instead of usize

crates/vector-store/src/config_manager.rs
```diff
@@ -951,17 +962,39 @@ mod tests {
...
+            "VECTOR_STORE_DISKANN_MAX_POINTS",
+            "0".into(),
+        )]));
+        let result = load_config(env).await;
+        assert!(result.is_err());
+        assert!(
+            result
+                .unwrap_err()
+                .to_string()
+                .contains("Invalid VECTOR_STORE_DISKANN_MAX_POINTS")
```
QuerthDP: The test fails here. Now the message is different

Akvear: fixed

QuerthDP (approved):
Thanks, looks good now.

Akvear:
rebased on master

Akvear:
Different validator tests keep failing in the CI, seemingly randomly.

QuerthDP:
> Different validator tests keep failing in the CI, seemingly randomly.

There is an issue with auth in CI. We're working on a fix. I'll ping you to rebase once it is fixed.

QuerthDP:
Once https://github.com/scylladb/vector-store/pull/562 is merged it should be all fine. Please try again then

PR #558, opened 2026-08-18
vector-store: implement FTS highlighting engine

QuerthDP:
This PR adds **Highlighting** support for Full-Text Search. Given a query and a document's text, return the matched terms wrapped in markup so an application can show why a row matched.

This PR lands the FTS index engine piece. Computing a highlighted excerpt for a query against caller-supplied text, per the [Full Text Search Highlighting Design](https://scylladb.atlassian.net/wiki/spaces/RND/pages/404750371/Full+Text+Search+Highlighting+Design) doc. We went with the "stateless" design rather than having the index store document text itself. Storing text would mean a `STORED` field for every FTS-indexed column, which is RAM we don't need to spend — the coordinator already has the row content available when it wants a highlight, so the index only needs to analyze the query and weight terms by document frequency, keeping highlights consistent with the BM25 score of that same query.

The HTTP endpoint that exposes this to callers follow in separate PRs.

Fixes: VECTOR-807

### DIFF
--- a/crates/vector-store/src/fts_index/tantivy.rs
+++ b/crates/vector-store/src/fts_index/tantivy.rs
@@ -234,6 +234,12 @@ fn handle_remove_document(
     state.writer.write().unwrap().rm_document(term, in_progress)
 }
 
+/// A query-related failure caused by the caller's input (an unparsable query, or a query
+/// construct that this endpoint cannot process) rather than an internal/actor failure.
+#[derive(Debug, thiserror::Error)]
+#[error("{0}")]
+pub(crate) struct QueryError(pub(crate) String);
+
 fn make_query(
     index: &tantivy::Index,
     body_field: tantivy::schema::Field,
@@ -242,7 +248,7 @@ fn make_query(
     let query_parser = QueryParser::for_index(index, vec![body_field]);
     query_parser
         .parse_query(query_str)
-        .map_err(|e| anyhow!("fts: failed to parse query: {e}"))
+        .map_err(|e| QueryError(format!("fts: failed to parse query: {e}")).into())
 }
 
 fn find_partition_id(
--- a/crates/vector-store/src/fts_index/tantivy.rs
+++ b/crates/vector-store/src/fts_index/tantivy.rs
@@ -15,6 +15,10 @@ use tantivy::ReloadPolicy;
 use tantivy::TantivyDocument;
 use tantivy::collector::TopDocs;
 use tantivy::indexer::IndexWriterOptions;
+use tantivy::query::BooleanQuery;
+use tantivy::query::BoostQuery;
+use tantivy::query::Occur;
+use tantivy::query::Query;
 use tantivy::query::QueryParser;
 use tantivy::schema::INDEXED;
 use tantivy::schema::IndexRecordOption;
@@ -23,6 +27,7 @@ use tantivy::schema::Schema;
 use tantivy::schema::TextFieldIndexing;
 use tantivy::schema::TextOptions;
 use tantivy::schema::Value;
+use tantivy::snippet::SnippetGenerator;
 use tantivy::tokenizer::Language;
 use tantivy::tokenizer::LowerCaser;
 use tantivy::tokenizer::SimpleTokenizer;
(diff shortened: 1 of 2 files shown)


crates/vector-store/src/fts_index/tantivy.rs
```diff
@@ -300,6 +302,35 @@ fn handle_search(
...
+const HIGHLIGHT_POST_TAG: &str = "</b>";
+
+fn handle_highlight(state: &IndexState, query_str: &str, documents: &[String]) -> FtsHighlightR {
+    let body_field = state.schema.get_field("body").unwrap();
+    let searcher = state.reader.searcher();
+    let query = make_query(&state.index, body_field, query_str)?;
+
+    // The generator weights terms by their document frequency in the live index,
+    // so highlighting stays consistent with what BM25 scored - even though the
+    // text itself comes from the caller.
```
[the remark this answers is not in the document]
QuerthDP: Reworded

crates/vector-store/src/fts_index/tantivy.rs
```diff
@@ -300,6 +302,35 @@ fn handle_search(
...
+    let mut generator = SnippetGenerator::create(&searcher, query.as_ref(), body_field)
+        .map_err(|e| anyhow!("fts: failed to create snippet generator: {e}"))?;
```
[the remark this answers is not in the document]
QuerthDP: Now we pre-check the boolean query for negations and drop those before highlighting. Once we cannot check that, we skip the query returning NULL values.

QuerthDP:
Changelog:
- rebased on master

QuerthDP:
Changelog:
- add `QueryError` to distinguish failed queries due to bad querying from internal errors
- error out on boost queries instead of returning misleading null as none match

QuerthDP:
Changelog:
- remove unused import

crates/vector-store/src/fts_index/tantivy.rs
```diff
@@ -242,7 +248,7 @@ fn make_query(
     let query_parser = QueryParser::for_index(index, vec![body_field]);
     query_parser
         .parse_query(query_str)
-        .map_err(|e| anyhow!("fts: failed to parse query: {e}"))
+        .map_err(|e| QueryError(format!("fts: failed to parse query: {e}")).into())
```
knowack1, line 257: I think this commit is a good starting point to fix https://scylladb.atlassian.net/browse/VECTOR-744.

crates/vector-store/src/fts_index/tantivy.rs
```diff
@@ -850,6 +934,286 @@ mod tests {
...
+        let highlights = highlight(&sender, "café", &["I love a nice café over über coffee"]).await;
+
+        assert_eq!(
+            highlights[0].as_deref(),
+            Some("I love a nice <b>café</b> over über coffee")
+        );
+    }
+
+    #[tokio::test]
+    async fn highlight_batched_matches_per_document_analysis() {
```
knowack1: What value does this test bring? It's not something I would expect. Also, the implementation actually iterates over documents, so whether we provide one or N documents, it seems to do the same job for a specific document. Consider removing this test.

knowack1 (dismissed):
Looks good. One unit test that I would consider to remove.

crates/vector-store/src/fts_index/tantivy.rs
QuerthDP: Thanks for noticing this. I'll fix that in separate PR.

QuerthDP:
Changelog: 
- removed `highlight_batched_matches_per_document_analysis` test

crates/vector-store/src/fts_index/tantivy.rs
QuerthDP: That's right, let's remove it.

crates/vector-store/src/fts_index/tantivy.rs
QuerthDP: https://github.com/scylladb/vector-store/pull/566

knowack1 (approved):

QuerthDP:
CI failed due to known issue: VECTOR-877

PR #559, opened 2026-08-18
vector-store: expose FTS highlighting over the HTTP API

QuerthDP:
A ScyllaDB coordinator running a BM25 search needs a way to show why a result matched — a highlighted excerpt of the matched text — without duplicating the query's term weighting on the coordinator side.

This PR exposes the FTS highlighting engine over a new `POST /indexes/{keyspace}/{index}/highlight` endpoint. It accepts a query and a list of caller-supplied document texts and returns one highlighted excerpt per document, in the same order, with null for documents that matched no query term. It adds the request/response types, wires the route (reusing the existing not-serving/in-progress status handling from the bm25 endpoint), and covers the new endpoint with integration tests.

Fixes: VECTOR-808

### DIFF
--- a/api/openapi.json
+++ b/api/openapi.json
@@ -1064,6 +1064,46 @@
           }
         }
       },
+      "PostIndexHighlightRequest": {
+        "type": "object",
+        "description": "Request body for highlighting.",
+        "required": [
+          "query",
+          "documents"
+        ],
+        "properties": {
+          "documents": {
+            "type": "array",
+            "items": {
+              "type": "string"
+            },
+            "description": "The documents to highlight."
+          },
+          "query": {
+            "type": "string",
+            "description": "The text query whose terms should be marked in the documents."
+          }
+        }
+      },
+      "PostIndexHighlightResponse": {
+        "type": "object",
+        "description": "Response for highlighting.",
+        "required": [
+          "highlights"
+        ],
+        "properties": {
+          "highlights": {
+            "type": "array",
+            "items": {
+              "type": [
+                "string",
+                "null"
+              ]
+            },
+            "description": "The highlighted documents. Each document is a string of text with the query terms marked."
+          }
+        }
+      },
       "SimilarityFunction": {
         "type": "string",
         "description": "Similarity function used to compare vectors, as set at index creation.",
--- a/api/openapi.json
+++ b/api/openapi.json
@@ -309,6 +309,100 @@
         }
       }
     },
+    "/api/v1/indexes/{keyspace}/{index}/highlight": {
+      "post": {
+        "tags": [
+          "scylla-vector-store-index"
+        ],
+        "description": "Computes a highlighted excerpt (snippet) for each of the supplied document texts. The index is used only to analyze the query and to weight its terms by their document frequency, prioritizing rarer terms when picking which fragment of a long text to show. Highlights are returned in the same order as the requested documents. A document that matches no query term yields `null`. If TLS is enabled on the server, clients must connect using a HTTPS protocol.",
+        "operationId": "post_index_highlight",
+        "parameters": [
+          {
+            "name": "keyspace",
+            "in": "path",
+            "description": "The name of the ScyllaDB keyspace containing the index.",
+            "required": true,
+            "schema": {
+              "$ref": "#/components/schemas/KeyspaceName"
+            }
+          },
+          {
+            "name": "index",
+            "in": "path",
+            "description": "The name of the full-text index within the specified keyspace to highlight against.",
+            "required": true,
+            "schema": {
+              "$ref": "#/components/schemas/IndexName"
+            }
+          }
+        ],
+        "requestBody": {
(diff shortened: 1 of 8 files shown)


QuerthDP:
Based on: https://github.com/scylladb/vector-store/pull/558

crates/vector-store/src/httproutes.rs
```diff
@@ -1053,6 +1073,141 @@ async fn post_index_bm25(
...
+        .highlight(index_key, request.query, request.documents)
+        .await;
+
+    timer.observe_duration();
+
+    match result {
+        Err(err) => {
+            let msg = format!("index.highlight request error: {err}");
+            debug!("post_index_highlight: {msg}");
+            (StatusCode::INTERNAL_SERVER_ERROR, msg).into_response()
```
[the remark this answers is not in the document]
QuerthDP: Now returns 400 properly.

crates/vector-store/src/fts_index/tantivy.rs
```diff
@@ -300,6 +307,68 @@ fn handle_search(
...
+/// in caller-supplied text even though the query excludes documents containing it.
+///
+/// Returns `Err` for a query construct we cannot see inside since we cannot rule
+/// out a `MustNot` clause hidden behind it. The caller treats this the same
+/// as "cannot be highlighted" rather than as a hard error.
+fn strip_negated_clauses(query: &dyn Query) -> anyhow::Result<Box<dyn Query>> {
+    let Some(boolean_query) = query.downcast_ref::<BooleanQuery>() else {
+        if query.downcast_ref::<BoostQuery>().is_some() {
+            bail!("fts: boosted sub-queries are not supported for highlighting");
+        }
```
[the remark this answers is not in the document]
QuerthDP: Fixed. Now the queries are rejected.

crates/vector-store/src/httproutes.rs
```diff
@@ -1073,6 +1074,146 @@ async fn post_index_bm25(
...
+            debug!("post_index_highlight: {msg}");
+            let status = if err.downcast_ref::<QueryError>().is_some() {
+                StatusCode::BAD_REQUEST
+            } else {
+                StatusCode::INTERNAL_SERVER_ERROR
+            };
+            (status, msg).into_response()
+        }
+        Ok(highlights) => {
+            if highlights.len() != documents_len {
```
knowack1: Can this if ever be false? I think not as we iterate over documents. Consider removing it.

crates/vector-store/tests/integration/fts.rs
```diff
@@ -409,6 +409,168 @@ async fn fts_bm25_search_returns_503_and_index_building_reason_when_index_is_bui
     );
 }
 
+#[tokio::test]
```
knowack1, line 465: Wondering why we're adding this test suite in the middle instead of at the end of the existing tests.

knowack1 (commented):
Looks great. Nice and clean suite of integration tests.

QuerthDP:
Changelog:
- rebased on master

crates/vector-store/tests/integration/fts.rs
QuerthDP: No particular reason, I'll move them to the end.

crates/vector-store/src/httproutes.rs
QuerthDP: It is true that this is always false. Removing

QuerthDP:
Changelog:
- moved integration tests to the end of the file
- removed the impossible check on `highlights.len() != documents_len`

api/openapi.json
```diff
@@ -1064,6 +1064,46 @@
...
+          "query",
+          "documents"
+        ],
+        "properties": {
+          "documents": {
+            "type": "array",
+            "items": {
+              "type": "string"
+            },
+            "description": "The documents to highlight. Each document is a string of text."
```
knowack1: What do we mean exactly by "string of text"? As string is already a type of this field (specified in schema) I think there is not reason to say this again in description

api/openapi.json
```diff
@@ -1064,6 +1064,46 @@
...
+          },
+          "query": {
```
knowack1, line 1176: I think we should be more specific what this query actually is. This comment also applies to the "query" in bm25 request. Something to be considered for the future.

knowack1 (dismissed):

api/openapi.json
QuerthDP: VECTOR-879

api/openapi.json
QuerthDP: Removed the second sentence.

QuerthDP:
Changelog:
- updated the "documents" description in OpenAPI

QuerthDP:
CI failed due to: VECTOR-881

knowack1 (approved):
Thanks for addressing all the comments. Approving.

PR #560, opened 2026-08-18
validator: add E2E tests for CQL `HIGHLIGHT()` function

QuerthDP:
The coordinator's `HIGHLIGHT(column, 'query')` reaches the Vector Store over a second request per query, and nothing exercised that path as a whole.

The E2E tests cover what no lower layer reaches: each row's excerpt cut from its own content and marked with the terms that row matched, the BM25 relevance order and LIMIT left undisturbed, deleted rows dropping out, multi-byte text surviving the round trip, a column without a fulltext index rejected, and the request body limit that bounds how much content one query can carry — the coordinator ships every matched row's text in a single request, so a large enough answer is refused with 413 rather than truncated.

Highlighting behaviour that lives entirely in the Tantivy actor is covered by unit tests in the first patch instead, where the assertions can be exact.

Requires the ScyllaDB side to be merged first: scylladb/scylladb#31166.

Fixes: VECTOR-818

### DIFF
--- a/crates/validator/src/fts.rs
+++ b/crates/validator/src/fts.rs
@@ -13,6 +13,7 @@ use httpapi::KeyspaceName;
 use httpclient::HttpClient;
 use scylla::client::session::Session;
 use scylla::response::query_result::QueryRowsResult;
+use std::collections::HashMap;
 use std::sync::Arc;
 use std::sync::RwLock;
 use tracing::info;
@@ -169,6 +170,40 @@ impl Fixture {
     }
 
     #[framed]
+    async fn highlight_search(&self, query: &str, limit: usize) -> Vec<(i32, Option<String>)> {
+        let cql = self.highlight_select_query(query, limit);
+        let result = get_opt_query_results(&cql, &self.session)
+            .await
+            .unwrap_or_else(|| panic!("HIGHLIGHT query '{query}' failed to return results"));
+        Self::extract_pk_excerpts(&result)
+    }
+
+    #[framed]
+    async fn highlight_excerpts(&self, query: &str) -> HashMap<i32, String> {
+        self.highlight_search(query, 100)
+            .await
+            .into_iter()
+            .map(|(pk, excerpt)| {
+                let excerpt = excerpt
+                    .unwrap_or_else(|| panic!("expected a highlight for pk {pk} of '{query}'"));
+                (pk, excerpt)
+            })
+            .collect()
+    }
+
+    fn highlight_select_query(&self, query: &str, limit: usize) -> String {
+        highlight_select_query(&self.table, query, limit)
+    }
+
+    fn extract_pk_excerpts(result: &QueryRowsResult) -> Vec<(i32, Option<String>)> {
+        result
+            .rows::<(i32, Option<String>)>()
+            .expect("failed to get rows")
+            .map(|row| row.expect("failed to get row"))
+            .collect()
+    }
+
+    #[framed]
     async fn init_fts_table(
         actors: &TestActors,
     ) -> (Arc<Session>, Vec<HttpClient>, KeyspaceName, TableName) {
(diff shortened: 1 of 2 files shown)


QuerthDP:
Requires https://github.com/scylladb/vector-store/pull/559 and Scylla side to be merged first.

crates/vector-store/src/httproutes.rs
```diff
@@ -1053,6 +1074,146 @@ async fn post_index_bm25(
...
+    request_body = httpapi::PostIndexHighlightRequest,
+    responses(
+        (
+            status = 200,
+            description = "Successful highlighting. Returns one highlight per requested document, in the requested order.",
+            body = httpapi::PostIndexHighlightResponse
+        ),
+        (
+            status = 400,
+            description = "Bad request. Possible causes: malformed input, unparsable query, or missing required fields.",
```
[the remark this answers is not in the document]
QuerthDP: Not about this PR. Although the BoostQueries are not parsable by us and we cannot check if the negation is a subquery so we don't allow for any.

crates/vector-store/src/httproutes.rs
```diff
@@ -1053,6 +1074,146 @@ async fn post_index_bm25(
...
+            content_type = "application/json",
+            body = ErrorMessage
```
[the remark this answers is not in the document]
QuerthDP: Not related to this PR.

crates/validator/src/fts.rs
```diff
@@ -155,6 +156,52 @@ impl Fixture {
...
 
+    #[framed]
+    async fn highlight_search(
+        &self,
+        query: &str,
+        limit: usize,
+        expected_count: usize,
+    ) -> Vec<(i32, Option<String>)> {
+        let cql = self.highlight_select_query(query, limit);
+        let result = wait_for_value(
```
knowack1: I think the `wait_for` value is not needed here, as this function is usually called after `fixture.create_fts_index()`, which already ensures that all documents are in the index  - so no active polling is required. This will also simplify the interface so that expected_count is not needed:
```
let excerpts = fixture.highlight_excerpts("fox", 2).await;
```
->
```
let excerpts = fixture.highlight_excerpts("fox").await;
```

crates/validator/src/fts.rs
```diff
@@ -741,3 +796,432 @@ async fn fts_large_document_set(fixture: Arc<Fixture>) {
...
+            (1, "the quick brown fox jumps over the lazy dog"),
+            (2, "a slow turtle walks through the garden"),
+            (3, "the fox runs across the meadow"),
+        ])
+        .await;
+    fixture.create_fts_index().await;
+
+    let excerpts = fixture.highlight_excerpts("fox", 2).await;
+
+    for (pk, excerpt) in &excerpts {
```
knowack1: Consider (I prefer) explicit asserts instead of loop.

crates/validator/src/fts.rs
```diff
@@ -741,3 +796,432 @@ async fn fts_large_document_set(fixture: Arc<Fixture>) {
...
+    assert!(
+        excerpt.contains("<b>Fox</b>"),
+        "expected the original casing '<b>Fox</b>' for pk 1, got: {excerpt}"
+    );
+
+    info!("finished");
+}
+
+#[e2etest::test(group = fts)]
+async fn highlight_escapes_html_in_content_but_not_the_marker_tags(fixture: Arc<Fixture>) {
```
knowack1: Some of this test (also others) overlaps with unit tests that we already have in tantivy.rs. Since this seems to be solely a Tantivy / tantivy.rs actor implementation, I'm not sure if we need to check this at the E2E level as well. Something to consider, though I'm fine keeping them.

crates/validator/src/fts.rs
```diff
@@ -741,3 +796,432 @@ async fn fts_large_document_set(fixture: Arc<Fixture>) {
 
     info!("finished");
 }
+
```
knowack1, line 833: I would be interested in seeing a test that verifies the limit we have on document(s) size due to a single, non-chunked POST to the /highlights endpoint.

QuerthDP:
Changelog:
- rebased on master

crates/validator/src/fts.rs
QuerthDP: Good suggestion. Fixed

crates/validator/src/fts.rs
QuerthDP: Changed.

QuerthDP:
Changelog:
- changed `highlight_excerpts` not to take expected count argument (it is checked within index creation, and additional asserts if required)
- removed loops to be more explicit with the asserted excerpts

QuerthDP:
Changelog:
- removed some overlapping tests
- moved some tests to Tantivy's unit tests as the E2E path didn't excercise anything beyond
- added a request limit test

crates/validator/src/fts.rs
QuerthDP: Added a test for that. The limit is 2MB, default axum's limit.

crates/validator/src/fts.rs
QuerthDP: Audited that and indeed there were some overlaps. Removed some tests, other was moved to Tantivy's unit tests.

PR #563, opened 2026-08-19
vector-store: make diskann generic over the data provider and strategy

Akvear:
### Motivations
We want to keep both diskann inmem provider and a new hybrid provider that will keep vectors in ScyllaDB.
To avoid code duplication we can make the current diskann usage generic over a structs implementing a new `DiskannBackend` trait. This way we can keep the integration centralized and switch to different DataProviders easily, *if* they implement the correct Diskann traits.

Refs: VECTOR-831
Refs: VECTOR-830

### DIFF
--- a/crates/vector-store/src/vs_index/diskann/mod.rs
+++ b/crates/vector-store/src/vs_index/diskann/mod.rs
@@ -38,14 +38,15 @@ use diskann::graph::InplaceDeleteMethod;
 use diskann::graph::config::Builder;
 use diskann::graph::config::MaxDegree;
 use diskann::graph::config::defaults::ALPHA as DISKANN_DEFAULT_ALPHA;
+use diskann::graph::glue::DefaultPostProcessor;
+use diskann::graph::glue::InplaceDeleteStrategy;
+use diskann::graph::glue::InsertStrategy;
 use diskann::graph::index::SearchStats;
 use diskann::graph::search::Knn;
 use diskann::neighbor::Neighbor;
-use diskann_inmem::Context as InmemContext;
-use diskann_inmem::Provider as InmemProvider;
-use diskann_inmem::Strategy as InmemStrategy;
-use diskann_inmem::layers::Full;
-use diskann_inmem::provider::Config as InmemProviderConfig;
+use diskann::provider::DataProvider;
+use diskann::provider::Delete;
+use diskann::provider::SetElement;
 use diskann_vector::distance::Metric;
 use itertools::Itertools;
 use std::collections::BTreeMap;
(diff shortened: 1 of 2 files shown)


crates/vector-store/src/vs_index/diskann/mod.rs
```diff
@@ -446,17 +482,19 @@ async fn search(
...
+    backend: DB,
+    partition: &Partition<DB>,
     table: &RwLock<T>,
     params: &DiskannParams,
     embedding: Vector,
     limit: Limit,
 ) -> AnnR
 where
     T: TableSearch + Send + Sync + 'static,
+    DB: DiskannBackend,
```
QuerthDP: DB is too well correlated with "DataBase". Let's use something other to not make it confusing.

QuerthDP (dismissed):
Let's change the `DB`. Apart from that it looks all right to me.

Akvear:
Changes:
- changed generic DB to B to inform about the diskann backend

QuerthDP (approved):

crates/vector-store/src/vs_index/diskann/mod.rs
```diff
@@ -120,7 +146,8 @@ pub fn new_diskann(
     })
 }
 
-fn new(
+fn new<DP: DiskannBackend>(
+    backend: DP,
```
QuerthDP: nit: DP while everywhere else uses B

crates/vector-store/src/vs_index/diskann/mod.rs
```diff
@@ -150,13 +177,27 @@ fn new(
                         continue;
                     }
 
-                    let Some((partition, size, msg)) =
-                        preprocess(&mut partitions, &mut sizes, table.as_ref(), &params, msg)
-                    else {
+                    let Some((partition, size, msg)) = preprocess(
+                        backend.clone(),
```
QuerthDP: > Optional: this clones the generic backend for every incoming message, although preprocess only calls create_index(&self). Since Clone is not guaranteed to be cheap, pass &backend here and change preprocess to accept &DB; the separate clone used to own the dispatched task can remain.

I think Copilot is right here. Try to optimize it.

QuerthDP (changes requested):

crates/vector-store/src/vs_index/diskann/mod.rs
Akvear: Fixed

crates/vector-store/src/vs_index/diskann/mod.rs
Akvear: fixed

Akvear:
Changed:
- changed generic DP to B 
- passed backend by reference instead of clonning when calling `preprocess()`

crates/vector-store/src/vs_index/diskann/mod.rs
```diff
@@ -150,13 +177,27 @@ fn new(
...
+                        table.as_ref(),
+                        &params,
+                        msg,
+                    ) else {
                         continue;
                     };
 
-                    dispatch_task(partition, &table, &params, &worker, size, msg).await;
+                    dispatch_task(
+                        backend.clone(),
```
QuerthDP: The reference could be also used here right?

Akvear: It will need to be cloned in dispatch_task either way, because the worker requires ownership. We can Arc it and not have to worry about this.

Akvear:
Changes:
- Arc'ed the backend in `new()`, rc is bumped on every message

QuerthDP (approved):

PR #423, opened 2026-04-22
Refactor processing usearch operations

ewienik:
The PR tries to solve issue discovered on customer setup, where during rebuilding two indexes over CDC we observe a surge in search latency. The root cause seems to be actual low priority of ann search while indexes are built or modified in the meantime. The PR creates a benches tests for searching while indexes are built or modified using CDC.

The set of commits provides refactoring to the pipeline driven by benchmarks. Currently the changes are in parts which retrieve index data from engine or db_index, provide helpers for tokio runtime for hotpath scopes, setup channel buffering based on worker count.

The change introduces new module worker which improves management of usearch tasks. In the future it ought to manage also other long calculations.

The last commit refactors usearch capacity calculations - we don't need to call index usearch method as it complicates workflow.

In attachments there are logs from running benches at three points - base @ `benches: add new pipeline's tests`, buffers at `vector-store: increase channel buffers`, and worker at the last commit. There are clearly visible improvements between base and buffers, improvements between buffers and worker with one exception - clear search operations seems a little slower (~2%) - but there are still huge improvements from base and worker. As providing separate worker actor seems a better approach I would accept this difference.

Fixes: VECTOR-566

PR #564, opened 2026-08-21
vector-store: fix the stale rayon comment above main()

swasik:
The comment directly above `main()` told the reader that index creating and
querying run on a rayon `ThreadPool`:

```rust
// Index creating/querying is CPU bound task, so that vector-store uses rayon ThreadPool for them.
```

That has been false since May 2026.

## How it went stale

|  | Approach | Commit |
|---|---|---|
| 1 | USearch CPU tasks on Tokio threads | `adec0d9` |
| 2 | Dedicated rayon pool for blocking `add`/`remove` | `6935a83` (#294) |
| 3 | Worker actor with `spawn_blocking` | `27ddf4b` (2026-05-13, #423) |

The comment describes design 2; master runs design 3. **vector-store no longer
schedules any work on rayon** — no workspace crate depends on it. It does survive
in `Cargo.lock` transitively, via `tantivy` and `criterion`, so it is still linked
in and still doing work *inside tantivy*; it is simply not something this code
drives any more. (An earlier version of this description said "not a dependency of
the workspace", which was wrong — thanks @QuerthDP.)

## The replacement

Not "keeps it off the async runtime" either, which would have been a fresh
inaccuracy: `SpawnBlocking` runs the closure on the worker actor itself and only
escalates to a dedicated thread when `try_acquire_thread` finds every worker busy
and that thread free. The starvation guard is the exception, not the normal path —
thanks @Copilot for catching that. The comment now describes the scheduling.

It also **names no file and no symbol**, per @QuerthDP: a pointer to `worker.rs`
would be a second thing that can drift, which is exactly how the rayon sentence
got here.

The sentence about one thread being enough for network I/O is still accurate and
is kept verbatim.

Comment only, no behaviour change. `cargo fmt --check` clean.

Fixes: VECTOR-874

### DIFF
--- a/crates/vector-store/src/main.rs
+++ b/crates/vector-store/src/main.rs
@@ -25,7 +25,9 @@ fn dotenvy_to_std_var(key: &str) -> anyhow::Result<String> {
     Ok(dotenvy::var(key)?)
 }
 
-// Index creating/querying is CPU bound task, so that vector-store uses rayon ThreadPool for them.
+// Index creation and querying are CPU-bound tasks, so they are dispatched through the worker
+// actors. A blocking task is moved to a dedicated thread only when every worker actor is
+// already busy — so that the runtime does not get starved.
 // From the start there was no need (network traffic seems to be not so high) to support more than
 // one thread per network IO bound tasks.
 fn main() -> anyhow::Result<()> {


swasik:
Good catch — you're right, and the amended commit follows your reading of the code.

`Worker::SpawnBlocking` calls `f()` directly on the worker actor unless
`try_acquire_thread` succeeds, and that requires **both** `in_flow == workers`
(every worker busy) and the single dedicated thread being free
(`thread_in_flow == 1`). So the dedicated thread is the starvation guard, not the
normal path, and "keeps it off the async runtime" was wrong.

Reworded to describe the scheduling instead: dispatched through the worker actors;
a blocking task normally runs on the worker actor itself; only when every worker is
busy is one handed to the dedicated thread so the runtime is not starved. Also took
the grammar fix — "Index creation and querying are CPU-bound tasks".

crates/vector-store/src/main.rs
```diff
@@ -25,7 +25,8 @@ fn dotenvy_to_std_var(key: &str) -> anyhow::Result<String> {
     Ok(dotenvy::var(key)?)
 }
 
-// Index creating/querying is CPU bound task, so that vector-store uses rayon ThreadPool for them.
+// Index creating/querying is a CPU bound task, so vector-store keeps it off the async runtime:
+// see the worker actor in worker.rs, whose WorkerExt::spawn_blocking callers use for that work.
```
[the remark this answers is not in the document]
swasik: Fixed (see comment below).

crates/vector-store/src/main.rs
QuerthDP, line 1: > Rayon is not a dependency
of the workspace and this comment was the only mention of it left in the tree.

nit: it still it a dependency, check `Cargo.lock`.

crates/vector-store/src/main.rs
```diff
@@ -25,7 +25,9 @@ fn dotenvy_to_std_var(key: &str) -> anyhow::Result<String> {
     Ok(dotenvy::var(key)?)
 }
 
-// Index creating/querying is CPU bound task, so that vector-store uses rayon ThreadPool for them.
+// Index creation and querying are CPU-bound tasks, so they are dispatched through the worker
+// actors in worker.rs. A blocking task normally runs on the worker actor itself; only when every
```
QuerthDP: I wouldn't mention the code/file in the comment. It creates another dependency which may drift in the future like the rayon comment did.

QuerthDP (changes requested):

swasik:
Changelog for 88cfc93

- drop the `worker.rs` pointer from the comment — @QuerthDP's point stands, and it
  is the same failure mode as the rayon sentence: a reference that can drift. The
  comment now names no file and no symbol.
- correct the rayon claim. @QuerthDP is right that it is still in `Cargo.lock`, so
  "not a dependency of the workspace" was wrong. Precisely: no workspace crate
  depends on it, and it survives transitively via **tantivy** and **criterion** —
  so rayon is still linked in and still doing work inside tantivy, just not
  scheduled by vector-store. The commit message says that instead.

I have also corrected the same overstatement in the PR description above, in
VECTOR-874, and in the knowledge-base entry this came from.

<details>
<summary>range diff</summary>

```
1:  0d2e227 ! 1:  88cfc93 vector-store: fix the stale rayon comment above main()
    @@ Commit message
         vector-store: fix the stale rayon comment above main()
     
         The comment told the reader that index creating and querying run on a rayon
    -    ThreadPool. That stopped being true in May 2026, when the worker actor
    -    (27ddf4b) replaced the rayon pool introduced by #294. Rayon is not a dependency
    -    of the workspace and this comment was the only mention of it left in the tree.
    +    ThreadPool. That stopped being true in May 2026, when the worker actor (27ddf4b)
    +    replaced the rayon pool introduced by #294. vector-store no longer schedules any
    +    work on rayon: no workspace crate depends on it, and it survives in Cargo.lock
    +    only transitively, via tantivy and criterion.
     
         It sits directly above main(), so it is the first thing a newcomer reads when
    -    working out how the service keeps CPU-bound work off the async runtime — and it
    -    pointed them at a pool that does not exist instead of at worker.rs.
    +    working out how CPU-bound work is scheduled — and it pointed them at a pool this
    +    code no longer drives.
     
    -    Describe what worker.rs actually does. Saying the work is moved off the runtime
    -    would be its own inaccuracy: SpawnBlocking runs the closure on the worker actor
    -    itself, and only escalates to the dedicated thread when try_acquire_thread finds
    -    every worker busy and that thread free. The starvation guard is the exception,
    -    not the normal path, so the comment says so.
    +    Describe what the worker actors actually do. Saying the work is moved off the
    +    runtime would be its own inaccuracy: SpawnBlocking runs the closure on the worker
    +    actor itself, and only escalates to a dedicated thread when try_acquire_thread
    +    finds every worker busy and that thread free. The starvation guard is the
    +    exception, not the normal path, so the comment says so.
    +
    +    The comment names no file or symbol. A pointer to worker.rs would be a second
    +    thing that can drift out of date, which is how the rayon sentence got here.
     
         The sentence about one thread being enough for network IO is still accurate and
         is kept verbatim. Comment only; no behaviour change.
    @@ crates/vector-store/src/main.rs: fn dotenvy_to_std_var(key: &str) -> anyhow::Res
      
     -// Index creating/querying is CPU bound task, so that vector-store uses rayon ThreadPool for them.
     +// Index creation and querying are CPU-bound tasks, so they are dispatched through the worker
    -+// actors in worker.rs. A blocking task normally runs on the worker actor itself; only when every
    -+// worker is busy is one handed to the dedicated thread, so that the runtime is not starved.
    ++// actors. A blocking task normally runs on the worker actor itself; only when every worker is
    ++// busy is one handed to a dedicated thread, so that the runtime is not starved.
      // From the start there was no need (network traffic seems to be not so high) to support more than
      // one thread per network IO bound tasks.
      fn main() -> anyhow::Result<()> {
```

</details>

crates/vector-store/src/main.rs
```diff
@@ -25,7 +25,9 @@ fn dotenvy_to_std_var(key: &str) -> anyhow::Result<String> {
     Ok(dotenvy::var(key)?)
 }
 
-// Index creating/querying is CPU bound task, so that vector-store uses rayon ThreadPool for them.
+// Index creation and querying are CPU-bound tasks, so they are dispatched through the worker
+// actors. A blocking task normally runs on the worker actor itself; only when every worker is
+// busy is one handed to a dedicated thread, so that the runtime is not starved.
```
QuerthDP: nit: Rephrase to something more clear to parse: "A blocking task is moved to a dedicated thread only when every worker actor is already busy — so that the runtime does not get starved."

QuerthDP (dismissed):

swasik:
Changelog for 43f5dba

- take @QuerthDP's suggested wording for the second sentence verbatim. Leading with
  the condition ("only when every worker actor is already busy") reads better than
  stating the normal path first and then the exception, and it carries the same
  meaning in one clause instead of two.

Comment text now:

```rust
// Index creation and querying are CPU-bound tasks, so they are dispatched through the worker
// actors. A blocking task is moved to a dedicated thread only when every worker actor is
// already busy — so that the runtime does not get starved.
```

<details>
<summary>range diff</summary>

```
1:  88cfc93 ! 1:  43f5dba vector-store: fix the stale rayon comment above main()
    @@ Commit message
         runtime would be its own inaccuracy: SpawnBlocking runs the closure on the worker
         actor itself, and only escalates to a dedicated thread when try_acquire_thread
         finds every worker busy and that thread free. The starvation guard is the
    -    exception, not the normal path, so the comment says so.
    +    exception, not the normal path, so the comment leads with the condition that
    +    reaches it.
     
         The comment names no file or symbol. A pointer to worker.rs would be a second
         thing that can drift out of date, which is how the rayon sentence got here.
    @@ crates/vector-store/src/main.rs: fn dotenvy_to_std_var(key: &str) -> anyhow::Res
      
     -// Index creating/querying is CPU bound task, so that vector-store uses rayon ThreadPool for them.
     +// Index creation and querying are CPU-bound tasks, so they are dispatched through the worker
    -+// actors. A blocking task normally runs on the worker actor itself; only when every worker is
    -+// busy is one handed to a dedicated thread, so that the runtime is not starved.
    ++// actors. A blocking task is moved to a dedicated thread only when every worker actor is
    ++// already busy — so that the runtime does not get starved.
      // From the start there was no need (network traffic seems to be not so high) to support more than
      // one thread per network IO bound tasks.
      fn main() -> anyhow::Result<()> {
```

</details>

crates/vector-store/src/main.rs
swasik: You are right - this is transitive dependency. The claim is now softened.

crates/vector-store/src/main.rs
swasik: Good point. Fixed.

crates/vector-store/src/main.rs
swasik: Done.

QuerthDP (approved):

PR #566, opened 2026-08-21
vector-store: return 400 for unparsable FTS queries

QuerthDP:
The BM25 endpoint mapped every search failure to 500, so a query the parser rejects came back as an internal error even though the query is the caller's input.

`QueryError` already distinguishes a bad query from an internal failure, but nothing consumed it. Re-export it from `fts_index` and have `post_index_bm25` downcast on it to pick the status code, mirroring how `post_index_ann` downcasts on `vs_index::Error`. Other failures keep returning 500.

Widen the documented 400 cause for the endpoint accordingly and regenerate the OpenAPI specification.

Fixes: [VECTOR-744](https://scylladb.atlassian.net/browse/VECTOR-744)

[VECTOR-744]: https://scylladb.atlassian.net/browse/VECTOR-744?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/openapi.json
+++ b/api/openapi.json
@@ -264,7 +264,7 @@
             }
           },
           "400": {
-            "description": "Bad request. Possible causes: malformed input, or missing required fields.",
+            "description": "Bad request. Possible causes: malformed input, unparsable query, or missing required fields.",
             "content": {
               "application/json": {
                 "schema": {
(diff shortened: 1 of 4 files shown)


QuerthDP:
Based on https://github.com/scylladb/vector-store/pull/558, waiting for merge.

knowack1 (approved):
Thanks for jumping on this problem.

PR #567, opened 2026-08-24
vector-store: create a skeleton for the scylla diskann provider

Akvear:
This PR introduces a skeleton that will make be a base for our own implementation of the DiskANN `DataProvider` trait. 

This PR is best to read alongside 2 .md files in https://github.com/scylladb/vector-search-docs/pull/25:
- one explaining what a `DataProvider` is and what needs to be implemented for one
- one explaining the choice for the implementation 

Refs: VECTOR-830
Refs: VECTOR-831

### DIFF
--- a/crates/vector-store/src/table/primary_id.rs
+++ b/crates/vector-store/src/table/primary_id.rs
@@ -11,6 +11,9 @@ use tracing::info;
 /// PrimaryId is a unique identifier for a row in the table. It consists of an index and an
 /// epoch. The index is used to access the row in the column vectors, while the epoch is used
 /// to determine if the row has been updated.
+///
+/// [`bytemuck::Pod`] and the `repr(transparent)` layout let a `PrimaryId` be used
+/// as a DiskANN vector id.
 #[derive(
     Copy,
     Clone,
@@ -20,10 +23,14 @@ use tracing::info;
     PartialOrd,
     Ord,
     Hash,
+    bytemuck::Pod,
+    bytemuck::Zeroable,
     derive_more::AsRef,
+    derive_more::Display,
     derive_more::From,
     derive_more::Into,
 )]
+#[repr(transparent)]
 pub struct PrimaryId(u64);
 
 const _: () = assert!(
@@ -33,7 +40,9 @@ const _: () = assert!(
 
 impl PrimaryId {
     const EPOCH_SHIFT: usize = (mem::size_of::<u64>() - mem::size_of::<Epoch>()) * 8;
-    const MAX: u64 = !((Epoch::MAX as u64) << Self::EPOCH_SHIFT);
+
+    const RESERVED: u64 = !((Epoch::MAX as u64) << Self::EPOCH_SHIFT);
+    const MAX: u64 = Self::RESERVED - 1;
 
     pub(super) fn try_new(idx: usize, epoch: Epoch) -> anyhow::Result<Self> {
         if idx as u64 > Self::MAX {
@@ -45,7 +54,7 @@ impl PrimaryId {
     }
 
     pub(super) fn new_epoch(mut self, epoch: Epoch) -> Self {
-        self.0 &= Self::MAX;
+        self.0 &= Self::RESERVED;
         self.0 |= (epoch.0 as u64) << Self::EPOCH_SHIFT;
         self
     }
@@ -53,11 +62,21 @@ impl PrimaryId {
     pub(super) fn epoch(&self) -> Epoch {
         Epoch((self.0 >> Self::EPOCH_SHIFT) as u16)
     }
+
+    pub(crate) const fn new_reserved() -> Self {
+        Self(Self::RESERVED)
+    }
+}
+
+impl Default for PrimaryId {
+    fn default() -> Self {
+        Self(Self::RESERVED)
+    }
 }
 
 impl Idx for PrimaryId {
     fn idx(&self) -> usize {
-        (self.0 & Self::MAX) as usize
+        (self.0 & Self::RESERVED) as usize
     }
 }
 
--- a/crates/vector-store/src/vs_index/diskann/scylla.rs
+++ b/crates/vector-store/src/vs_index/diskann/scylla.rs
@@ -1,0 +1,957 @@
+/*
+ * Copyright 2026-present ScyllaDB
+ * SPDX-License-Identifier: LicenseRef-ScyllaDB-Source-Available-1.1
+ */
+
+//! A DiskANN backend that keeps the graph in RAM but not the vectors.
+//!
+//! The vectors we index already live in ScyllaDB. This backend fetches these
+//! vectors instead of keeping a copy in RAM; the graph — adjacency lists and
+//! node liveness — stays in RAM.
+//!
+//! # Identity ids
+//!
+//! We use [`crate::PrimaryId`] *as* the internal id: `to_internal_id` and
+//! `to_external_id` are the identity, and there is no id map, no slot allocator
+//! and no free list.
+//!
+//! A `PrimaryId` is a table row slot ([`PrimaryId::idx`]) plus a 16-bit epoch
+//! bumped on every update of that row, so identity ids are sparse and the graph
+//! is a `BTreeMap` keyed by the whole id. Every epoch of a row is its own key,
+//! and that is what makes the ordering safe.
+//!
+//! Ids are not reclaimed, and they are not reuse-free either: the epoch wraps, so
+//! after 65,536 updates to one row the same id is re-issued for different vector
+//! data while this graph may still hold edges to it, but it is very unlikely.
+//! That costs recall, not correctness.
+//!
+//! # What lives in RAM
+//!
+//! * The graph: one [`Node`] per vector, holding its adjacency list.
+//! * Start points, vectors included. Their ids are [`PrimaryId::RESERVED`], not
+//!   backed by a row, so they can never be read back from ScyllaDB.
+//! * Vectors of in-flight inserts. Back-edge pruning asks for the new vector
+//!   before the insert finishes. The row is already in ScyllaDB by then — every
+//!   `AddVector` comes from a full scan or from CDC, both reads of the base
+//!   table — so this only saves a point read. [`InflightGuard`] removes it on
+//!   both success and failure.
+//! * Per-operation working sets: candidates fetched for one prune, dropped when
+//!   it ends.
+
+use super::DiskannBackend;
+use super::DiskannParams;
+use crate::PrimaryId;
+use crate::Vector;
+use anyhow::Context as _;
+use anyhow::bail;
+use async_trait::async_trait;
+use diskann::ANNError;
+use diskann::ANNResult;
+use diskann::default_post_processor;
+use diskann::graph::AdjacencyList;
+use diskann::graph::DiskANNIndex;
+use diskann::graph::SearchOutputBuffer;
+use diskann::graph::glue;
+use diskann::graph::workingset;
+use diskann::graph::workingset::map::Project;
+use diskann::graph::workingset::map::Ref;
+use diskann::neighbor::Neighbor;
+use diskann::provider;
+use diskann::provider::DataProvider;
+use diskann::provider::DefaultContext;
+use diskann::provider::Delete;
+use diskann::provider::ElementStatus;
+use diskann::provider::HasId;
+use diskann::provider::NeighborAccessor;
+use diskann::provider::NeighborAccessorMut;
+use diskann::provider::SetElement;
+use diskann::utils::TypeStr;
+use diskann::utils::VectorRepr;
+use diskann_utils::Reborrow;
+use diskann_vector::PreprocessedDistanceFunction;
+use diskann_vector::contains::ContainsSimd;
+use diskann_vector::distance::Metric;
+use std::collections::BTreeMap;
+use std::collections::HashMap;
+use std::collections::btree_map::Entry;
+use std::fmt::Debug;
+use std::future::Future;
+use std::sync::Arc;
+use std::sync::RwLock;
+use tracing::warn;
+
+/// Lets a [`Vector`] be stored in DiskANN's working set and handed back as a
+/// `&[f32]` view. The crate provides these for `Vec<T>` and `Box<T>` only, so
+/// a newtype has to supply its own.
+impl Project<Ref<[f32]>> for Vector {
+    fn project(&self) -> &[f32] {
+        self.as_slice()
+    }
+}
+/// Required of [`Delete::DeleteElementGuard`]: the guard's lifetime scopes the
+/// borrowed element.
+impl<'this> Reborrow<'this> for Vector {
+    type Target = &'this [f32];
+
+    fn reborrow(&'this self) -> Self::Target {
+        self.as_slice()
+    }
+}
+
+const START_ID: PrimaryId = PrimaryId::new_reserved();
+
+impl TypeStr for PrimaryId {
+    fn type_str() -> &'static str {
+        u64::type_str()
+    }
+}
+
+impl ContainsSimd for PrimaryId {
+    fn contains_simd(vector: &[Self], target: Self) -> bool {
+        u64::contains_simd(bytemuck::cast_slice(vector), u64::from(target))
+    }
+}
+
+/// The source of vector data behind a [`ScyllaProvider`].
+///
+/// One source serves every per-partition index. Results are index-aligned with
+/// the requested ids: `None` is a candidate with no vector to score, skipped at
+/// a cost in recall. A source that cannot answer returns `Err`.
+#[async_trait]
+pub(super) trait VectorSource: Debug + Send + Sync + 'static {
+    async fn get(&self, ids: &[PrimaryId]) -> anyhow::Result<Vec<Option<Vector>>>;
+}
+
+#[derive(Clone)]
+pub(super) struct ScyllaBackend {
+    strategy: ScyllaStrategy,
+    context: DefaultContext,
+    source: Arc<dyn VectorSource>,
+}
+
+impl ScyllaBackend {
+    #[allow(dead_code, reason = "wired into DiskannIndexFactory in a follow-up")]
+    pub(super) fn new(source: Arc<dyn VectorSource>) -> Self {
+        Self {
+            strategy: ScyllaStrategy::default(),
+            context: DefaultContext,
+            source,
+        }
+    }
+}
+
+impl DiskannBackend for ScyllaBackend {
+    type Provider = ScyllaProvider;
+    type Strategy = ScyllaStrategy;
+
+    fn create_index(
+        &self,
+        params: &DiskannParams,
+        start_point: &[f32],
+    ) -> anyhow::Result<DiskANNIndex<Self::Provider>> {
+        let provider = ScyllaProvider::new(
+            start_point,
+            usize::from(params.dim.0),
+            params.metric,
+            params.config.max_degree().get(),
+            Arc::clone(&self.source),
+        )
+        .context("failed to create ScyllaProvider")?;
+
+        Ok(DiskANNIndex::new(params.config.clone(), provider, None))
+    }
+
+    fn strategy(&self) -> &Self::Strategy {
+        &self.strategy
+    }
+
+    fn context(&self) -> &DefaultContext {
+        &self.context
+    }
+}
+
+#[derive(Debug)]
+enum Node {
+    Live(AdjacencyList<PrimaryId>),
+    Dead(AdjacencyList<PrimaryId>),
+}
+
+impl Node {
+    fn new(capacity: usize) -> Self {
+        Self::Live(AdjacencyList::with_capacity(capacity))
+    }
+
+    fn is_live(&self) -> bool {
+        matches!(self, Self::Live(_))
+    }
+
+    fn neighbors(&self) -> &AdjacencyList<PrimaryId> {
+        let (Self::Live(neighbors) | Self::Dead(neighbors)) = self;
+        neighbors
+    }
+
+    fn neighbors_mut(&mut self) -> &mut AdjacencyList<PrimaryId> {
+        let (Self::Live(neighbors) | Self::Dead(neighbors)) = self;
+        neighbors
+    }
+
+    fn mark_deleted(&mut self) {
+        if let Self::Live(neighbors) = self {
+            *self = Self::Dead(std::mem::take(neighbors));
+        }
+    }
+}
+
+/// Vectors held in RAM only for the duration of an insert.
+type Inflight = Arc<RwLock<BTreeMap<PrimaryId, Vector>>>;
+
+/// The graph, keyed by internal id.
+type Nodes = Arc<RwLock<BTreeMap<PrimaryId, Node>>>;
+
+/// A DiskANN provider whose vector reads go to a [`VectorSource`].
+#[derive(Debug)]
+pub(super) struct ScyllaProvider {
+    nodes: Nodes,
+    inflight: Inflight,
+    start: Vector,
+    start_neighbors: RwLock<AdjacencyList<PrimaryId>>,
+    source: Arc<dyn VectorSource>,
+    max_degree: usize,
+    dim: usize,
+    metric: Metric,
+}
+
+impl ScyllaProvider {
+    pub(super) fn new(
+        start_point: &[f32],
+        dim: usize,
+        metric: Metric,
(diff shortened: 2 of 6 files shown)


crates/vector-store/src/vs_index/diskann/scylla.rs
```diff
@@ -0,0 +1,817 @@
...
+                "vector source returned {} results for {} ids; \
+                 implementations must return one entry per id, in order",
+                fetched.len(),
+                wanted.len()
+            )));
+        }
+
+        for (position, vector) in slots.into_iter().zip(fetched) {
+            out[position] = vector;
+        }
```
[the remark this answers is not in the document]
Akvear: Dimension validation is on the vector-store side. We can also have a guard here but in real scenarios it should not be necessary.

crates/vector-store/src/vs_index/diskann/scylla.rs
```diff
@@ -0,0 +1,817 @@
...
+        &self,
+        params: &DiskannParams,
+        start_point: &[f32],
+    ) -> anyhow::Result<DiskANNIndex<Self::Provider>> {
+        let provider = ScyllaProvider::new(
+            start_point,
+            usize::from(params.dim.0),
+            params.metric,
+            params.config.max_degree().get(),
+            Arc::clone(&self.source),
```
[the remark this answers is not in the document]
Akvear: This is correct, we do not need max_points in this new provider. There is a decision to be made about what MAX_POINTS env var should do for this provider. The memory is still checked on the VS side with `check_memory_allocation` before any VsIndex msg.

crates/vector-store/src/vs_index/diskann/scylla.rs
```diff
@@ -0,0 +1,817 @@
...
+const START_ID: u64 = u64::MAX;
+
+/// The source of vector data behind a [`ScyllaProvider`].
+///
+/// Results are index-aligned with the requested ids. `None` means "unavailable,
+/// skip this candidate": both `expand_beam` and `fill` tolerate it, at a cost in
+/// recall rather than correctness.
+#[async_trait]
+pub(super) trait VectorSource: std::fmt::Debug + Send + Sync + 'static {
+    async fn get(&self, ids: &[PrimaryId]) -> Vec<Option<Box<[f32]>>>;
```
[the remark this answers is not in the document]
Akvear: It is documented that missing rows are traded for lower recall instead of returning an error. Whether this is an acceptable approach is a different thing.

crates/vector-store/src/vs_index/diskann/scylla.rs
```diff
@@ -0,0 +1,817 @@
...
+//!
+//! # Identity ids
+//!
+//! [`crate::PrimaryId`] is a `u64`, so we use it *as* the internal id:
+//! `to_internal_id` and `to_external_id` are casts, and there is no id map, no
+//! slot allocator and no free list. It also removes a hazard rather than
+//! managing one: a slot-recycling provider can hand a stale id found in some
+//! adjacency list to a *different* vector, which `diskann-inmem` guards against
+//! with epoch-based reclamation and `diskann-garnet` accepts as a recall
+//! penalty. A `PrimaryId` embeds an epoch and is never reused, so here a stale
```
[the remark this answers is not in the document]
Akvear: While true, this doesn't necessarily break anything, just costs recall. This already has precedent, with diskann-garnet and diskann-bftree. This is a design approach rather than an oversight.

crates/vector-store/src/vs_index/diskann/scylla.rs
```diff
@@ -0,0 +1,817 @@
...
+    inflight: Inflight,
+    start: Box<[f32]>,
+    source: Arc<dyn VectorSource>,
+    max_degree: usize,
+    dim: usize,
+    metric: Metric,
+}
+
+impl ScyllaProvider {
+    pub(super) fn new(
```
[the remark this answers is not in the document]
Akvear: TBD

crates/vector-store/src/vs_index/diskann/scylla.rs
```diff
@@ -0,0 +1,817 @@
...
+    inflight: Inflight,
+    id: u64,
+}
+
+impl provider::Guard for InflightGuard {
+    type Id = u64;
+
+    async fn complete(self) {
+        // The `Drop` impl does the work.
+    }
```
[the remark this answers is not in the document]
Akvear: Added tracking

crates/vector-store/src/vs_index/diskann/scylla.rs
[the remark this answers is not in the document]
Akvear: We could do something like in usearch with dynamic changes to the partition size, or just leave the sizing to the DashMap.

A different approach would be to make our own id translation instead of identity. Then the nodes could be kept in an fixed, or doubling in size when needed array.

Akvear:
In order to prove this solution has potential I used a 'MockDb' (HashMap) and added simple tests validating some implementation choices. I also temporarily changed the backend in mod.rs to the new one and ran the 11 integration tests that are the standard for diskann in the CI. All tests were green so it is pretty safe to assume that there are no glaring flaws in this implementation.

crates/vector-store/src/vs_index/diskann/scylla.rs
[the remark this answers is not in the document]
Akvear: Done in #575

crates/vector-store/src/table/primary_id.rs
```diff
@@ -50,14 +50,18 @@ impl PrimaryId {
...
     }
 
-    pub(super) fn epoch(&self) -> Epoch {
+    pub(crate) fn epoch(&self) -> Epoch {
         Epoch((self.0 >> Self::EPOCH_SHIFT) as u16)
     }
+
+    pub(crate) fn idx(&self) -> usize {
+        (self.0 & Self::MAX) as usize
+    }
```
ewienik, line 74: IMHO Idx::idx trait should be enough, no need to add additional method to PrimaryId. Can you check it?

crates/vector-store/src/vs_index/diskann/scylla.rs
```diff
@@ -0,0 +1,946 @@
...
+}
+
+#[derive(Debug)]
+struct Node {
+    epoch: Epoch,
+    live: bool,
+    neighbors: AdjacencyList<u64>,
+}
+
```
ewienik, line 178: What about such enum?
```
enum Node {
     Live { epoch, neighbors },
     Dead { neighbors },
     None,
}
```
If so `Nodes = Arc<RwLock<Vec<Node>>>`

crates/vector-store/src/vs_index/diskann/scylla.rs
```diff
@@ -0,0 +1,946 @@
...
+            return std::future::ready(Ok(()));
+        }
+
+        let mut nodes = self.nodes.write().unwrap();
+        if let Some(node) = node_mut(&mut nodes, id) {
+            if neighbors.is_empty() && !node.live {
+                // `drop_adj_list` at the end of an in-place delete. This is
+                // the only place a completed delete clears its slot; one that
+                // never gets here is a tombstone, and nothing else clears it.
+                nodes[PrimaryId::from(id).idx()] = None;
```
ewienik: It could panic if for some reason id is wrong

crates/vector-store/src/vs_index/diskann/scylla.rs
```diff
@@ -0,0 +1,946 @@
...
+use std::fmt::Debug;
+use std::future::Future;
+use std::sync::Arc;
+use std::sync::RwLock;
+use tracing::warn;
+
+/// Lets a [`Vector`] be stored in DiskANN's working set and handed back as a
+/// `&[f32]` view. The crate provides these for `Vec<T>` and `Box<T>` only, so
+/// a newtype has to supply its own.
+impl workingset::map::Project<workingset::map::Ref<[f32]>> for Vector {
```
ewienik: nit: use workingset::map::Project and use workingset::map::Ref

crates/vector-store/src/vs_index/diskann/scylla.rs
```diff
@@ -0,0 +1,946 @@
...
+ */
+
+//! A DiskANN backend that keeps the graph in RAM but not the vectors.
+//!
+//! The vectors we index already live in ScyllaDB, so the default
+//! [`InmemBackend`](super::inmem::InmemBackend) stores every one of them a
+//! second time. This backend removes that copy: the graph — adjacency lists and
+//! node liveness — stays in RAM, while vector data is fetched on demand through
+//! [`VectorSource`].
+//!
```
ewienik: nit: Consider rewording without comparison to Inmem - IIUC it was a part of decision to implement separate Provider. Such comparison could be located in commit description.

crates/vector-store/src/vs_index/diskann/scylla.rs
```diff
@@ -0,0 +1,946 @@
...
+//! operation finds the slot no longer holds its epoch and does nothing. In-edges
+//! written before the update keep pointing at the old id, and they are skipped
+//! for the same reason — the recall cost of a changed row, without a read of the
+//! base table to discover it. Epoch wrap-around (65,536 updates to one row)
+//! aliases as before, but it now needs the slot to still hold that exact epoch.
+//!
+//! Neighbor ids are still 8 bytes rather than the 4 of `diskann-inmem`'s
+//! `max_points * (max_degree + 1) * 4` arena. Narrowing them is a separate
+//! change, and one that would not need an id map either: the row slot already
+//! fits a `u32` at any workable `max_points`.
```
ewienik: nit: Consider rewording without comparison to Inmem - IIUC it was a part of decision to implement separate Provider. Such comparison could be located in commit description.

crates/vector-store/src/vs_index/diskann/scylla.rs
```diff
@@ -0,0 +1,946 @@
...
+            nodes: &self.nodes,
+            start_neighbors: &self.start_neighbors,
+            max_degree: self.max_degree,
+        }
+    }
+}
+
+impl DataProvider for ScyllaProvider {
+    type Context = DefaultContext;
+    type InternalId = u64;
```
ewienik: Can PrimaryId be used here? If so, define needed trait for PrimaryId.

crates/vector-store/src/vs_index/diskann/scylla.rs
```diff
@@ -0,0 +1,946 @@
...
+    nodes
+        .get_mut(id.idx())?
+        .as_mut()
+        .filter(|node| node.epoch == id.epoch())
+}
+
+/// Is `epoch` at least as recent as `current`?
+fn is_recent(epoch: Epoch, current: Epoch) -> bool {
+    epoch.as_ref().wrapping_sub(*current.as_ref()) < 1 << (u16::BITS - 1)
+}
```
ewienik: It seems wrong as Epoch has no guarantee of being monotonic. There is no way to check if the epoch from one PrimaryId::Epoch is recent to other.

crates/vector-store/src/vs_index/diskann/scylla.rs
```diff
@@ -0,0 +1,946 @@
...
+
+impl Drop for InflightGuard {
+    fn drop(&mut self) {
+        self.inflight.write().unwrap().remove(&self.id);
+
+        if self.armed {
+            let mut nodes = self.nodes.write().unwrap();
+            if node(&nodes, self.id).is_some() {
+                nodes[PrimaryId::from(self.id).idx()] = None;
+            }
```
ewienik: Consider using node_mut

crates/vector-store/src/vs_index/diskann/scylla.rs
```diff
@@ -0,0 +1,946 @@
...
+                self.dim
+            )));
+        }
+
+        let internal = u64::from(*id);
+        if internal == START_ID {
+            return Err(ANNError::message(format!(
+                "id {internal} is reserved for the start point"
+            )));
+        }
```
ewienik, line 394: The user can remove starting point - then we need to change it.

ewienik (changes requested):
I will continue with review

crates/vector-store/src/vs_index/diskann/scylla.rs
Akvear: pub trait VectorId:
    std::cmp::Eq
    + std::cmp::PartialEq
    + std::cmp::Ord
    + std::hash::Hash
    + TypeStr
    + Copy
    + Sized
    + Send
    + Sync
    + std::fmt::Debug
    + std::fmt::Display
    + Default
    + ContainsSimd
    + bytemuck::Pod
    + static
    
These are the required traits, are we OK with these being implemented for PrimaryId. We could alternatively add a newtype for the InternalId.

crates/vector-store/src/vs_index/diskann/scylla.rs
Akvear: How should we choose the new starting point? The way it is now, the starting point is not part of the results either way. It is a frozen point from which the beam search starts, it does not really matter whether it was deleted.

Akvear:
Changes:
- reverted to keeping the graph as a map instead of a vector
- revised module doc comments
- changed Node to be enum with Live and Dead variants (same can't be done for the Guard because of the Drop trait)
-  added missing imports

crates/vector-store/src/vs_index/diskann/scylla.rs
ewienik: Default seems problematic :-(

crates/vector-store/src/vs_index/diskann/scylla.rs
ewienik: As the vector value for starting point is copied and only neighbors can be updated then user can remove the origin vector for starting point.

crates/vector-store/src/vs_index/diskann/scylla.rs
ewienik: Default can be provided by PrimaryId as current MAX and MAX decremented by 1 similar to:

```
const DEFAULT: u64 = !((Epoch::MAX as u64) << Self::EPOCH_SHIFT);
const MAX: u64 = Self::DEFAULT - 1;                              

impl Default for PrimaryId {
    fn default() -> Self {  
        Self(Self::DEFAULT) 
    }                       
}                           
```

Starting point could be declared by PrimaryId::default()

crates/vector-store/src/vs_index/diskann/scylla.rs
```diff
@@ -0,0 +1,925 @@
...
+
+/// Cleans up after an insert: the in-flight vector copy always, and the node
+/// `set_element` created unless the insert completed.
+#[derive(Debug)]
+pub(super) struct InflightGuard {
+    inflight: Inflight,
+    nodes: Nodes,
+    id: u64,
+    armed: bool,
+}
```
ewienik: Check if it is possible to refactor this as enum

crates/vector-store/src/vs_index/diskann/scylla.rs
QuerthDP: Why MAX and not 0?

crates/vector-store/src/vs_index/diskann/scylla.rs
Akvear: I think nodes + armed can be minimized into an enum but the entire struct can't since Drop is activated when trying to change from Self::Armed to Self::Disarmed.

Akvear:
Changes:
- added rollback enum instead of the armed flag in InflightGuard
- replaced InternalId = u64 with InternalId = PrimaryId, which required implementing a few traits on PrimaryId

crates/vector-store/src/vs_index/diskann/scylla.rs
ewienik: The version with forget is now wrong - it forgets internal values also. After my investigation it seems that we need unsafe + ManuallyDrop or indirection. I'd prefere indirection. What about something along the lines of:
```
pub(super) struct InflightGuard(Option<(Inflight, PrimaryId, Nodes)>);                 
                                                                                       
impl provider::Guard for InflightGuard {                                               
    type Id = PrimaryId;                                                               
                                                                                       
    async fn complete(mut self) {                                                      
        let (inflight, id, _) = self.0.take().expect("InflightGuard already consumed");
        inflight.write().unwrap().remove(&id);                                         
    }                                                                                  
                                                                                       
    fn id(&self) -> Self::Id {                                                         
        self.0.as_ref().expect("InflightGuard already consumed").1                     
    }                                                                                  
}                                                                                      
                                                                                       
impl Drop for InflightGuard {                                                          
    fn drop(&mut self) {                                                               
        if let Some((inflight, id, nodes)) = self.0.take() {                           
            inflight.write().unwrap().remove(&id);                                     
            nodes.write().unwrap().remove(&id);                                        
        }                                                                              
    }                                                                                  
}                                                                                      

```
panic should never fired. Instead of tuple there could be some internal structure for better visibility

crates/vector-store/src/table/primary_id.rs
```diff
@@ -33,7 +40,10 @@ const _: () = assert!(
 
 impl PrimaryId {
     const EPOCH_SHIFT: usize = (mem::size_of::<u64>() - mem::size_of::<Epoch>()) * 8;
-    const MAX: u64 = !((Epoch::MAX as u64) << Self::EPOCH_SHIFT);
+
+    const DEFAULT: u64 = !((Epoch::MAX as u64) << Self::EPOCH_SHIFT);
+    const MAX: u64 = Self::DEFAULT - 1;
+    pub(crate) const RESERVED: PrimaryId = Self(Self::DEFAULT);
```
ewienik: I think now that RESERVED name is better for const than DEFAULT, but it shouldn't be visible outside. What about this?

```
const RESERVED: u64 = !((Epoch::MAX as u64) << Self::EPOCH_SHIFT);
const MAX: u64 = Self::RESERVED - 1;                              

pub(crate) const fn new_reserved() -> Self {
    Self(Self::RESERVED)
}

fn default() -> Self { 
    Self::reserved()
}                      

```

ewienik (changes requested):

crates/vector-store/src/table/primary_id.rs
```diff
@@ -45,19 +54,29 @@ impl PrimaryId {
...
     }
+
+    pub(crate) const fn new_reserved() -> Self {
+        Self(Self::RESERVED)
+    }
+}
+
+impl Default for PrimaryId {
+    fn default() -> Self {
+        Self(Self::MAX)
```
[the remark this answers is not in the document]
Akvear: @ewienik should the default be the `RESERVED` slot or the `MAX` value?

Akvear:
Changes: 
- renamed DEFAULT to RESERVED in `PrimaryId`
- updated InflightGuard according to discussion

crates/vector-store/src/table/primary_id.rs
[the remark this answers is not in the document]
ewienik: I think default() should return the same as RESERVED -> for the reason that default() mustn't be connected to the real PrimaryId - in case we use it for some reason (creating collection or others), it won't be pointing at an actual primary key.

crates/vector-store/src/vs_index/diskann/scylla.rs
```diff
@@ -0,0 +1,934 @@
...
+#[async_trait]
+pub(super) trait VectorSource: Debug + Send + Sync + 'static {
+    async fn get(&self, ids: &[PrimaryId]) -> anyhow::Result<Vec<Option<Vector>>>;
+}
+
+/// A [`VectorSource`] that has nothing to offer.
+///
+/// Placeholder until the ScyllaDB read path is wired up.
+#[derive(Debug)]
+#[allow(dead_code, reason = "replaced by the real read path in a follow-up")]
```
QuerthDP: Is it really a dead code? And if so, why to include that now? Wouldn't it be substituted by real `VectorSource` in a follow-up?

ewienik (dismissed):
I thing PrimaryId::default() should return RESERVED

crates/vector-store/src/vs_index/diskann/scylla.rs
```diff
@@ -0,0 +1,934 @@
...
+impl ScyllaProvider {
+    fn status(&self, id: PrimaryId) -> ElementStatus {
+        if self.is_live(id) {
+            ElementStatus::Valid
+        } else {
+            ElementStatus::Deleted
+        }
+    }
+}
+/// What an [`InflightGuard`] needs to undo the insert it guards.
```
QuerthDP, line 489: missing newline

crates/vector-store/src/vs_index/diskann/scylla.rs
```diff
@@ -0,0 +1,934 @@
...
+    nodes: &'a RwLock<BTreeMap<PrimaryId, Node>>,
+    start_neighbors: &'a RwLock<AdjacencyList<PrimaryId>>,
+    max_degree: usize,
+}
+
+impl NodeAccessor<'_> {
+    /// Replace an adjacency list with `neighbors`, capped at the max degree.
+    fn overwrite(list: &mut AdjacencyList<PrimaryId>, neighbors: &[PrimaryId], max_degree: usize) {
+        list.clear();
+        list.extend_from_slice(neighbors);
```
QuerthDP, line 533: Why clear and extend instead of just assignment?

crates/vector-store/src/vs_index/diskann/scylla.rs
```diff
@@ -0,0 +1,934 @@
...
+pub(super) struct ScyllaPruneAccessor<'a> {
+    provider: &'a ScyllaProvider,
+    adjacency: NodeAccessor<'a>,
+    set: WorkingSet,
+    distance: <f32 as VectorRepr>::Distance,
+}
+
+impl<'a> ScyllaPruneAccessor<'a> {
+    fn new(provider: &'a ScyllaProvider, capacity: usize) -> Self {
+        let set = workingset::map::Builder::new(workingset::map::Capacity::Default).build(capacity);
```
QuerthDP, line 736: Why not 
```suggestion
        let set = workingset::map::Builder::new(workingset::map::Capacity::Some(capacity));
```

According to the code:
```
pub enum Capacity {
    /// No capacity — clear the map on every fill. Equivalent to `Some(0)`.
    None,
    /// Use the default capacity provided by the caller (typically `max_occlusion_size`).
    Default,
    /// Use a specific capacity.
    Some(usize),
    /// Unbounded — never evict entries.
    Unbounded,
}
```
I would guess that the `Some` is intended use for us here.

crates/vector-store/src/vs_index/diskann/scylla.rs
```diff
@@ -0,0 +1,934 @@
...
+        let mut pending: HashMap<PrimaryId, Vector> = if misses.is_empty() {
+            HashMap::new()
+        } else {
+            let loaded = self.provider.load(&misses).await.unwrap_or_else(|err| {
+                warn!(
+                    "prune: scoring {} candidates as unavailable: {err}",
+                    misses.len(),
+                );
+                vec![None; misses.len()]
+            });
```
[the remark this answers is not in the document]
QuerthDP: Sounds reasonable. @Akvear what do you think?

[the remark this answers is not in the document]
Akvear: Added enum to distinguish between `load()` failing during insert or delete. An insert should return an error, because the newly created node can be separated from the rest of the graph and become useless. While a delete marks the node as dead before any fallible operations, erroring during the delete means that we will not clean up after this node and just leave it marked as dead until an epoch wrap-around.

Akvear:
Changes:
- changed PrimaryId::default() to RESERVED
- added enum to distinguish when to return error and when to ignore a vector source failure

ewienik (dismissed):
Thanks for changes

QuerthDP (dismissed):
Looks promising. Please consider my comments.

crates/vector-store/src/vs_index/diskann/scylla.rs
Akvear: That's the way it is upstream in DiskANN

crates/vector-store/src/vs_index/diskann/scylla.rs
Akvear: It's basically the same thing and upstream providers use Capacity::Default.

Akvear:
changes:
- removed the Unavailable Vector Source implementation which also fixed missing newline

QuerthDP (approved):

PR #568, opened 2026-08-24
vector-store: fix an index staying undiscovered until the next schema change

QuerthDP:
`monitor_indexes` reads the cluster schema version once per tick and skips the whole discovery round while that version has not changed. The version is stored as soon as it is read, so a round that completes successfully while dropping an index leaves nothing to retry: that index is never looked at again until an unrelated schema change bumps the version.

`get_indexes` had exactly that shape. `system_schema` is node-local, so the listing of all custom indexes and the per-index follow-up reads (the index options row, the target column type) can be served by different coordinators, and a coordinator whose schema is older than the listing simply returns no row. Both reads mapped a missing row to `Ok(None)`, which the caller read as "there is nothing to index here": the index was dropped from the discovered set and the round still returned `Ok`.

A node that loses this race answers `404 missing index` for an index its peers serve, and an index it had already built and was serving is removed from it, until any DDL happens in the cluster. Both skips logged at `debug!` only, so nothing reported it.

VECTOR-882 collects the occurrences seen in CI, the log evidence for each, and why this change fixes the individual reports linked to it.

Fixes: VECTOR-882
Fixes: VECTOR-680
Fixes: VECTOR-868
Fixes: VECTOR-881

### DIFF
--- a/crates/vector-store/src/db_index_backend.rs
+++ b/crates/vector-store/src/db_index_backend.rs
@@ -11,6 +11,7 @@ use crate::KeyspaceIdentifier;
 use crate::KeyspaceName;
 use crate::TableIdentifier;
 use crate::TableName;
+use anyhow::bail;
 use futures::TryStreamExt;
 use regex::Regex;
 use scylla::client::session::Session;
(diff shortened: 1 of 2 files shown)


Akvear (approved):

crates/vector-store/src/monitor_indexes.rs
knowack1, line 184: It looks like get_index_version, reads options from `system_schema.indexes`, the we already get in `get_indexes`. My point is that maybe this query can be eliminated at all.

crates/vector-store/src/db_index_backend.rs
```diff
@@ -145,21 +146,28 @@ async fn get_dimensions_from_column_type(
...
         .await?;
-    let dimensions = column_type
-        .and_then(|(typ,)| {
-            re_get_index_target_type
-                .captures(&typ)
-                .and_then(|captures| captures["dimensions"].parse::<usize>().ok())
-        })
+
+    // A missing row means the node that served the read has not applied the schema change yet.
+    let Some((column_type,)) = column_type else {
```
knowack1, line 165: I would consider to open a jira ticket to add dementions to the index options on ScyllaDB side. This will eliminate this extra query and unify the source of index metadata.

crates/vector-store/src/db.rs
```diff
@@ -956,7 +956,10 @@ impl Statements {
             debug!("is_valid_index: no active session for {}", metadata.key());
             return false;
         };
-        let Ok(version_begin) = session.await_schema_agreement().await else {
+
+        // Awaiting the agreement would block the discovery round for up to the driver's 60s
+        // timeout, so report the index as not valid yet and let the next tick retry it.
+        let Ok(Some(version_begin)) = session.check_schema_agreement().await else {
```
knowack1: I do not fully understand this change. Seems to be very intrusive. If this is optimization I would skip it in this PR.

Maybe the agreement check (guard) should start prior to fetching any index info (before get_index_version).

QuerthDP: Dropped the commit. Proposed follow up: VECTOR-884

crates/vector-store/src/db_index_backend.rs
QuerthDP: SCYLLADB-3936 and VECTOR-886

crates/vector-store/src/monitor_indexes.rs
QuerthDP: VECTOR-885

QuerthDP:
Changelog:
- dropped 3rd commit. This will be defered to separate PR

knowack1 (approved):

PR #569, opened 2026-08-24
Don't route ANN requests across indexes with different options

nyh:
If a table has two vector indexes on the same column but with different DistanceFunction (e.g., one COSINE and one EUCLIDEAN), an ANN query addressed to one of them could silently be served by the other, using the wrong distance metric and returning a wrong nearest-neighbor result with no error.
    
Commit 5326ed26bc6b9d66dbe0b1892034cd8c5eb00f1c ("vector-store: route ANN requests to best matching serving index") deliberately grouped all indexes over the same keyspace/table/target column into one "routing group" and, for a query addressed to any index in the group, routes it to whichever member of the group is the best-scoring and newest, not necessarily the one actually named. This was intentional: it lets an index be rebuilt under a new name and take over serving without the caller having to know the new name, and it lets a more specific local index be preferred over a global one. Our own test suite (tests/integration/routing.rs) confirms this is deliberate: it asserts that a query addressed to an older, fully-serving index gets routed to a newer, fully-serving one on the same column, purely because the newer one is presumed to be the intended upgrade.    

That grouping was too broad: it only considered keyspace, table, and target column, so any two indexes on the same column ended up interchangeable regardless of what they actually compute. Two indexes that disagree on any part of their definition are not safe substitutes for one another - obviously not distance function or dimensionality, but not even the ANN tuning parameters (connectivity, expansion_add, expansion_search, quantization), since those govern an approximate search structure and can themselves make two indexes return a different top-K for the same query.
    
Fix this by requiring full equality of IndexOptionsVs (not just a hand-picked subset of its fields) for two indexes to be considered part of the same routing group. Indexes may still differ in name, partitioning, filtering columns, or version and be routed between each other - that's the point of the feature - but only when their definition is otherwise identical.
    
The second patch adds a regression test that reproduces this bug - and checks its fix.

Fixes: VECTOR-880

### DIFF
--- a/crates/vector-store/src/indexes.rs
+++ b/crates/vector-store/src/indexes.rs
@@ -58,21 +58,24 @@ impl Ord for NeedsFiltering {
     }
 }
 
-/// Key for grouping indexes that can be routed between each other
-/// (i.e., indexes over the same keyspace, table, and target column).
+/// Key for grouping indexes that can be routed between each other: the same
+/// keyspace, table, and target column, with identical index options. Name,
+/// partitioning, filtering columns, and version may still differ.
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 pub(crate) struct RoutingGroupKey {
     keyspace: KeyspaceName,
     table: TableName,
     columns: NonemptyArc<ColumnName>,
+    options: crate::IndexOptionsVs,
 }
 
-impl From<&IndexMetadata> for RoutingGroupKey {
-    fn from(metadata: &IndexMetadata) -> Self {
+impl RoutingGroupKey {
+    fn new(metadata: &IndexMetadata, options: &crate::IndexOptionsVs) -> Self {
         Self {
             keyspace: metadata.keyspace_name.clone(),
             table: metadata.table_name.clone(),
             columns: metadata.target_columns.clone(),
+            options: options.clone(),
         }
     }
 }
@@ -152,13 +155,13 @@ impl VsIndexEntry {
         db_index: mpsc::Sender<DbIndex>,
         metadata: IndexMetadata,
     ) -> anyhow::Result<Self> {
-        let routing_group = RoutingGroupKey::from(&metadata);
         let options = metadata
             .vs()
             .ok_or_else(|| {
                 anyhow::anyhow!("add_index_vs must be called with a vector-search index")
             })?
             .clone();
+        let routing_group = RoutingGroupKey::new(&metadata, &options);
         let filtering_columns = metadata
             .primary_key_columns
             .iter()
@@ -364,7 +367,8 @@ impl Indexes {
     ///
     /// 1. Returns `NotFound` if the requested index key does not exist.
     /// 2. Identifies all candidate indexes within the same routing group
-    ///    (i.e., sharing the same keyspace, table, and target column).
+    ///    (i.e., sharing the same keyspace, table, target column, and
+    ///    index options).
     /// 3. Filters out candidates whose `score_index` returns `None` (invalid).
     /// 4. Narrows down the remaining candidates to those that are actively serving.
     /// 5. Picks the candidate with the highest score, breaking ties by
--- a/crates/vector-store/tests/integration/routing.rs
+++ b/crates/vector-store/tests/integration/routing.rs
@@ -5,6 +5,7 @@
 
 use crate::common::add_table;
 use crate::common::blocking_scan_fn;
+use crate::common::make_index_with_kind;
 use crate::common::make_vs_index;
 use crate::common::ordered_timeuuid;
 use crate::common::setup;
@@ -21,8 +22,11 @@ use scylla::value::CqlValue;
 use std::num::NonZeroUsize;
 use std::time::Duration;
 use vector_store::DbIndexPartitioning;
+use vector_store::IndexKind;
 use vector_store::IndexMetadata;
+use vector_store::IndexOptionsVs;
 use vector_store::NonemptyArc;
+use vector_store::SpaceType;
 
 const ANN_LIMIT: usize = 5;
 
@@ -126,6 +130,51 @@ async fn assert_ann_served_by(
         1,
         "expected ANN request to be served by {}",
         expected.index_name,
+    );
+    response
+}
+
+/// Like `assert_ann_served_by`, but also asserts the request was not served
+/// by `not_expected`.
+async fn assert_ann_served_by_and_not(
+    client: &HttpClient,
+    expected: &IndexMetadata,
+    not_expected: &IndexMetadata,
+    request: impl std::future::Future<Output = reqwest::Response>,
+) -> reqwest::Response {
+    client
+        .internals_clear_counters()
+        .await
+        .expect("internals counters must be cleared");
+    let expected_counter = format!(
+        "ann-served-request--{}--{}",
+        expected.keyspace_name, expected.index_name
+    );
+    let not_expected_counter = format!(
+        "ann-served-request--{}--{}",
+        not_expected.keyspace_name, not_expected.index_name
+    );
+    for counter in [&expected_counter, &not_expected_counter] {
+        client
+            .internals_start_counter(counter.clone())
+            .await
+            .expect("internals served counter must be registered");
+    }
+
+    let response = request.await;
+
+    let counters = client.internals_counters().await.unwrap();
+    assert_eq!(
+        counters.get(&expected_counter).copied().unwrap_or(0),
+        1,
+        "expected ANN request to be served by {}",
+        expected.index_name,
+    );
+    assert_eq!(
+        counters.get(&not_expected_counter).copied().unwrap_or(0),
+        0,
+        "expected ANN request not to be served by {}",
+        not_expected.index_name,
     );
     response
 }
@@ -668,5 +717,86 @@ async fn ann_routes_to_global_index_without_pk_restrictions() {
 
     let response =
         assert_ann_served_by(&client, &global_index, post_ann(&client, &local_index)).await;
+    assert_eq!(response.status(), StatusCode::OK);
+}
+
+fn vs_options_with_space_type(space_type: SpaceType) -> IndexOptionsVs {
+    IndexOptionsVs {
+        dimensions: NonZeroUsize::new(3).unwrap().into(),
+        connectivity: Default::default(),
+        expansion_add: Default::default(),
+        expansion_search: Default::default(),
+        space_type,
+        quantization: Default::default(),
+    }
+}
+
+/// Regression test for indexes on the same column being routed to each
+/// other despite disagreeing on their distance function (or any other part
+/// of their definition). Two indexes on the same column, one COSINE and one
+/// EUCLIDEAN, are both fully serving; the EUCLIDEAN one is the newer of the
+/// two, so before the fix it would win the "prefer the newest index in the
+/// routing group" tie-break even for a query addressed to the COSINE index,
+/// silently answering it under the wrong distance function.
+#[rstest]
+#[timeout(Duration::from_secs(10))]
+#[tokio::test]
+#[cfg_attr(not(feature = "slow-test-hooks"), ignore = "requires slow-test-hooks")]
+async fn ann_does_not_route_between_indexes_with_different_distance_function() {
+    crate::enable_tracing();
+    let (client, db, _keep) = setup().await;
+
+    add_table(
+        &db,
+        ["pk".into()],
+        1,
+        [("pk".into(), NativeType::Int)],
+        ["embedding".into()],
+    );
+
+    let cosine = make_index_with_kind(
+        "cosine",
+        &["pk"],
+        1,
+        "embedding",
+        DbIndexPartitioning::Global,
+        &[],
+        ordered_timeuuid(1),
+        IndexKind::Vs(vs_options_with_space_type(SpaceType::Cosine)),
+    );
+    db.add_index(
+        cosine.clone(),
+        Some(single_row_scan([CqlValue::Int(1)])),
+        None,
+    )
+    .unwrap();
+    wait_for_serving(&client, &cosine).await;
+
+    let euclidean = make_index_with_kind(
+        "euclidean",
+        &["pk"],
+        1,
+        "embedding",
+        DbIndexPartitioning::Global,
+        &[],
+        ordered_timeuuid(2),
+        IndexKind::Vs(vs_options_with_space_type(SpaceType::Euclidean)),
+    );
+    db.add_index(
+        euclidean.clone(),
+        Some(single_row_scan([CqlValue::Int(1)])),
+        None,
+    )
+    .unwrap();
+    wait_for_serving(&client, &euclidean).await;
+
+    let response =
+        assert_ann_served_by_and_not(&client, &cosine, &euclidean, post_ann(&client, &cosine))
+            .await;
+    assert_eq!(response.status(), StatusCode::OK);
+
+    let response =
+        assert_ann_served_by_and_not(&client, &euclidean, &cosine, post_ann(&client, &euclidean))
+            .await;
     assert_eq!(response.status(), StatusCode::OK);
 }


crates/vector-store/src/indexes.rs
```diff
@@ -364,7 +382,8 @@ impl Indexes {
     ///
     /// 1. Returns `NotFound` if the requested index key does not exist.
     /// 2. Identifies all candidate indexes within the same routing group
-    ///    (i.e., sharing the same keyspace, table, and target column).
+    ///    (i.e., sharing the same keyspace, table, target column, and
+    ///    `IndexOptionsVs`).
```
QuerthDP: nit: and index options. Do not mention `IndexOptionsVs` as this will be used by FTS indexes in future as well.

crates/vector-store/src/indexes.rs
```diff
@@ -59,20 +59,38 @@ impl Ord for NeedsFiltering {
 }
 
 /// Key for grouping indexes that can be routed between each other
-/// (i.e., indexes over the same keyspace, table, and target column).
+/// (i.e., indexes over the same keyspace, table, and target column, built
+/// with identical `IndexOptionsVs`).
```
QuerthDP: ditto. Don't mention `IndexOptionsVs`, let's just use "index options".

crates/vector-store/src/indexes.rs
```diff
@@ -59,20 +59,38 @@ impl Ord for NeedsFiltering {
...
+/// over a global one. But two indexes whose `IndexOptionsVs` differ in any
+/// way are never interchangeable: besides the obvious case of a different
+/// distance function or dimensionality, even two indexes that only differ
+/// in ANN tuning (`connectivity`, `expansion_add`, `expansion_search`,
+/// `quantization`) are approximate structures that can legitimately return
+/// a different top-K for the same query, so one must never be silently
+/// substituted for the other. Requiring full equality of `IndexOptionsVs`,
+/// rather than singling out the fields that happened to matter for a known
+/// bug, avoids reintroducing this class of bug for whichever option is
+/// picked next.
```
QuerthDP: Good explanaition, but it would be enough as a commit message. Let's make it more concise here.

nyh:
Beyond the unit test included in this PR (the second patch), I also have an Alternator test `test_searchvectors_multiple_indexes_same_attribute` which exercises this situation (two vector indexes on the same column, with different distance functions configured), and reproduces this bug - and that the fix proposed here works.

crates/vector-store/tests/integration/routing.rs
```diff
@@ -670,3 +674,80 @@ async fn ann_routes_to_global_index_without_pk_restrictions() {
...
+        None,
+    )
+    .unwrap();
+    wait_for_serving(&client, &euclidean).await;
+
+    let response = assert_ann_served_by(&client, &cosine, post_ann(&client, &cosine)).await;
+    assert_eq!(response.status(), StatusCode::OK);
+
+    let response = assert_ann_served_by(&client, &euclidean, post_ann(&client, &euclidean)).await;
+    assert_eq!(response.status(), StatusCode::OK);
```
QuerthDP: Maybe we should also assert that it wasn't served by the other?

QuerthDP (approved):
Looks really good. Thanks for finding it out!

QuerthDP:
@knowack1 please take a look as well. I think this means we should prioritize the task to use the routing for FTS indexes as well, since we would like to support the analyzer options setup (VECTOR-788).

crates/vector-store/src/indexes.rs
nyh: Done.

crates/vector-store/src/indexes.rs
nyh: done,

crates/vector-store/src/indexes.rs
nyh: I'll shorten significantly. Project differs on where they want their good explanations to reside. I'm not a big fan on making commit messages the source of truth (although I still write them seriously), but this is your project you know best how you guys prefer it.

QuerthDP (changes requested):
On the second thought, I don't know if that is a correct approach.
ScyllaDB picks the index it queries at random. Once we have 2 indexes with different parameters we don't have any option to choose which one to query. With the "bugged" approach, we don't either, yet we have a semantic that the newest index on the column will be chosen. Now it's not the case, the newest index of each option state will be chosen, but we choose the state at random.

I assume that alternator has a way to specify which index to query right? That's why it fixes your issue, but it's not a proper way to do it without any other Scylla changes I think.

swasik:
> obviously not distance function or dimensionality, but not even the ANN tuning parameters (connectivity, expansion_add, expansion_search, quantization), since those govern an approximate search structure and can themselves make two indexes return a different top-K for the same query.

@nyh this is not fully true - we may want to use this mechanism to change the values of parameters and this is what customers already do. Maybe not distance function - it should match the one from query but hnsw parameters - it is completely fine to change then

crates/vector-store/tests/integration/routing.rs
nyh: Ok. Added a new helper assert_ann_served_by_and_not() to implement this (we can't check it wit the existing helper).

nyh:
> On the second thought, I don't know if that is a correct approach. ScyllaDB picks the index it queries at random. Once we have 2 indexes with different parameters we don't have any option to choose which one to query.

This is only the case for CQL, **not** for Alternator.

If Scylla CQL knows this is the case, it should actively refuse to create a second index with different parameters. If it doesn't, it's a CQL bug that should be fixed in Scylla. (or, it's intentional, not a bug? more on this below).

Meanwhile, the DynamoDB API does ask you to pick the index (`IndexName`) during the query, and does allow you to create two different indexes on the same column. And we need the vector store to support this ability - even if CQL doesn't want to use it.

This bug is breaking a real DynamoDB API feature, that I have a real test for (the test passes on DynamoDB, and fails on Alternator when the wrong index is used).

> With the "bugged" approach, we don't either, yet we have a semantic that the newest index on the column will be chosen. Now it's not the case, the newest index of each option state will be chosen, but we choose the state at random.

If I understand correctly, the original reason we have this logic is the desire to allow in CQL to create a second **identical** index which will get used when its ready. CQL never intended to allow you to create a second non-identical index - because as you said you'll never be able to pick between them during search. If I remember correctly, in the past CQL had a check that didn't allow you to create a second index on the same column. I am guessing you dropped this check to allow creating a second **identical** index on the column - but forgot to add the "identical" check. This should be added - it's a CQL bug. We shouldn't fix a CQL bug by introducing a second bug in the vector store that "cancels" it.

> I assume that alternator has a way to specify which index to query right?

Yes.

> That's why it fixes your issue, but it's not a proper way to do it without any other Scylla changes I think.

I want to understand if what you're trying to "preserve" is a CQL bug, or intentional behavior:

Did you want to have the ability to change the definition of a vector index (e.g., the distance function), by creating a new index with a **different name** with a new distance function, and then when the new index finishes building, in one moment you'll suddenly switch index (with suddenly different results! Really strange behavior!)

Was this an intentional feature you wanted to support?

Do we have a test that claims that this is an intentional feature we want to support? (I didn't find it in the vector store's unit tests, but maybe it's somewhere else)

If this is an intentional feature (although it sounds a rather bad feature to me...), then I agree, we won't be able to fix this bug (that breaks Alternator) without thinking how to preserve that intentional feature in CQL. But if it was never intentional - the original idea was to allow you to create a second identical index, not a second different table - then we don't need to preserve it.

nyh:
> > obviously not distance function or dimensionality, but not even the ANN tuning parameters (connectivity, expansion_add, expansion_search, quantization), since those govern an approximate search structure and can themselves make two indexes return a different top-K for the same query.
> 
> @nyh this is not fully true - we may want to use this mechanism to change the values of parameters and this is what customers already do. Maybe not distance function - it should match the one from query but hnsw parameters - it is completely fine to change then

Oh, this is sad. You basically invented a new CQL feature - allowing you to **change the definition of an existing index** by creating a new one, but then you can't actively pick which index is used - it's the newest fully-built index that is used. And you only implemented this new trick in vector store, not in other indexes (like secondary index), and the implementation is by introducing a bizarre bug into the vector store that broke its ability work with other APIs like DynamoDB.

You said "Maybe not distance function". Where is this "not distance function" enforced?

**I can make my fix specifically check just the distance function - and not any of the other "index options". This will be good enough for the DynamoDB API because the distance function is the only option it cares about (the first version of my patch actually compared just the distance function, not all the index options). Should I do that instead?**

You also said "hnsw parameters - it is completely fine to change then". I don't understand why it's fine. If you create a second index with different options, then you'll continue to make requests with the old options until sometime later - maybe in a minute, maybe in an hour - suddenly all the queries switch to the new options. Who would want this behavior?

QuerthDP:
> You also said "hnsw parameters - it is completely fine to change then". I don't understand why it's fine. If you create a second index with different options, then you'll continue to make requests with the old options until sometime later - maybe in a minute, maybe in an hour - suddenly all the queries switch to the new options. Who would want this behavior?

Consider that: You want to have your index fully working and available all the time. You're fine that the recall is a bit lower, but you keep to serve the queries. You come to an idea that you can tradeoff some of the latency to get better results, so that you query the CREATE INDEX on the same column with better parameters to substitute your working index. Once the index is built, it automatically keeps serving your requests of a better quality. There is no break in between, everything happens on flight once the index is built. No need to check for the status or anything. It's really easy for the user. 

On the other hand, changing the similarity function which actually changes the whole point of that index is quite bizarre I can agree. And this could be a grouping factor in a routing group, just like you need for the Alternator implementation. 
It's a pity we cannot do anything with it on the CQL side right now, so that we would just introduce a behavior (or better to call it a bug?) that makes the queries to pick a random index with random similarity function which is really bad. On the other hand, we cannot restrict that, because once we do, the Alternator requirements will not be satisfied? Or is it a totally different path?

QuerthDP:
Furthermore, I'm not sure if we really want to restrict the users from creating Vector Indexes of different similarity functions on one column. This might be useful for testing which of those perform the best based on their dataset. There should be some way in CQL to instruct the query, which index/similarity function to pick exactly, but I think it must be introduced first, before we make any changes here.

nyh:
> Furthermore, I'm not sure if we really want to restrict the users to create Vector Indexes of different similarity functions on one column. This might be useful for testing which of those perform the best based on their dataset. There should be some way in CQL to instruct the query, which index/similarity function to pick exactly, but I think it must be introduced first, before we make any changes here.

You'll not be able to introduce the feature you propose to CQL without fixing this here - which is what the Alternator implementation of this feature shows.

In my opinion, something is completely broken in the whole approach here. Alternator calls ANN with a **specific index name**, and instead of using that index, vector store goes on to pick a different index. I consider this a bug. Maybe this bug helped you implement a strange CQL feature without touching CQL, but I think the right solution would have been to touch CQL: If CQL knows it has two matching indexes (those are indexes in the CQL level), it should pick which one to use - it shouldn't pick one at random (!) and then expect the vector store to override this decision :-(

The fact that CQL picks a random index and expects the vector store to override it, means that when DynamoDB that doesn't have this problem, and lets you pick the index you want to use explicitly, suddenly can't pick the right index, because no matter what index it will pick - the vector store will override it.

QuerthDP:
> In my opinion, something is completely broken in the whole approach here. Alternator calls ANN with a **specific index name**, and instead of using that index, vector store goes on to pick a different index. I consider this a bug. Maybe this bug helped you implement a strange CQL feature without touching CQL, but I think the right solution would have been to touch CQL: If CQL knows it has two matching indexes (those are indexes in the CQL level), it should pick which one to use - it shouldn't pick one at random (!) and then expect the vector store to override this decision :-(

That's right, we used it deliberately. Since CQL always picked one index, even when it had two, we implemented a whole routing on the Vector Store side. The `{index}` in the endpoint only serves as a starting point from which the table/column/type, etc. are known. I think this might be not documented however.

> The fact that CQL picks a random index and expects the vector store to override it, means that when DynamoDB that doesn't have this problem, and lets you pick the index you want to use explicitly, suddenly can't pick the right index, because no matter what index it will pick - the vector store will override it.

I could think of an approach to handle this problem, which makes it win-win situation on both Vector Store's routing/Alternator sides.

We could have added an additional endpoint, for example `/indexes/{keyspace}/{index}/ann/explicit` or just an additional `explicit` (default `false`) or `routing` (default `true`) flag to make the Vector Store respect whether we want to choose "best" index based on our internal routing, or explicitly used the queried index - the Alternator style.
On CQL side we could have added some `USING {'routing' : false}` or something, but it wouldn't require any changes right now, only some future work and better design.

QuerthDP:
> > In my opinion, something is completely broken in the whole approach here. Alternator calls ANN with a **specific index name**, and instead of using that index, vector store goes on to pick a different index. I consider this a bug. Maybe this bug helped you implement a strange CQL feature without touching CQL, but I think the right solution would have been to touch CQL: If CQL knows it has two matching indexes (those are indexes in the CQL level), it should pick which one to use - it shouldn't pick one at random (!) and then expect the vector store to override this decision :-(
> 
> That's right, we used it deliberately. Since CQL always picked one index, even when it had two, we implemented a whole routing on the Vector Store side. The `{index}` in the endpoint only serves as a starting point from which the table/column/type, etc. are known. I think this might be not documented however.
> 
> > The fact that CQL picks a random index and expects the vector store to override it, means that when DynamoDB that doesn't have this problem, and lets you pick the index you want to use explicitly, suddenly can't pick the right index, because no matter what index it will pick - the vector store will override it.
> 
> I could think of an approach to handle this problem, which makes it win-win situation on both Vector Store's routing/Alternator sides.
> 
> We could have added an additional endpoint, for example `/indexes/{keyspace}/{index}/ann/explicit` or just an additional `explicit` (default `false`) or `routing` (default `true`) flag to make the Vector Store respect whether we want to choose "best" index based on our internal routing, or explicitly used the queried index - the Alternator style. On CQL side we could have added some `USING {'routing' : false}` or something, but it wouldn't require any changes right now, only some future work and better design.

To make it clear.

I propose to add additional parameter to the `/indexes/{keyspace}/{index}/ann`/`bm25` endpoints.
That would be `routing` (default `true`) which would instruct the Vector Store to route the query based on the rules already implemented.
That would preserve the original behavior we have now. Require no CQL changes, and allow Alternator to use it with `routing : false` to query the exact index based on similarity function they want.

nyh:
> > > In my opinion, something is completely broken in the whole approach here. Alternator calls ANN with a **specific index name**, and instead of using that index, vector store goes on to pick a different index. I consider this a bug. Maybe this bug helped you implement a strange CQL feature without touching CQL, but I think the right solution would have been to touch CQL: If CQL knows it has two matching indexes (those are indexes in the CQL level), it should pick which one to use - it shouldn't pick one at random (!) and then expect the vector store to override this decision :-(
> > 
> > 
> > That's right, we used it deliberately. Since CQL always picked one index, even when it had two, we implemented a whole routing on the Vector Store side. The `{index}` in the endpoint only serves as a starting point from which the table/column/type, etc. are known. I think this might be not documented however.
> > > The fact that CQL picks a random index and expects the vector store to override it, means that when DynamoDB that doesn't have this problem, and lets you pick the index you want to use explicitly, suddenly can't pick the right index, because no matter what index it will pick - the vector store will override it.
> > 
> > 
> > I could think of an approach to handle this problem, which makes it win-win situation on both Vector Store's routing/Alternator sides.
> > We could have added an additional endpoint, for example `/indexes/{keyspace}/{index}/ann/explicit` or just an additional `explicit` (default `false`) or `routing` (default `true`) flag to make the Vector Store respect whether we want to choose "best" index based on our internal routing, or explicitly used the queried index - the Alternator style. On CQL side we could have added some `USING {'routing' : false}` or something, but it wouldn't require any changes right now, only some future work and better design.
> 
> To make it clear.
> 
> I propose to add additional parameter to the `/indexes/{keyspace}/{index}/ann`/`bm25` endpoints. That would be `routing` (default `true`) which would instruct the Vector Store to route the query based on the rules already implemented. That would preserve the original behavior we have now. Require no CQL changes, and allow Alternator to use it with `routing : false` to query the exact index based on similarity function they want.

I'll do that and close this PR. Alternator will just set this flag to its non-default value.

nyh:
Replaced by https://github.com/scylladb/vector-store/pull/570

swasik:
> Oh, this is sad. You basically invented a new CQL feature - allowing you to **change the definition of an existing index** by creating a new one, but then you can't actively pick which index is used - it's the newest fully-built index that is used. And you only implemented this new trick in vector store, not in other indexes (like secondary index), and the implementation is by introducing a bizarre bug into the vector store that broke its ability work with other APIs like DynamoDB.

@nyh I do not fully agree. Similar mechanism already exist in CQL. Currently, if there are two secondary indexes available that can match the query - local and global - the Scylla will automatically select the "better" one, in this case the local that has better performance. So the assumption that this is Scylla who selects which index to use was already there in CQL. We just reused it. First to prefer local vector indexes vs. global automatically and  then to prefer the fresher ones. Of course the latter did not exist in case of secondary indexes as it does not make sense there. But the assumption "this is Scylla not the user who selects index" was there and we reused it in vector search.

PR #570, opened 2026-08-24
vector-store: add "routing" option to disable index-routing per-request

nyh:
The vector-search ANN endpoint let a query for one index be silently served by a different, "better" index on the same table and column - the newest fully-serving one, or one requiring less server-side filtering. This routing exists for CQL: CQL can create a second index over the same column (e.g. to swap in new HNSW parameters, or even a different distance function) without any way for a query to say which of the two it wants, so the vector store instead always picks the best match (e.g., the last-built index) on the fly.

Alternator has no such ambiguity: DynamoDB's API lets the caller name the exact index to query (IndexName), and different indexes on the same column with different distance functions are an intentional, documented feature, not a fallback. Silently overriding that explicit choice - as could happen here - breaks the DynamoDB semantics: a query addressed to a COSINE index could be served by a EUCLIDEAN one instead, returning a "top-K" that is simply wrong for the query that was asked, without any error.

Rather than changing what makes indexes eligible for routing (which was attempted and rejected during review), this patch adds an optional "routing" flag (default true) to the ANN request bodies. Alternator will set it to false so its queries are always served by exactly the index they name, while CQL keeps relying on the default routing behavior unchanged.

Add a test to the vector-search routing test suite, right after the one confirming an index is by default routed to a newer replacement, that same setup now queried with routing disabled - showing the older index keeps serving its own queries instead of being routed away.

Fixes: VECTOR-880

### DIFF
--- a/api/openapi.json
+++ b/api/openapi.json
@@ -789,6 +789,10 @@
           "limit": {
             "$ref": "#/components/schemas/Limit"
           },
+          "routing": {
+            "type": "boolean",
+            "description": "By default, the query may be routed to a different, better-matching index on the same column. Set to `false` to force the query to be served by exactly the index named in the URL."
+          },
           "vector": {
             "$ref": "#/components/schemas/Vector"
           }
--- a/crates/httpapi/src/lib.rs
+++ b/crates/httpapi/src/lib.rs
@@ -371,6 +371,13 @@ pub struct PostIndexAnnRequest {
     pub filter: Option<PostIndexAnnFilter>,
     #[serde(default)]
     pub limit: Limit,
+    /// By default, the query may be routed to a different, better-matching index on the same column. Set to `false` to force the query to be served by exactly the index named in the URL.
+    #[serde(default = "default_routing")]
+    pub routing: bool,
+}
+
+fn default_routing() -> bool {
+    true
 }
 
 #[derive(serde::Deserialize, serde::Serialize, utoipa::ToSchema)]
--- a/crates/vector-store/src/indexes.rs
+++ b/crates/vector-store/src/indexes.rs
@@ -359,8 +359,8 @@ impl Indexes {
 
     /// Determines the index to route a query to, given a requested `IndexKey`.
     ///
-    /// To ensure queries are routed to the most up-to-date and best-matching index,
-    /// this function applies the following routing logic:
+    /// When `routing` is `true`, ensures queries are routed to the most
+    /// up-to-date and best-matching index by applying the following logic:
     ///
     /// 1. Returns `NotFound` if the requested index key does not exist.
     /// 2. Identifies all candidate indexes within the same routing group
@@ -370,19 +370,27 @@ impl Indexes {
     /// 5. Picks the candidate with the highest score, breaking ties by
     ///    the newest `IndexVersion`.
     /// 6. Returns `NotServing` if no candidate meets the criteria.
+    ///
+    /// When `routing` is `false`, the only candidate considered is the
+    /// requested index itself - the query is served by exactly the named
+    /// index, or not at all.
     pub(crate) fn best_index(
         &self,
         key: &IndexKey,
         equality_columns: &[ColumnName],
         range_columns: &[ColumnName],
+        routing: bool,
     ) -> BestIndexState {
         let Some(requested_entry) = self.vs_entries.get(key) else {
             return BestIndexState::NotFound;
         };
-        let candidates = self
-            .vs_routing
-            .get(&requested_entry.data.routing_group)
-            .expect("routing_map must contain group for every index in indexes");
+        let candidates: &[IndexKey] = if routing {
+            self.vs_routing
+                .get(&requested_entry.data.routing_group)
+                .expect("routing_map must contain group for every index in indexes")
+        } else {
+            std::slice::from_ref(key)
+        };
 
         let routed_key = candidates
             .iter()
--- a/crates/httpclient/src/lib.rs
+++ b/crates/httpclient/src/lib.rs
@@ -92,6 +92,7 @@ impl HttpClient {
             vector,
             filter,
             limit,
+            routing: true,
         };
         self.post_ann_data(keyspace_name, index_name, &request)
             .await
--- a/crates/testclient/src/lib.rs
+++ b/crates/testclient/src/lib.rs
@@ -59,6 +59,7 @@ impl TestClient {
                 vector,
                 filter,
                 limit,
+                routing: true,
             })
             .await
     }
--- a/crates/vector-store/src/httproutes.rs
+++ b/crates/vector-store/src/httproutes.rs
@@ -734,12 +734,12 @@ async fn post_index_ann(
         let index_key = IndexKey::new(&keyspace, &index_name);
         let (equality_cols, range_cols) = restriction_columns(&request.filter);
         let allow_filtering = request.filter.as_ref().is_some_and(|f| f.allow_filtering);
-        let best_index_state =
-            state
-                .indexes
-                .read()
-                .unwrap()
-                .best_index(&index_key, &equality_cols, &range_cols);
+        let best_index_state = state.indexes.read().unwrap().best_index(
+            &index_key,
+            &equality_cols,
+            &range_cols,
+            request.routing,
+        );
         let (routed_key, index, primary_key_columns, filtering_columns, table_columns) =
             match best_index_state {
             indexes::BestIndexState::Serving {
--- a/crates/vector-store/tests/integration/https.rs
+++ b/crates/vector-store/tests/integration/https.rs
@@ -99,6 +99,7 @@ async fn test_https_server_responds() {
             vector: vec![1.0].into(),
             filter: None,
             limit: NonZeroUsize::new(1).unwrap().into(),
+            routing: true,
         })
         .send()
         .await
@@ -111,6 +112,7 @@ async fn test_https_server_responds() {
             vector: vec![1.0].into(),
             filter: None,
             limit: NonZeroUsize::new(1).unwrap().into(),
+            routing: true,
         })
         .send()
         .await
--- a/crates/vector-store/tests/integration/routing.rs
+++ b/crates/vector-store/tests/integration/routing.rs
@@ -12,6 +12,7 @@ use crate::common::single_row_scan;
 use crate::wait_for;
 use httpapi::IndexStatus;
 use httpapi::PostIndexAnnFilter;
+use httpapi::PostIndexAnnRequest;
 use httpapi::PostIndexAnnRestriction;
 use httpclient::HttpClient;
 use reqwest::StatusCode;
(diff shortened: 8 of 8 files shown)


crates/httpapi/src/lib.rs
```diff
@@ -468,6 +479,13 @@ pub struct PostIndexBm25Request {
     pub query: String,
     #[serde(default)]
     pub limit: Limit,
+    /// Whether the query may be routed to a different, better-matching index
+    /// on the same column than the one named in the URL. Defaults to `true`,
+    /// preserving the existing routing behavior. Set to `false` to force the
+    /// query to be served by exactly the named index, e.g. when the caller
+    /// (such as Alternator) has already picked the index it wants by name.
+    #[serde(default = "default_routing")]
+    pub routing: bool,
```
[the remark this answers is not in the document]
nyh: I don't know what to do about the BM25 version, I think it doesn't have routing at all? I wanted to leave this option to remind whoever implement routing that it should be optional, but maybe I shouldn't touch bm25 at all? @QuerthDP what do you think?

crates/httpapi/src/lib.rs
```diff
@@ -371,6 +371,17 @@ pub struct PostIndexAnnRequest {
     pub filter: Option<PostIndexAnnFilter>,
     #[serde(default)]
     pub limit: Limit,
+    /// Whether the query may be routed to a different, better-matching index
+    /// on the same column than the one named in the URL. Defaults to `true`,
+    /// preserving the existing routing behavior. Set to `false` to force the
+    /// query to be served by exactly the named index, e.g. when the caller
+    /// (such as Alternator) has already picked the index it wants by name.
+    #[serde(default = "default_routing")]
+    pub routing: bool,
```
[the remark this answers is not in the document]
nyh: I'll fix. This is indeed important because the Scylla-side code I'm planning to write was planning to omit "routing" altogether if true - and not set it to true explicitly.

nyh:
This PR replaces the old PR https://github.com/scylladb/vector-store/pull/569 whose approach was rejected by @QuerthDP's review.

I verified that beyond passing the new unit tests, this fix also helps Alternator: when this patch is used and Alternator is modified to set routing=false, the Alternator test `test_searchvectors_multiple_indexes_same_attribute` which checks multiple different indexes on the same vector attribute - passes.

crates/httpapi/src/lib.rs
```diff
@@ -371,6 +371,17 @@ pub struct PostIndexAnnRequest {
     pub filter: Option<PostIndexAnnFilter>,
     #[serde(default)]
     pub limit: Limit,
+    /// Whether the query may be routed to a different, better-matching index
+    /// on the same column than the one named in the URL. Defaults to `true`,
+    /// preserving the existing routing behavior. Set to `false` to force the
+    /// query to be served by exactly the named index, e.g. when the caller
+    /// (such as Alternator) has already picked the index it wants by name.
+    #[serde(default = "default_routing")]
```
QuerthDP, line 375: This is really weird that we cannot just use `#[serde(default = true)]` but it seems like it's not supported within that macro.

crates/httpapi/src/lib.rs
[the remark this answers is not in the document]
QuerthDP: I think that it should be deferred to the PR that will enable routing for BM25 calls. Let's not bother with it now and only change the ANN path.

crates/vector-store/src/indexes.rs
```diff
@@ -370,19 +370,30 @@ impl Indexes {
...
+        let group_candidates;
+        let candidates: &[IndexKey] = if routing {
+            group_candidates = self
+                .vs_routing
+                .get(&requested_entry.data.routing_group)
+                .expect("routing_map must contain group for every index in indexes");
+            group_candidates
+        } else {
+            std::slice::from_ref(key)
+        };
```
QuerthDP:
```suggestion
let candidates: &[IndexKey] = if routing {
    self.vs_routing
        .get(&requested_entry.data.routing_group)
        .expect("routing_map must contain group for every index in indexes")
} else {
    std::slice::from_ref(key)
};
```

Let's do it more Rustic way. It feels really weird seeing the uninitialized variable here.

crates/httpapi/src/lib.rs
```diff
@@ -371,6 +371,17 @@ pub struct PostIndexAnnRequest {
     pub filter: Option<PostIndexAnnFilter>,
     #[serde(default)]
     pub limit: Limit,
+    /// Whether the query may be routed to a different, better-matching index
+    /// on the same column than the one named in the URL. Defaults to `true`,
+    /// preserving the existing routing behavior. Set to `false` to force the
+    /// query to be served by exactly the named index, e.g. when the caller
+    /// (such as Alternator) has already picked the index it wants by name.
```
QuerthDP: I think we don't need the example here. Let's just inform that false forces the query to be served by exactly the named index.

QuerthDP: Also look at other multi-line descriptions. Let's use `\` at the end of each so that `\n` is not generated in the OpenAPI.

QuerthDP (changes requested):
This one looks really good. I've left some comments, please don't bother with BM25 for now (please open a Jira ticket if you mind so that we won't forget).
Additionally, I proposed one small change for more Rustic style.
What I don't personally like are those excessive comments in tests, I think those could be shorter but I know you like those so let's leave them. At least we'll clearly know what happens :)

crates/httpapi/src/lib.rs
nyh: The most interesting thing that happened with AI is that I can write code in a language that I can't even read. I have no idea what this ugly default "indirection" does, or even what the word "serde" (!?) means, but it works ;-)

crates/httpapi/src/lib.rs
nyh: Ok. Seems the AI shares my love of long, pedagogical, comments, or somehow picked up this love from me. I'll try to reduce the amount and length of comments I write in this project.

crates/httpapi/src/lib.rs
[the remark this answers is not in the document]
nyh: Ok. I'll remove the BM25 option. You asked me to open an issue, but what issue? We can't ask fix the routing to allow making it optional, if BM25 didn't implement routing at all yet. Is there an existing issue about routing in BM25 I should add a comment to?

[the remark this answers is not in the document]
QuerthDP: VECTOR-754

Let's comment here

api/openapi.json
```diff
@@ -1127,6 +1037,10 @@
           "query": {
             "type": "string",
             "description": "The text query to search for."
+          },
+          "routing": {
```
knowack1: BM25 at this stage does not use routing, so I would not add this option here.

api/openapi.json
```diff
@@ -789,6 +695,10 @@
           "limit": {
             "$ref": "#/components/schemas/Limit"
           },
+          "routing": {
```
knowack1, line 792: Indeed, the behavior of this endpoint is unintuitive - or, I would even say, incorrect.
The API explicitly targets a specific index, while others indexes could actually be used to respond to this query.

Moreover, the implementation currently returns a 404 if the specified index does not exist, even though another index that can satisfy this query may still be present. I see that the reason for this is that we need a specific index definition to discover the column, because it is not present in the query itself.

Hence, I think this API should respond using the specific index as requested.

However, To maintain backward compatibility, we could create a sibling v2 endpoint:
/api/**v2**/indexes/{keyspace}/{index}/ann

For requests that will be routed:
/api/v1/tables/{keyspace}/{table}/{column}/ann

And mark /api/v1/indexes/{keyspace}/{index}/ann as deprecated.

I'm not sure if the effort is worth it compared to the proposed single-knob solution, but maybe more correct.

@nyh  @QuerthDP  What are your thoughts on this?

QuerthDP: I think that the current behavior is correct only because it is tightly bound to how the Scylla behaves. 
From the outside perspective (looking only at the Vector Store), I can agree it may feel strange that we query the VS with a specific index and then respond from other.
As I would like to interpret that as "search in the {index} or any better that could also serve that query".
And with routing off - "search in this specific {index}".
In that way I think this is ok, and what Nadav did here is IMO a really good way to keep the old behavior not changed, without any required changes to CQL.

Although I also like the proposition that we could've introduced the `/api/v2/indexes/{keyspace}/{index}/ann` for explicit index search (what Alternator wants to do now, and what routing = false achieves).
However we should keep the behavior unchanged for the CQL path. And introduce the `/api/v1/indexes/{keyspace}/{table}/{column}/ann` with deprecation of `/api/v1/indexes/{keyspace}/{index}/ann` in a separate PR. We would have to decide how would we handle the deprecation. When could it be dropped?
We don't have any ScyllaDB-VectorStore dependency matrix or anything yet.

What I don't like although, is that we introduce `api/v2`. That could be misleading that we only have one function attached to it, and that we still add new endpoints in the `api/v1`, but maybe that's how it should be done, I don't know.

Apart from that, the new endpoint here would also mean some not small amount of duplication I suppose and some boilerplate, which we do omit by introducing the flag.

To sum up, I'd say that I will agree on both options, but I'd like to have it decided consciously.
From my perspective, option with new endpoint is slightly better work of art, but creates more troubles with deprecation.

QuerthDP: > /api/v1/tables/{keyspace}/{table}/{column}/ann

BTW, why tables and not indexes? Do we list tables anyhow now?

nyh: I'm not sure what to say here.

As you know, initially I felt like @knowack1's comment, that this a broken API: You ask to query a specific index given its name, and get response from a completely different index.

But @QuerthDP explained to me that CQL **wants** this seemingly broken behavior, to allow for transparently switching between indexes with deliberately different parameters - and because the Scylla side is missing code to pick the right index, we do in the vector store.

I don't see how we can deprecate the old behavior unless you want to add the missing code to pick the right index to the Scylla side.

This is something you'll need to decide on what you guys want to do. In the mean time I just want a non-broken API where if you give a specific index name, this name is obeyed, so I added this "routing" option exactly as @QuerthDP suggested.

api/openapi.json
nyh: I'll remove.

crates/httpapi/src/lib.rs
nyh: I see that all other comments are a single long line. Wow, I'm not impressed with Rust :-( I'll make it a single long (but not as long as it was) line.

crates/httpapi/src/lib.rs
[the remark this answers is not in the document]
nyh: Added test ann_with_routing_field_omitted_routes_to_newest_serving_index

crates/httpapi/src/lib.rs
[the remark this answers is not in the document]
nyh: Done - removed BM25 changes in this PR, and commented on that issue.

crates/vector-store/src/indexes.rs
nyh: Done.

api/openapi.json
knowack1: Sure, let's proceed with reviewing the change according to @QuerthDP suggestion. Maybe we should create a jira issue for v2 API -  and  maintain there a backlog of breaking changes that we would like to introduce in the future, starting with the incorrect behavior of `/api/v1/indexes/{keyspace}/{index}/ann`.

nyh:
Pushed a new version addressing review comments (see also individual replies to these comments).

QuerthDP (approved):

QuerthDP:
CI failing due to known issue: VECTOR-877.
Restarting.

knowack1 (approved):

PR #571, opened 2026-08-25
vector-store: honor the analyzer and positions options of FTS indexes

QuerthDP:
ScyllaDB lets a full-text index choose its `analyzer` and whether token positions are stored, and records both in the index schema. The Vector Store ignored them. Every FTS index was built with a fixed standard-analyzer pipeline and positions always on, and the API reported that hardcoded pair whatever the index had actually been created with. An index created with `'analyzer': 'english'` was searched as if it were a standard one.

Both options are now read from the index schema and applied to the index the service builds, so the analyzer decides the token pipeline and positions are stored only when the index asks for them. The analyzer is registered per index, which makes queries and highlights tokenize the way the indexed documents did, and the API reports what the index actually carries. Dropping positions makes phrase queries against that index fail with a 400, which is what the option means.

Fixes: [VECTOR-788](https://scylladb.atlassian.net/browse/VECTOR-788)

[VECTOR-788]: https://scylladb.atlassian.net/browse/VECTOR-788?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/crates/validator/src/fts.rs
+++ b/crates/validator/src/fts.rs
@@ -6,7 +6,9 @@
 use crate::TestActors;
 use crate::common::*;
 use async_backtrace::framed;
+use httpapi::FulltextIndexOptions;
 use httpapi::IndexInfo;
+use httpapi::IndexOptions;
 use httpapi::KeyspaceName;
 use httpclient::HttpClient;
 use scylla::client::session::Session;
@@ -111,6 +113,17 @@ impl Fixture {
     #[framed]
     async fn create_fts_index(&self) {
         let index = create_fts_index(&self.session, &self.clients, &self.table).await;
+        self.wait_for_index_on_all_nodes(index).await;
+    }
+
+    #[framed]
+    async fn create_fts_index_with_options(&self, options: &[(&str, &str)]) {
+        let index =
+            create_fts_index_with_options(&self.session, &self.clients, &self.table, options).await;
+        self.wait_for_index_on_all_nodes(index).await;
+    }
+
+    async fn wait_for_index_on_all_nodes(&self, index: IndexInfo) {
         for client in &self.clients {
             wait_for_index(client, &index).await;
         }
@@ -173,8 +186,19 @@ async fn create_fts_index(
     clients: &[HttpClient],
     table: &TableName,
 ) -> IndexInfo {
+    create_fts_index_with_options(session, clients, table, &[]).await
+}
+
+async fn create_fts_index_with_options(
+    session: &Session,
+    clients: &[HttpClient],
+    table: &TableName,
+    options: &[(&str, &str)],
+) -> IndexInfo {
     create_index(
-        CreateIndexQuery::new(session, clients, table, "content").index_type("fulltext_index"),
+        CreateIndexQuery::new(session, clients, table, "content")
+            .index_type("fulltext_index")
+            .options(options.iter().copied()),
     )
     .await
 }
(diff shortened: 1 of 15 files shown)


crates/vector-store/src/db.rs
```diff
@@ -917,30 +930,12 @@ impl Statements {
...
-            let space_type = options
-                .remove("similarity_function")
-                .and_then(|s| s.parse().ok())
-                .unwrap_or_default();
-            let quantization = options
-                .remove("quantization")
-                .and_then(|s| s.parse::<Quantization>().ok())
-                .unwrap_or_default();
+        Ok(options.map(|options| {
+            let connectivity = parse_index_option(&options, "maximum_node_connections");
```
knowack1: Consider adding types explicitly here. Swapping the variable order in the returned tuple will compile seamlessly, leading to incorrect parameter parsing.

crates/vector-store/src/db.rs
```diff
@@ -946,6 +1004,23 @@ impl Statements {
...
 
+    async fn get_fts_index_params(
+        &self,
+        keyspace: KeyspaceName,
+        table: TableName,
+        index: IndexName,
+    ) -> GetFtsIndexParamsR {
+        let options = self.get_index_options(keyspace, table, index).await?;
+        Ok(options.map(|options| {
+            let analyzer = parse_index_option(&options, "analyzer");
```
knowack1: Same here: consider explicit types .

crates/vector-store/src/monitor_indexes.rs
```diff
@@ -271,6 +271,18 @@ async fn build_vs_index_kind(
     })))
 }
 
+async fn build_fts_index_kind(db: &Sender<Db>, idx: &DbCustomIndex) -> anyhow::Result<IndexKind> {
+    let options = db
+        .get_fts_index_params(idx.keyspace.clone(), idx.table.clone(), idx.index.clone())
+        .await
+        .inspect_err(|err| warn!("unable to get index params: {err}"))?
```
knowack1: Consider distinguishing between fts/vector index, as exactly the same log occurs in vs case.

crates/vector-store/tests/integration/index_info.rs
```diff
@@ -89,9 +91,22 @@ async fn index_info_reports_vector_index_options() {
 }
 
 #[rstest]
+#[case::defaults(IndexOptionsFts::default(), "standard", true)]
+#[case::non_defaults(
+    IndexOptionsFts {
```
knowack1, line 96: Nice feature!

crates/vector-store/src/fts_index/factory.rs
```diff
@@ -3,13 +3,26 @@
...
 use crate::IndexKey;
+use crate::Positions;
 use crate::fts_index::actor::FtsIndex;
 use crate::table::Table;
 use std::sync::Arc;
 use std::sync::RwLock;
 use tokio::sync::mpsc;
 
+#[derive(Clone, Debug)]
+pub(crate) struct FtsIndexConfiguration {
```
knowack1, line 16: I'm not sure why we need a new struct (instead of reusing `IndexOptionsFts`), but I see that this follows the pattern from vs_index.

crates/vector-store/src/fts_index/tantivy.rs
```diff
@@ -165,26 +174,67 @@ impl IndexState {
...
-    let stop_words = StopWordFilter::new(Language::English)
-        .ok_or_else(|| anyhow!("fts: english stop words unavailable"))?;
+fn stop_words(language: Language) -> anyhow::Result<StopWordFilter> {
+    StopWordFilter::new(language)
+        .ok_or_else(|| anyhow!("fts: stop words unavailable for {language:?}"))
+}
+
+/// Builds the token pipeline of an analyzer. The same pipeline runs over the
+/// indexed documents and over the queries, so both sides produce matching tokens.
+fn build_analyzer(analyzer: Analyzer) -> anyhow::Result<TextAnalyzer> {
```
knowack1: build_analyzer takes analyzer, at first glance this looks wired. Consider either:
1. rename parameter to analyzer_name
2. rename function to build_analyzer_pipeline or build_token_pipeline or something like this.

crates/vector-store/src/fts_index/tantivy.rs
```diff
@@ -1238,4 +1327,133 @@ mod tests {
...
+    #[test]
+    fn standard_analyzer_does_not_stem() {
+        assert_eq!(
+            tokenize_with_standard_analyzer("the running runners"),
+            vec!["running", "runners"]
+        );
+    }
+
+    #[test]
+    fn english_analyzer_stems_and_removes_stop_words() {
```
knowack1: Why there is dedicated test case for English instead of being included in language_analyzers_stem_and_remove_stop_words?

crates/vector-store/src/fts_index/tantivy.rs
```diff
@@ -1238,4 +1327,133 @@ mod tests {
...
+    #[test]
+    fn whitespace_analyzer_keeps_case_and_punctuation() {
+        assert_eq!(
+            tokenize_with(Analyzer::Whitespace, "The Running, Runners!"),
+            vec!["The", "Running,", "Runners!"]
+        );
+    }
+
+    #[rstest]
+    #[case::stored(true, Some(1))]
```
knowack1: Consider splitting this into two dedicated test cases. Assertions containing `if` statements are hard to read. In general, a parameterized test whose body contains `ifs` driven by parameter values probably shouldn't be a parameterized test.
This will lead to some duplication, however this can be fixed by helper function. Or we can accept duplication.

crates/validator/src/fts.rs
```diff
@@ -671,6 +695,29 @@ async fn fts_index_on_int_column_returns_error(actors: Arc<TestActors>) {
     info!("finished");
 }
 
+#[e2etest::test(group = fts)]
+async fn fts_index_with_unsupported_options_returns_error(fixture: Arc<Fixture>) {
```
knowack1: Why this test has been added before fts_large_document_set, while other tests after? What is the logic behind?

knowack1:
Very nicely structured patch set. It was a pleasure reviewing this commit by commit!

crates/vector-store/src/fts_index/factory.rs
QuerthDP: `FtsIndexConfiguration` isn't the option set, it's the factory's input - it also carries `IndexKey`, which isn't an index option and doesn't belong in `IndexOptionsFts`. Passing `IndexOptionsFts` directly would just move the key to a second parameter, and the two would drift apart as the factory grows inputs that aren't schema options.

crates/vector-store/src/fts_index/tantivy.rs
QuerthDP: Renamed to `build_token_pipeline`

crates/validator/src/fts.rs
QuerthDP: No logic behind, moved.

crates/vector-store/src/fts_index/tantivy.rs
QuerthDP: It was added in contrast to the standard analyzer test that does not stem words, but it should be added to other language analyzers I suppose, so that I moved that there.

crates/vector-store/src/fts_index/tantivy.rs
QuerthDP: I implemented it at 2 separate tests initially, but I thought that we don't need the `true` version, since it probably is already tested somewhere else. That's why I merged them into one to contrast both boolean options. But I can agree, it looks better as short separate tests with a helper, so I changed to that version.

crates/vector-store/src/monitor_indexes.rs
QuerthDP: Added index type in log message

QuerthDP:
Changelog:
- moved `fts_index_with_unsupported_options_returns_error` to the end of file
- added explicit types for index params creation
- renamed `build_analyzer` to `build_token_pipeline`
- merged `english_analyzer_stems_and_removes_stop_words` into `language_analyzers_stem_and_remove_stop_words`
- split `phrase_query_needs_positions` into two separate `true/false` tests

Akvear (approved):

knowack1 (approved):
Thanks for addressing/answering comments. Approving.

knowack1 (approved):
Thanks for addressing/answering comments. Approving.

crates/vector-store/src/fts_index/factory.rs
knowack1: That is true. I was thinking more about something like:
```
FtsIndexConfiguration {
  key: Key,
  options: IndexOptionsFts
}
```
But this is minor, lets proceed as is.

PR #572, opened 2026-08-25
vector-store: allow local/filtered indexes on Alternator tables

nyh:
When a vector index was created on an Alternator vector column, the implementation knows it is not a real CQL column - and needs to be found in the ":attr" map of attributes.

Howver, the index's "target", instead of being given as a single vector column, can list tc,pk,fc - target column, partition key and filtering columns respectively. In this case, the vector search reaches a separate code path - that for some reason includes extra validation (that did not exist in bare single-column target case) that all these columns are real CQL columns.

This patch avoids this extra validation in the case of an Alternator table - these columns may very well NOT be real columns.

We'll need this change for Alternator's support for DynamoDB's new API's "SearchSchema", where INLINE_FILTER and HASH attributes can be specified, which correspond to our "pk" and "fc" respectively.

from_target_option() now takes an is_alternator flag (derived from the keyspace name) and skips the target-column existence/type check and the pk-columns existence check when set, leaving validation for CQL-native tables unchanged.

### DIFF
--- a/crates/vector-store/src/db.rs
+++ b/crates/vector-store/src/db.rs
@@ -791,9 +791,10 @@ impl Statements {
                                 .context(InvalidMetadata)
                         })?;
                     let partition_key_count = NonZeroUsize::new(table.partition_key.len()).unwrap();
+                    let is_alternator = KeyspaceName::from(&keyspace_name).is_alternator();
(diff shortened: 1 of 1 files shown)


nyh:
Closing in favor of a PR with more necessary fixes

PR #575, opened 2026-08-27
vector-store: read vectors from ScyllaDB in the DiskANN ScyllaProvider

Akvear:
This PR implements an actual `VectorSource` for the ScyllaDB provider introduced in #567 that actually fetches the vectors directly from the base table. 

## Benchmark results without any major optimizations: ScyllaDB-backed DiskANN provider vs in-mem DiskANN vs USearch

**Setup.** `openai_small_50k` (50,000 × 1536-dim OpenAI embeddings, cosine,
1,000 test queries with exact-KNN ground truth). Single-node ScyllaDB in Docker
plus a host-run `vector-store`, both on one machine. Identical index options
for every engine:

{'similarity_function': 'cosine', 'maximum_node_connections': 16,
 'construction_beam_width': 128, 'search_beam_width': 64}
| | DiskANN / Scylla | DiskANN / in-mem | USearch |
| --- | --- | --- | --- |
| Index build (50k) | **1417 s** | 16.9 s | 14.3 s |
| recall@10 (avg) | **98.6–99.0** | 98.8 | 98.0 |
| recall@100 (avg) | **95.2** | 95.2 | 93.1 |
| Serial p50 (k=10) | **166 ms** | 4.6 ms | 1.1 ms |
| Serial p99 (k=10) | **229 ms** | 7.0 ms | 2.0 ms |
| Peak QPS (CQL) | **12.1** | 1,912 | 4,317 |

Fixes: VECTOR-831


crates/vector-store/src/vs_index/diskann/scylla.rs
```diff
@@ -0,0 +1,975 @@
...
+
+/// A [`VectorSource`] reading vectors back from the index's base table.
+pub(super) struct BaseTableSource<T> {
+    table: Arc<RwLock<T>>,
+    db_index: mpsc::Sender<DbIndex>,
+}
+
+impl<T> BaseTableSource<T> {
+    #[allow(dead_code, reason = "wired into DiskannIndexFactory in a follow-up")]
+    pub(super) fn new(table: Arc<RwLock<T>>, db_index: mpsc::Sender<DbIndex>) -> Self {
```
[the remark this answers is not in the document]
Akvear: TBD

crates/vector-store/src/vs_index/diskann/scylla.rs
```diff
@@ -0,0 +1,975 @@
...
+            source,
+        }
+    }
+}
+
+impl DiskannBackend for ScyllaBackend {
+    type Provider = ScyllaProvider;
+    type Strategy = ScyllaStrategy;
+
+    fn create_index(
```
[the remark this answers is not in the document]
Akvear: TBD

crates/vector-store/src/vs_index/diskann/scylla.rs
```diff
@@ -0,0 +1,975 @@
...
+                    fetched.len(),
+                    len = ids.len(),
+                );
+                return vec![None; ids.len()];
+            }
+            Err(err) => {
+                warn!(
+                    "BaseTableSource::get: failed to read {requested} vectors \
+                     for partition {partition_id:?}: {err:#}; skipping {len} candidates",
+                    len = ids.len(),
```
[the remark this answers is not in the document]
Akvear: I am aware of this. This was chosen right now as the easy option that works just fine most of the time. Keeping this open to remember this.

swasik:
So this is without any cache? I am sure after few nights of AI trying to optimize it we can have much better results and adding local cache will make it even faster. But as for now - great starting point.

Akvear:
> So this is without any cache? 

There is a small cache for vectors that are being inserted, so they are kept in memory for the duration of the insert and dropped after.

Akvear:
| | Earlier run | **This run** | change |
| --- | --- | --- | --- |
| Index build | 1417 s | **494 s** | 2.9× faster |
| Serial QPS (k=10) | 6.1 | **40.2** | 6.6× |
| Serial p50 | 166 ms | **24.0 ms** | 6.9× faster |
| Serial p99 | 229 ms | **42.7 ms** | 5.4× |
| Peak QPS | 12.1 | **48.3** | 4× |
| recall@10 / @100 | 98.6 / 95.2 | 98.6–98.9 / 95.3 | unchanged |

This is a different run that was made after some AI optimizations to minimize allocation when reading rows from the query result.

swasik:
> This is a different run that was made after some AI optimizations to minimize allocation when reading rows from the query result.

Amazing - especially latency starts to look reasonably. Two more rounds like this and we are ready to launch :-)

Akvear:
Opus about memory:
| | Index cost (Δ) | Peak | After search |
| --- | --- | --- | --- |
| **DiskANN / Scylla** | **168 MiB** | 292 MiB | 445 MiB |
| **DiskANN / in-mem** | **393 MiB** | 526 MiB | 525 MiB |
| **USearch** | **506 MiB** | 597 MiB | 550 MiB |

2.3× less than in-mem DiskANN, 3× less than USearch — the resident f32 copy
(300 MiB) is gone, only the graph stays in RAM.

Most of that 168 MiB still isn't the index: the graph measures **246 B/node**,
i.e. ~16 MiB total. The rest is live per-query fetch memory — `expand_beam`
loads a whole round at once, so a query holds up to
`beam_width × degree × d × 4B` ≈ 6.8 MiB at `search_beam_width=64`. It should not scale with
dataset size — it scales with concurrency and dimensionality instead.

So: `graph (~250 B/point, linear in n)` + `concurrency × beam × degree × d × 4B
(flat in n)`. At 50k the second term dominates, so 168 MiB overstates
per-vector cost by ~10×.

TLDR: Measuring memory would require testing on bigger dataset, as right now per-query memory dominates the graph structure size.

swasik:
> TLDR: Measuring memory would require testing on bigger dataset, as right now per-query memory dominates the graph structure size.

Plus to get really good results here we have to add PQ.

crates/vector-store/src/config_manager.rs
```diff
@@ -459,11 +459,11 @@ pub async fn load_config(env: impl Fn(&str) -> anyhow::Result<String>) -> anyhow
...
-    config.use_diskann = env("VECTOR_STORE_USE_DISKANN")
-        .unwrap_or("false".into())
-        .trim()
-        .parse()
-        .map_err(|_| anyhow!("Unable to parse VECTOR_STORE_USE_DISKANN env (bool)"))?;
+    if let Ok(diskann_backend) = env("VECTOR_STORE_DISKANN_BACKEND") {
+        config.diskann_backend = Some(diskann_backend.trim().parse().map_err(|_| {
+            anyhow!("Unable to parse VECTOR_STORE_DISKANN_BACKEND env (inmem/scylla)")
+        })?);
+    }
```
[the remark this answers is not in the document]
Akvear: not necessary, diskann is not production ready and only tests use this env var

crates/vector-store/src/db_index.rs
```diff
@@ -418,6 +466,52 @@ impl Statements {
...
+            .execute_unpaged(&self.st_fetch_vector, &key)
+            .await?
+            .into_rows_result()?;
+        let Some(row) = rows_result.maybe_first_row::<Row>()? else {
+            // The row is gone from the base table.
+            return Ok(None);
+        };
+        parse_vector_row(row.columns)
+    }
+
```
QuerthDP: It feels like it's done really weird.

First we split into one-vector-per-scylla-query. Then we fetch maybe the first row (there is at most 1), and then we iterate through it's columns (??, how are 2 of them here?) and take only the first one which assumably is the vector we wanted. I think this can be done in a more elegant way. I don't know how, but Rust Driver has some features for it.

crates/vector-store/src/vs_index/diskann/scylla.rs
```diff
@@ -127,6 +133,72 @@ pub(super) trait VectorSource: Debug + Send + Sync + 'static {
...
+                 for partition {partition_id:?}"
+            )
+        })?;
+
+        if fetched.len() != requested {
+            bail!(
+                "BaseTableSource::get: expected {requested} vectors, got {}",
+                fetched.len()
+            );
+        }
```
QuerthDP: I think I've seen that we just silently skipped not fetched vectors before, and now we panic? Why don't we panic before?

crates/vector-store/src/db_index.rs
```diff
@@ -401,6 +432,22 @@ impl Statements {
                 stmt
             });
 
+        let query = db_index_backend::request_query(
```
ewienik, line 435: It should be different query than query for getting vectors, timestamps and columns.

crates/vector-store/src/db_index.rs
```diff
@@ -418,6 +466,52 @@ impl Statements {
...
+                        .await
+                        .context("fetch_vectors: permit semaphore closed")?;
+                    self.fetch_vector(session, key).await
+                }
+            })
+            .buffered(concurrency_limit())
+            .collect()
+            .await;
+
+        fold_fetched_vectors(results)
```
ewienik, line 496: using
```
futures::stream::iter(keys)                                         
    .map(|key| {                                                    
...                                
    .try_collect()                                                  
    .await                                                          
```

there is no need for fold_fetched_vectors

crates/vector-store/src/vs_index/diskann/mod.rs
```diff
@@ -110,6 +111,7 @@ impl VsIndexFactory for DiskannIndexFactory {
         &self,
         index: VsIndexConfiguration,
         table: Arc<RwLock<Table>>,
+        db_index: mpsc::Sender<DbIndex>,
```
ewienik, line 114: consider adding new parameter for new_index_factory_diskann instead of changing this trait

crates/vector-store/tests/integration/db_basic.rs
```diff
@@ -471,6 +498,13 @@ pub(crate) fn new_db_index(
 
     let (tx_index, mut rx_index) = mpsc::channel(10);
     let (tx_embeddings, rx_embeddings) = mpsc::channel(10);
+    // Only a vector index feeds the base-table vectors read back by
+    // DbIndex::GetVectors; an FTS index stores documents instead.
+    let tx_embeddings = if metadata.vs().is_some() {
+        record_streamed_vectors(db.clone(), tx_embeddings)
```
ewienik, line 504: There should be additional parameter to DbMock - storing vectors shouldn't be default

crates/vector-store/src/vector.rs
```diff
@@ -60,6 +67,77 @@ impl TryFrom<CqlValue> for Vector {
...
+    pub(crate) fn into_vector(self) -> Vector {
+        self.0
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+#[error("embedding column: {0}")]
+struct VectorColumnError(String);
+
+impl<'frame, 'metadata> DeserializeValue<'frame, 'metadata> for VectorColumn {
```
ewienik, line 84: There is a special CqlValue::Vector. What is the reason for the new type?

ewienik (changes requested):

crates/vector-store/src/vs_index/diskann/mod.rs
Akvear: No `DbIndex` exists when `new_index_factory_diskann` is called. I think there should also be a db_index per index, not per factory.

ewienik: You're right - while run creates diskann factor db_index is unknown yet. There could be passed db and get db_index, but get_db_index is rather create-db-index. To support passing index in factory we should add here engine -> but there will be cyclic dependency (engine depends on db_index and vs_index and vs_index depends on db_index). So better is to stay with passing db_Index by index_create method.

crates/vector-store/src/db_index.rs
Akvear: `try_collect()` would fail all futures if one failed. `fold_fetched_vectors()` serves an additional purpose – failed read turns into None, while letting other queries finish. Only if all queries fail, an Error is returned.

crates/vector-store/src/vector.rs
Akvear: We skip a few unnecessary allocations. We decode the bytes returned from the database straight into a shape that the DiskANN Scylla Provider needs (which is the VS Vector), without the intermediate CqlValue::Vector. From testing I can tell that this is quite a substantial improvement. 

Instead of having a VectorColumn we can just deserialize into Vector.

Akvear:
Changes:
- added separate query for fetching vectors on the DiskANN path
- store incoming vectors in DbMock only if specified

crates/vector-store/src/db_index.rs
ewienik: So you can filter map errors without need of special fold_feched vectors.

crates/vector-store/src/vector.rs
ewienik: I understand that you mean the coversion from CqlValue::Vector(CqlValue). This seems orthogonal issue - we have the same problem in our current code using `TryFrom<CqlValue> for Vector`. Consider creating additional issue to solve performance problem instead of fixing it in these functionality changes. We need to think if we want to fix it in scylla driver or vector-store (more probably in scylla driver).

Akvear:
Changes:
- dropped last commit
- moved fold_fetched_vectors error handlng logic into fetch_vectors()

crates/vector-store/src/db_index.rs
```diff
@@ -230,7 +247,7 @@ pub(crate) async fn new(
                     msg = rx_index.recv() => {
                         match msg {
                             Some(msg) => {
-                                tokio::spawn(process(Arc::clone(&statements), msg, completed_scan_length.clone()));
+                                tokio::spawn(process(Arc::clone(&statements), msg, completed_scan_length.clone(), Arc::clone(&fetch_permits)));
```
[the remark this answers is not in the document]
Akvear: It's bounded by the callers.

ewienik (approved):

crates/vector-store/src/db_index.rs
```diff
@@ -418,6 +467,56 @@ impl Statements {
...
+
+    async fn fetch_vector(
+        &self,
+        session: Arc<Session>,
+        key: PrimaryKey,
+    ) -> anyhow::Result<Option<Vector>> {
+        let rows_result = session
+            .execute_unpaged(&self.st_fetch_vector, &key)
+            .await?
+            .into_rows_result()?;
```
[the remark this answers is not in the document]
ewienik: It could be done in a separate PR as diskann is a research project now. It needs more integration in validator.

PR #576, opened 2026-08-28
vector-store: add fbin dataset support to the latte workload

DarioMirovic:
The latte vector search workload reads datasets only in its own text format, prepared offline from the source datasets. For big-ann-benchmarks datasets that conversion is overhead: the source files are already a well-defined binary format, and the text copies are larger on disk and slower to parse. Add native fbin/ibin support to the workload scripts. The engine side is already released in latte `0.50.0-scylladb`.

The first two commits fix the merged workload: local index verification now probes every stratum, and query/ground-truth files are validated at prepare time.

Measured on openai_small_50k (50k x 1536):

| | text | fbin | fbin_packed |
|---|---|---|---|
| dataset parse (prepare) | 45.2 s | 15.3 s | 3.7 s |
| peak RSS | 6.39 GB | 6.38 GB | 1.60 GB |
| load throughput, no index | n/m | 6,157 op/s | 17,903 op/s (2.9x) |
| load throughput, index maintained | 5,417 op/s | 5,364 op/s | 11,476 op/s (2.1x) |
| recall after load | — | 0.9811 | 0.9816 (parity) |

Stored vectors are bit-exact against the source fbin in all formats.

Datasets are still read into memory up front; streaming during load is a separate follow up.

Fixes: VECTOR-759


DarioMirovic:
Question - should the first two commits be extracted in a separate PR, due to this rule: `Single Change: PR should consist of a single business change.`?

DarioMirovic:
> 🟡 Other comments (1)
> > latte/vector-search/recall_buckets.rn-63-65 (1)> `63-65`: _🎯 Functional Correctness_ | _🟡 Minor_ | _⚡ Quick win_
> > > **Reject non-positive `BUCKET_COUNT`.**
> > > When `LOCAL_INDEX` is true and `bucket_count=0`, Line 188 evaluates `i % BUCKET_COUNT` and fails with division by zero. Negative values also produce invalid bucket IDs. Reject `BUCKET_COUNT <= 0` during `prepare`.
> > > 
> > > Proposed fix
> > > ```diff
> > >  pub async fn prepare(db) {
> > > +    if LOCAL_INDEX && BUCKET_COUNT <= 0 {
> > > +        return Err(`bucket_count must be greater than 0`);
> > > +    }
> > >      if LOAD_DATA {
> > > ```

Done in `prepare` `BUILD_INDEX` branch.

> > > Also applies to: 184-188

Non-positive value shouldn't be possible here.

DarioMirovic:
V2:
- Fixed handling of non-positive `BUCKET_COUNT` values

DarioMirovic:
@swasik @m-szymon @ewienik can you review please?

latte/vector-search/fbin_dataset.rn
```diff
@@ -0,0 +1,77 @@
+//! Binary (fbin/ibin) dataset loaders for vector-search workloads.
+//!
+//! A benchmark dataset is three files in the big-ann-benchmarks binary
+//! formats:
+//!
+//!   base vectors  - fbin: header `count: u32 LE, dimension: u32 LE`, then
+//!                   count × dimension f32 LE; ids are the record indices.
```
ewienik: nit: non-ascii (utf-8) character

latte/vector-search/fbin_dataset.rn
```diff
@@ -0,0 +1,77 @@
...
+    let ids_only = 8 + count * k * 4;
+    let with_distances = 8 + count * k * 8;
+    if file.len() != ids_only && file.len() != with_distances {
+        return Err(`${path}: file is ${file.len()} bytes, expected ${ids_only} \
+            (ids) or ${with_distances} (ids + distances) for ${count} x ${k}`);
+    }
+    let rows = [];
+    for _ in 0..count {
+        rows.push(file.read_i32_vec(fs::LE, k)?);
+    }
```
ewienik, line 89: Is with_distances supported? If so, there seems to be only reading ids_only

ewienik (dismissed):

latte/vector-search/fbin_dataset.rn
DarioMirovic: Fixed, using ascii now

latte/vector-search/fbin_dataset.rn
DarioMirovic: Not right now. It can be a nice small addition as a separate PR. Currently no workload consumes it. The asymmetry also isn't new - fbin does not support buckets, while text mode does. This is caused by the definitions of our text and official big-ann-benchmarks fbin input formats.

DarioMirovic:
V3:
- Replaced non-ascii with ascii symbols

latte/vector-search/fbin_dataset.rn
```diff
@@ -0,0 +1,91 @@
...
+    let k = file.read_u32(fs::LE)?;
+    let ids_only = 8 + count * k * 4;
+    let with_distances = 8 + count * k * 8;
+    if file.len() != ids_only && file.len() != with_distances {
+        return Err(`${path}: file is ${file.len()} bytes, expected ${ids_only} \
+            (ids) or ${with_distances} (ids + distances) for ${count} x ${k}`);
+    }
+    let rows = [];
+    for _ in 0..count {
+        rows.push(file.read_i32_vec(fs::LE, k)?);
```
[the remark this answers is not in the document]
DarioMirovic: Fixed

DarioMirovic:
V4:
- Ground truth neighbor IDs are decoded as `u32`, per the big-ann spec
- `read_fbin_header` returns an object instead of tuple; removes three Rune warnings "Pattern might panic"

DarioMirovic:
@ewienik I addressed the comments. Can you check, re-approve, so we can merge in this sprint?

latte/vector-search/fbin_dataset.rn
ewienik: Now `with_distance` is checked for validation, but reading rows is supported only for u32 (`ids_only`). There will be wrong values in case of file `with_distance`. Consider removing validating `with_distances` or finish with some kind of **unsupported** information in case of `with_distance`

DarioMirovic: The ids are correct for both layouts. in the big-ann file the ids are one contiguous block right after the header (count * k * u32) and the distances block (count * k * f32) is appended after the ids block, not interleaved per row. Reading count * k u32 values from offset 8 (after header) reads exactly the ids in either layout. The trailing distances block is never read. We just make sure that such a file, with distances, is not discarded when there is no need. This does mean the classes of tests that need distances for their logic are not currently supported, but the existing tests needing only ids will not fail on a file that has everything they need, and more.

ewienik: > The ids are correct for both layouts. in the big-ann file the ids are one contiguous block right after the header (count * k * u32) and the distances block (count * k * f32) is appended after the ids block, not interleaved per row.

Ok, I didn't understand that there are two blocks (first ids, next distance). I was thinking that the layout is [id, distance) * k.

ewienik (approved):

ewienik:
ci failed with flaky validator tests that was supposed to  be fixed with VECTOR-877. Restarting ci

PR #578, opened 2026-08-30
vector-store: fix indexing bug for filtering on a primary-key column

nyh:
This patch fixes a bug where vector indexing silently failed if vector-store filtering was enabled on one of the base table's primary key columns.

For example,
```
	CREATE TABLE ks.tbl (
	    pk INT,
	    ck INT,
	    f  INT,
	    v  VECTOR<FLOAT, 3>,
	    PRIMARY KEY (pk, ck)
	);

	CREATE CUSTOM INDEX idx ON ks.tbl (v, ck, f) USING 'vector_index';
```

Did not index anything, as the indexing code got confused on what it is trying to insert into the index:

engine.rs's add_index() passed the raw, as-declared list of filtering columns straight into the index's in-memory state, unchanged. Separately, the code that fetches each row's actual values (in db_index.rs and db_cdc/consumer.rs) already leaves out any filtering column that is also a primary-key column, since such a column's value comes from the primary key itself, not from a fetched value.

That mismatch meant two lists could end up different lengths: the stored filtering-column names (unfiltered) versus the values actually fetched per row (already filtered). Table::upsert() pairs these up one by one, name with value, in order. Once the lengths diverged, that pairing shifted, so a value meant for one column got attributed to another. Best case, the shifted value landed on a primary-key column and the whole row failed to index ("cannot insert value into PrimaryKey column"); worst case, it landed on a different real column and silently corrupted that column's stored value instead.

I originally discovered this bug while writing test for the new Alternator vector store API with different combinations of base table keys and vector index filtering keys, but as can be seen in the example above - the bug actually affects CQL as well. Any CQL vector index, local or global, that reuses a primary-key column as a filtering column hits this bug.

Fixed by adding IndexMetadata::nonpk_filtering_columns(), used by engine.rs (the actual fix, via a new build_table() helper factored out of add_index() for testability) and substituted for the equivalent ad-hoc filters already present in db_index.rs/db_cdc/consumer.rs (pure dedup there, no behavior change, but removes the risk of the two lists drifting apart again).

This patch also adds unit tests, and an end-to-end CQL test (validator):

- lib.rs: unit test for IndexMetadata::nonpk_filtering_columns() itself.
- engine.rs: unit test driving build_table() with a raw, undeduped filtering-column list - the same shape real index metadata has - and checking that upserting a row lands the target vector and every filtering column's value in the right place. Verified this test fails against the pre-fix code (panics with "Cannot insert value into PrimaryKey column") and passes against the fix.
- validator (crates/validator/src/filtering.rs): a CQL end-to-end regression test, global_index_filter_by_filtering_column_shared_with_primary_key. It creates a table with a composite primary key (pk, ck), a global vector index on "v" that also filters on "ck" (a clustering-key column) and "f" (an ordinary column), inserts a single row, and checks the row comes back from an ANN query restricted by both "ck" and "f". Run against a real ScyllaDB + Vector Store cluster: on the pre-fix code every upsert into the index fails ("Cannot insert value into PrimaryKey column" in the Vector Store log), the index reports as fully built with 0 rows indexed, and the ANN query silently returns nothing; with the fix, the row is indexed and returned correctly.

Fixes VECTOR-892

### DIFF
--- a/crates/validator/src/filtering.rs
+++ b/crates/validator/src/filtering.rs
@@ -748,6 +748,75 @@ async fn local_index_filter_by_filtering_columns(actors: Arc<TestActors>) {
     info!("finished");
 }
 
+/// Regression test for VECTOR-892: a vector index can declare a filtering
+/// column that is also one of the base table's primary-key columns. But
+/// vector-store used to misalign the stored column values in that case:
+/// every upserted row failed with "cannot insert value into a PrimaryKey
+/// column", so the index stayed fully "built" at 0 rows and ANN queries
+/// against it silently returned nothing.
+///
+/// The order of the declared filtering columns matters: "ck" (the
+/// primary-key column) must come before "f" for the misalignment to
+/// manifest - the other order happens to leave "f" correctly aligned and
+/// just drops "ck", which produces no visible symptom for this test.
+#[e2etest::test(group = filtering)]
+async fn global_index_filter_by_filtering_column_shared_with_primary_key(actors: Arc<TestActors>) {
+    info!("started");
+
+    let (session, clients) = prepare_connection(&actors).await;
+
+    let keyspace = create_keyspace(&session).await;
+    let table = create_table(
+        &session,
+        "pk INT, ck INT, f INT, v VECTOR<FLOAT, 3>, PRIMARY KEY (pk, ck)",
+        None,
+    )
+    .await;
+
+    session
+        .query_unpaged(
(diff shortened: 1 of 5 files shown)


nyh:
Note: Although I framed this bug as affecting CQL to get it merged quickly, it actually bothered me while I was writing tests for **Alternator** (the new vector-search API which supports filtering columns), and while I was writing an extensive test suite involving all sorts of combinations of key columns and filtering columns, I ran across this bug.

ewienik (approved):

swasik:
Thanks @nyh for contribution!

PR #517, opened 2026-07-13
vector-store: implement local vector indexes based on nonpk columns

ewienik:
This PR adds support for nonpk partition key columns in local vector indexes. The solution is based on passing such column values using DbIndexedOperation::Upsert(Box<[DbIndexedValue]>) - the first are values for DbIndexedValue::Vector() or DbIndexedValue(Document) - target values to build index. The second are values for nonpk partition key column. The third are filtering values.

The PartitionKey for Index is built in table module - from values coming in DbIndexedRow. db_index and db_cdc actors retrieve additional data during fullscan or query after CDC. Table is responsible for synchronizing data incoming from fullscan or cdc (data coming from monitor_items).

This PR contains also refactor of IndexMetadata - to store primary_key_columns and partion_key_count (instead of calling for them from db_index actor).

This PR adds unit tests and validator e2e tests for new functionality.

Fixes: VECTOR-721

PR #580, opened 2026-08-30
ann: optionally return stored column values alongside primary keys

nyh:
The vector store already copies ("projects") some non-key columns into a vector index. We currently use those stored values for filtering, but this patch wants to allow a vector search to **optionally** return those values with the search results - without a separate base-table lookup per result.

This is a feature that Alternator wants: The new DynamoDB vector-search API only allows vector searches to retrieve projected columns (including filtering columns), and the feature we add here will make this efficient. CQL won't use the new feature yet, but it could also use it in the future: a SELECT that only needs key and filtering-column values could be served entirely from the index, making it significantly more efficient.

Until now there was no way to get these stored values back out of the vector store: ann()/filtered_ann() only ever returned primary keys and a similarity score, so every caller had to go back to the base table, even for columns already stored in the vector store.

This patch adds an optional `return_columns` request field and a matching `column_values` response field to the ANN HTTP API, implemented in all three index backends (usearch, OpenSearch and DiskANN). A column with no stored value for a given row comes back as null rather than being omitted, so the array of values stays aligned with the primary keys and scores.

Covered by a table-level unit test for the underlying column lookup, plus integration tests against all three backends: usearch and DiskANN in the vs_index suite (usearch also covers the filtered_ann() path, which DiskANN and OpenSearch don't support), and OpenSearch against its mock server.

Fixes VECTOR-893


ewienik (commented):
The single commit seems hard to be reviewed as it mix at least three things: HTTP api change, vs_index API change, implementation of returning stored columns. Consider splitting this commit.

Currently this commit can improve alternator path queries when user wants only primary key or filtering columns (or non-pk indexing columns). I doubt how many use cases it covers IMHO filtering is needed for retrieval of other values.

Additionally we need to provide two paths in scylla - one when we can get all columns from vector-store, second when at least one column is out of vector-store knowledge. vector-store client in scylla needs to be more complicated to cover all cases.

There is another possible way to avoid vector-store client complications, by assuming that vector-store always returns all column values (it asks scylladb itself for values when something is missing). But in this way we need different HTTP api.

IMHO it is still open question if we should return column values from vector-store and how we should do this.

nyh:
> The single commit seems hard to be reviewed as it mix at least three things: HTTP api change, vs_index API change, implementation of returning stored columns. Consider splitting this commit.
> 
> Currently this commit can improve alternator path queries when user wants only primary key or filtering columns (or non-pk indexing columns). I doubt how many use cases it covers IMHO filtering is needed for retrieval of other values.

By "currently" you mean the existing Alternator vector search API?
The existing Alternator will not use this new option, and never will. It could, but we won't be working on it. I already wrote the new Alternator API - https://github.com/scylladb/scylladb/pull/31181. It has full filtering support (both filtering columns and partition columns, using the new DynamoDB API) and it always uses the new return-column feature of the vector store: The new DynamoDB API allows retrieving in a vector search **only** columns that are stored in the vector index (key columns, partition/filter columns, and "projected" columns - the latter are implemented as filter columns which don't have a type and don't allow filtering).

> Additionally we need to provide two paths in scylla - one when we can get all columns from vector-store, second when at least one column is out of vector-store knowledge. vector-store client in scylla needs to be more complicated to cover all cases.

It doesn't need to be - but it **can** be. This patch adds a feature that CQL could use, in certain case, to become significantly more efficient in certain cases. Yes, it will require a bit more coding on the ScyllaDB CQL side. But it will be worth it - because the vector index already stores these columns, and it's a real shame not to use this data if it's available.

> There is another possible way to avoid vector-store client complications, by assuming that vector-store always returns all column values (it asks scylladb itself for values when something is missing). But in this way we need different HTTP api.

This will have terrible performance, and is exactly what I don't want to do.
I actually have in Alternator already an implementation that retrieves from the vector index only keys - and then retrieves from Scylla all the requested columns for the keys. This implementation fully works, but it's inefficient. The whole point of this patch is to provide a way to make it more efficient if we want. An in Alternator, we want **now**. In CQL, this requirement hasn't yet surfaced. I'm pretty sure it will once users start to use it. The fact that DynamoDB outright forbids you from selecting any column not stored in the vector index suggests users will never want to use the slow approach.

> IMHO it is still open question if we should return column values from vector-store and how we should do this.

The question of "if we should return" has an easy answer: make it a request option. The "user" of the vector store (Alternator or CQL) will say if they want to get the columns (and if so, which columns) back together with the results - or not. This is what this patch does.

This patch also answers "how we should do this", giving a specific API. If you think it's not a good answer, let me know, I can easily change the API or anything that bothers you on this approach.

nyh:
> The single commit seems hard to be reviewed as it mix at least three things: HTTP api change, vs_index API change, implementation of returning stored columns. Consider splitting this commit.

I'll split it later to these three parts, if this is what you feel are the natural layers of the vector store.

To be honest what bothers me personally about vector store PRs like this one are the **tests** which are mixed with the code, and even require adding various imports etc. to real code. Does this not bother you? Are you ok that the tests are inside the regular patches, mixed with the regular implementation code? If not I can split that too. But maybe that's overdoing the split?

QuerthDP:
> To be honest what bothers me personally about vector store PRs like this one are the **tests** which are mixed with the code, and even require adding various imports etc. to real code. Does this not bother you? Are you ok that the tests are inside the regular patches, mixed with the regular implementation code? If not I can split that too. But maybe that's overdoing the split?

I'd say that this is even better that those are mixed with the implementation. Each commit is it's atomic part that compiles and passes all tests (including new unit tests just added for the implementation).

Of course there are Validator tests, and mostly integration tests that could be added separately, but as for unit tests it's good place for them to be added alongside the implementation IMO.

Also, when only the tests needs an import, then it should be imported in the tests module, not the whole module with implementation.

ewienik:
> > The single commit seems hard to be reviewed as it mix at least three things: HTTP api change, vs_index API change, implementation of returning stored columns. Consider splitting this commit.
> 
> I'll split it later to these three parts, if this is what you feel are the natural layers of the vector store.
> 
> To be honest what bothers me personally about vector store PRs like this one are the **tests** which are mixed with the code, and even require adding various imports etc. to real code. Does this not bother you? Are you ok that the tests are inside the regular patches, mixed with the regular implementation code? If not I can split that too. But maybe that's overdoing the split?

Some PRs provides tests with the implementation in single commits. Some others not (ie #517). With bigger changes I put tests separately (even I split unit tests with validator tests). We don't have a strict rule for this, so fill free to separate test implementation as different commits. The good rule is that each commit should compile (`cargo check`) - and this sometimes depends on test changes (imports or small refactoring). I found that refactoring by moving files sometimes break this rule for compiling each commit to support better preserving git history.

Mixing unit tests with the code is an idiomatic way for Rust and I got used to it. Integration tests are in separate directory and end to end tests in separate crate.

ewienik:
> The new DynamoDB API allows retrieving in a vector search **only** columns that are stored in the vector index (key columns, partition/filter columns, and "projected" columns - the latter are implemented as filter columns which don't have a type and don't allow filtering).

I understand the reason now. This is really important information. I didn't catch "non-filtering-projected" columns. If we put "projected" columns into index options we will store them in the memory of vector-store. And this seems the biggest pain point - we copy all data from scylla table, not only needed columns (primary-key, non-pk local index columns, filtering columns; we use them for processing searches) but also values from columns only for responding from vector-store. IIUC such values could be big as they provide business values for customers and we will store them in memory, which seems not optimal.

nyh:
> > The new DynamoDB API allows retrieving in a vector search **only** columns that are stored in the vector index (key columns, partition/filter columns, and "projected" columns - the latter are implemented as filter columns which don't have a type and don't allow filtering).
> 
> I understand the reason now. This is really important information. I didn't catch "non-filtering-projected" columns. If we put "projected" columns into index options we will store them in the memory of vector-store. And this seems the biggest pain point - we copy all data from scylla table, not only needed columns (primary-key, non-pk local index columns, filtering columns; we use them for processing searches) but also values from columns only for responding from vector-store. IIUC such values could be big as they provide business values for customers and we will store them in memory, which seems not optimal.

As an Alternator-only extension, I want to provide the extra feature - which I already have coded because this is how the previous API worked (and also how LSIs work in DynamoDB) - that Alternator fetches only the primary keys from the vector store and then fetches the rest of the desired columns from the base table. 

Users will be able to use this Alternator-only extension to, indeed, save vector-index memory by not storing copies of "projected" columns there.

But, as I said, the official DynamoDB API does not support this. You must "project" any column you want to retrieve in vector searches into the vector index. I implemented this in our vector store by putting these "projected" (but non-filtered) columns in "fc" with a "blob" type (they just contain the JSON representation of the DynamoDB value, of any type).

I'll make it a point to include in the Alternator API re-implementation, right from the first version, the "SELECT" (the ability to read unprojected columns from the base table) extension to the API, since you consider it important to customers and we already had this feature in the previous API that I invented (inspired by the CQL API we had).

In any case I think there's a value of supporting **both** options: Some users will want to quickly fetch attributes already stored in the vector store for other purposes (i.e., filtering) or even add to the vector store additional columns not needed for filtering (i.e., projection), wanting better performance - but other users will worry less about performance and more about memory use, and will not want to project any unnecessary column values into the vector store.

crates/vector-store/src/httproutes.rs
```diff
@@ -884,15 +911,35 @@ async fn post_index_ann(
...
+                                Vec<Option<Value>>,
+                            > = return_columns
+                                .iter()
+                                .map(|col_name| {
+                                    let values = column_values_per_row
+                                        .iter()
+                                        .map(|row_vals| {
+                                            row_vals
+                                                .get(col_name)
+                                                .and_then(|v| try_to_json(v.clone()).ok())
```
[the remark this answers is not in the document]
nyh: The INET bug is being fixed separately in https://github.com/scylladb/vector-store/pull/574.

I would like to argue that your request to convert bad data found in the index into an error - i.e., failure of the entire ANN request - is a mistake. Perhaps (?) it makes sense to log these cases so we know we have a problem, but I think it is bad practice that some bug in validating PutItem (for example) allowing you to accidentally insert a bad item, could result in all vector searches failing when they encounter this bad item on every search.

There is a possible middle ground: We **could** , instead of changing a bad attribute value to a missing value (which you didn't like),  we drop the entire **item** (row) from the ANN results, but not abort the entire ANN with an error. But I'm not thrilled by this option: Remember that the ANN already found a valid nearest neighbor, and it has a valid primary key. The only thing it failed to do is to retrieve one of its attributes which had a "bad" value stored for it. Why shouldn't we return the correct primary key? What is the benefit of dropping this row from the search results?

nyh:
As requested in @ewienik's review, I split the patch into three, with clear commit messages explaining what each one dones). The resulting code is identical (git diff of the original branch and the result of the three commits is empty), but should hopefully be easier to review.

ewienik:
> In any case I think there's a value of supporting **both** options: Some users will want to quickly fetch attributes already stored in the vector store for other purposes (i.e., filtering) or even add to the vector store additional columns not needed for filtering (i.e., projection), wanting better performance - but other users will worry less about performance and more about memory use, and will not want to project any unnecessary column values into the vector store.

Currently I see gravity for optimizing memory consumption rather than latency (as the nodes for vector-search for big indexes are really expensive). Nevertheless, let's add this feature as it shouldn't harm performance, since the return of column values are optional and if customer asks for such improvement it will be done. The tradeoff is we will need to support this API in the future, but currently it seems not a big cost as we need to return primary keys + distances.

crates/vector-store/src/table/mod.rs
```diff
@@ -1274,6 +1286,27 @@ impl TableSearch for Table {
             }
         }
     }
+
+    #[hotpath::measure]
+    fn column_values_for(
```
QuerthDP, line 1291: Unused this commit

crates/vector-store/src/vs_index/diskann/mod.rs
```diff
@@ -504,22 +509,28 @@ where
...
+        let col_vals = if return_columns.is_empty() {
+            BTreeMap::new()
+        } else {
+            table.column_values_for(partition_id, primary_id, &return_columns)
+        };
+        primary_keys.push(primary_key);
+        distances.push(distance);
+        column_values_vec.push(col_vals);
+    }
+    Ok((primary_keys, distances, column_values_vec))
```
QuerthDP, line 533: Could this be done using `process_results`?

crates/httpapi/src/lib.rs
```diff
@@ -374,6 +374,10 @@ pub struct PostIndexAnnRequest {
     /// By default, the query may be routed to a different, better-matching index on the same column. Set to `false` to force the query to be served by exactly the index named in the URL.
     #[serde(default = "default_routing")]
     pub routing: bool,
+    /// Filtering-column names whose stored values should be returned alongside
+    /// the primary keys. Empty (the default) means return no column values.
```
QuerthDP, line 378: If you make it multiline descriptions, use `\` at the end.
Otherwise it is rendered with `\n` in OpenAPI, which we don't want.

crates/httpapi/src/lib.rs
```diff
@@ -385,6 +389,14 @@ pub struct PostIndexAnnResponse {
     pub primary_keys: HashMap<ColumnName, Vec<Value>>,
     pub distances: Vec<Distance>,
     pub similarity_scores: Vec<SimilarityScore>,
+    /// Per-column stored values for the filtering columns requested in
+    /// `return_columns`. Each entry maps a column name to a Vec of
+    /// `Option<Value>` — one entry per returned nearest neighbour, in the
+    /// same order as `similarity_scores`. `None` means the value was not
+    /// present for that row (e.g. the attribute did not exist when the row
+    /// was indexed). Absent when `return_columns` was empty.
```
QuerthDP, line 397: Ditto

crates/vector-store/src/httproutes.rs
```diff
@@ -820,9 +820,26 @@ async fn post_index_ann(
             ))
             .await;
 
-        // The HTTP API doesn't let callers request stored column values back
-        // yet, so ann()/filtered_ann() are never asked for any.
```
QuerthDP, line 824: If you remove this comment now, then there's no reason to add it before I think.

crates/vector-store/src/httproutes.rs
```diff
@@ -894,15 +911,35 @@ async fn post_index_ann(
...
-                                distances: distances.into_iter().map(|d| d.into()).collect(),
-                                similarity_scores,
-                            }),
-                        )
-                            .into_response(),
+                        Ok(primary_keys) => {
+                            let column_values: HashMap<
+                                httpapi::ColumnName,
+                                Vec<Option<Value>>,
+                            > = return_columns
```
QuerthDP, line 918: Such a weird formatting. But that's what `rustfmt` did so it must stay.

crates/httpapi/src/lib.rs
```diff
@@ -374,6 +374,10 @@ pub struct PostIndexAnnRequest {
     /// By default, the query may be routed to a different, better-matching index on the same column. Set to `false` to force the query to be served by exactly the index named in the URL.
     #[serde(default = "default_routing")]
     pub routing: bool,
+    /// Filtering-column names whose stored values should be returned alongside
+    /// the primary keys. Empty (the default) means return no column values.
+    #[serde(default)]
+    pub return_columns: Vec<ColumnName>,
```
ewienik, line 380: As this is backward compatible improvement the openapi minor version (httproutes.rs:103) should be increased to 3.1.0

crates/vector-store/tests/integration/opensearch.rs
QuerthDP, line 1: I'm wondering if we really need to add the opensearch tests. It's never used.

QuerthDP (dismissed):
Leaving some minor comments, but generally looks fine.

crates/vector-store/src/httproutes.rs
```diff
@@ -820,9 +820,26 @@ async fn post_index_ann(
...
+                timer.observe_duration();
+
+                let msg = format!(
+                    "Column '{column}' in return_columns is not part of the primary key or filtering columns, and cannot be returned"
+                );
+                debug!("post_index_ann: {msg}");
+                return (StatusCode::BAD_REQUEST, msg).into_response();
+            }
+        }
+        let return_columns: Arc<[crate::ColumnName]> = return_columns.into();
```
ewienik, line 842: nit: consider early return from collecting list. It could be something along the lines of:
```
let return_columns: Result<Arc<[_]>, _> = request
    .return_columns                              
    .into_iter()                                 
    .map(crate::ColumnName::from)                
    .map(|col| {                                 
        if primary_key_columns.contains(&col) {  
            Ok(col)                              
        } else {                                 
            Err(col)                                                     
        }                                        
    })                                           
    .collect();                                  
let return_columns = match return_columns        
{                                                
        Ok(cols) => cols,                        
        Err(column) => {                         
        //...
        }
};

```

crates/vector-store/src/httproutes.rs
```diff
@@ -894,15 +911,35 @@ async fn post_index_ann(
...
+                        Ok(primary_keys) => {
+                            let column_values: HashMap<
+                                httpapi::ColumnName,
+                                Vec<Option<Value>>,
+                            > = return_columns
+                                .iter()
+                                .map(|col_name| {
+                                    let values = column_values_per_row
+                                        .iter()
+                                        .map(|row_vals| {
```
ewienik, line 923: this is conversion from Vec<BTreeMap<ColumnName, Value>> into HashMap<ColumnName, Vec<Option<Value>>>. What about returning BTreeMap<ColumnName, Vec<Option<Value>>> (or HashMap) directly from index?

crates/vector-store/src/vs_index/actor.rs
```diff
@@ -12,10 +13,17 @@ use crate::PrimaryKey;
...
+use std::collections::BTreeMap;
+use std::sync::Arc;
 use tokio::sync::mpsc;
 use tokio::sync::oneshot;
 
-pub(crate) type AnnR = anyhow::Result<(Vec<PrimaryKey>, Vec<Distance>)>;
+pub(crate) type AnnR = anyhow::Result<(
+    Vec<PrimaryKey>,
+    Vec<Distance>,
+    Vec<BTreeMap<ColumnName, CqlValue>>,
```
ewienik, line 25: What about returning BTreeMap<ColumnName, Vec<Option<_>>> to avoid allocation BTreeMap for every PrimaryId and further recollecting?

crates/vector-store/src/table/mod.rs
```diff
@@ -1274,6 +1286,27 @@ impl TableSearch for Table {
...
+        &self,
+        partition_id: PartitionId,
+        primary_id: PrimaryId,
+        columns: &[ColumnName],
+    ) -> BTreeMap<ColumnName, CqlValue> {
```
ewienik, line 1296: What about `columns: ColumnName,) -> Option<CqlValue>`? If so, map could be allocated once per answer in vs_index.

ewienik (changes requested):
I have only small questions about collecting and recollecting maps and vectors.

crates/vector-store/src/vs_index/actor.rs
QuerthDP: I think this could work. Seems like a good idea.

crates/vector-store/src/table/mod.rs
nyh: Of course it's not yet used, I split the single patch into three implementation stages, as you (I believe) requested - this patch only adds this function, and the next patch uses it.  The commit message explains that.

Also, this unused function is not even a git-bisect build error - the function is not completely unused - it's used by its own unit test.

crates/vector-store/src/table/mod.rs
nyh: Good idea! I'll do that. I can actually simplify the API here considerably - There isn't even a need to return many columns, we can just do it one at a time. Bunching them all up in one map just made the efficincy worse (as you noticed), not better.

crates/vector-store/src/vs_index/actor.rs
nyh: Yes, good idea! The next version will have here `HashMap<ColumnName, Vec<Option<CqlValue>>>`

crates/vector-store/src/vs_index/diskann/mod.rs
nyh: Yes, but why? It wasn't used before this patch. Claude is weakly against it:

"It's a valid alternative, but not a clear win: it doesn't change behavior, doesn't reduce lines meaningfully, and introduces a closure indirection plus a new-to-this-codebase itertools idiom (nothing else here uses process_results, whereas the plain result? inside a for loop is the standard, more immediately-readable pattern and appears elsewhere in this codebase in similar shapes). "

crates/httpapi/src/lib.rs
nyh: I'll make it not multi-line :-(

crates/httpapi/src/lib.rs
nyh: Done. I didn't guess that it needs to be done. Maybe there should be a comment on the structure I changed that points people who change this structure to this version thing?

crates/httpapi/src/lib.rs
nyh: Yes, fixed too.
This is really a bug in this Rust tool, but I guess I won't be fixing that bug so just need to work around it :-(

crates/vector-store/src/httproutes.rs
nyh: The AI was trying to make the individual patches compile so the previous patch needed to pass some extra parameters, and this comment expained why they were empty in that version.
I agree it's not super-important, but since the AI already did this, I see no reason to remove the commend and leave the empty array (which I would have to leave).

crates/vector-store/src/httproutes.rs
nyh: What's the benefit? (by the way, the example has an error, primary_key_columns is the wrong thing). Efficiency it does **not** improve.

But I'll change it just so it looks more similar to other adjacent code...

crates/vector-store/src/httproutes.rs
nyh: Yes. I am doing exactly this in the new version I'll push in a few minutes.

crates/vector-store/tests/integration/opensearch.rs
nyh: I don't know. My AI found we have opensearch, usearch, and diskann. I had no idea one is "never used". Is it really **never** uses? If that's true, perhaps it should be removed? Or, if it's not removed, we need to keep it up-to-date and tested, so we do need these tests, no?

nyh:
Pushed a new version answering (I hope) all the reviewer comments.

The biggest change is a overhaul of the low-level API (the first patch) and its use in the second patch. Instead of a lookup function returning a map per-row (forcing us to do an allocation per-row), the new implementation has a function looking up just a single value - and this function is called to fill pre-allocated vectors of values and doesn't need to do any allocation per-row.

crates/vector-store/src/httproutes.rs
```diff
@@ -805,6 +805,34 @@ async fn post_index_ann(
...
 
+        // Filtering-column names whose stored values the caller wants back
+        // alongside the primary keys, instead of having to separately read
+        // every result row from the base table.
+        let return_columns: Result<Arc<[crate::ColumnName]>, crate::ColumnName> = request
+            .return_columns
+            .into_iter()
+            .map(crate::ColumnName::from)
+            .map(|column| {
+                if filtering_columns.contains(&column) {
```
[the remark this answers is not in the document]
nyh: What a mess. I'll fix, perhaps not exactly in the way you suggested because it **breaks** CQL's routing - which deliberately wants to use **the newest index**, even if its configuration changed.

Instead, I'll just drop the check that return_columns can only use valid filtering columns. A user could now ask for any column, and get nulls if it's missing in the index. The front-end can avoid sending such unprojected columns in the request - it knows which columns are projected.

crates/vector-store/src/httproutes.rs
```diff
@@ -869,15 +903,29 @@ async fn post_index_ann(
...
+                                column_values
+                                    .into_iter()
+                                    .map(|(col_name, values)| {
+                                        let values = values
+                                            .into_iter()
+                                            .map(|v| v.and_then(|v| cql_types::to_json(v).ok()))
+                                            .collect();
+                                        (col_name.into(), values)
+                                    })
+                                    .collect();
```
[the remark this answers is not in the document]
nyh: Right. I'll fix.

nyh:
Pushed a new version addressing copilot's and coderabbit's review.

**Copilot's review**: There's one fairly noticeable change in this version: If you ask the vector index to return column "x" and column x is not stored in the vector index, the request is **no longer rejected with an error** as it was in the previous version. Insted, now you will silently get null values for "x" in all rows. This is the same behavior as you would get when asking for column "x" that is projected in the index but doesn't have any values.

The reason for this change is support for the "index routing" feature in CQL. If we have an old index which allows filtering on column "x" and a new index which doesn't have filtering for column "x", we want to use the new index - but the validation will be done on one of the indexes on random. So it's better to just not do any validation. The search will succeed in any case, and will just return null values for x if it uses the  new index.

**Coderabbit's review** found an unrelated minor bug in the opensearch backend. Added a fix for it in the fourth patch. This fix has nothing to do with the previous patches, except that the problem is in the same area of the code that the previous patches modified, so coderabbit caught this bug in its review.

crates/vector-store/src/httproutes.rs
ewienik: IMHO there is a light improvement in efficiency - single iteration vs double iteration. In case of error the first iteration is cancelled instead.

crates/vector-store/tests/integration/opensearch.rs
ewienik: I wouldn't put a lot of effort to the opensearch as it is not used in production now. If it is too much burden to maintain it we'll remove it.

ewienik (dismissed):
LGTM, thanks for updates.

@QuerthDP Do you have more comments?

api/openapi.json
```diff
@@ -806,6 +813,18 @@
           "similarity_scores"
         ],
         "properties": {
+          "column_values": {
+            "type": "object",
+            "description": "Per-column stored values for the columns requested in `return_columns`. Each entry maps a column name to a Vec of `Option<Value>` — one entry per returned nearest neighbour, in the same order as `similarity_scores`. `None` means the value was not present for that row (e.g. the attribute did not exist when the row was indexed, or the column isn't tracked by the index). Absent when `return_columns` was empty.",
```
QuerthDP: I'd rather say "in the same order as `primary_keys`" as the primary keys are the main returned values, but this is also clear and correct.

crates/vector-store/src/httproutes.rs
```diff
@@ -853,7 +860,7 @@ async fn post_index_ann(
                     (StatusCode::INTERNAL_SERVER_ERROR, msg).into_response()
                 }
             },
-            Ok((primary_keys, distances, _column_values_per_row)) => {
```
QuerthDP: Why name it with the "per_row" suffix just to drop it in next commit?

crates/vector-store/tests/integration/opensearch.rs
QuerthDP: I still think that those tests are completely unnecessary. We're much closer to removing the OpenSearch than making it work properly. But since there are here already, let them stay.

crates/vector-store/tests/integration/routing.rs
```diff
@@ -285,6 +287,106 @@ async fn ann_routes_to_newest_serving_index() {
...
+/// return_columns must never influence routing: a request for a column
+/// that only an older index in the routing group projects must still be
+/// routed to the newest serving index, not fall back to the older one
+/// just because it happens to still track that column. The newer index
+/// simply reports the column as null, the same as any other column it
+/// doesn't track.
+#[rstest]
+#[timeout(Duration::from_secs(10))]
+#[tokio::test]
+#[cfg_attr(not(feature = "slow-test-hooks"), ignore = "requires slow-test-hooks")]
```
QuerthDP, line 321: Please verify if this is needed at all. As from my experience AI likes to add those fixture requirement while it's completely unnecessary and the test works just fine without it. And I think it is the case here as well, since we're not using the `internal` counters, for which the fixture was introduced.

crates/vector-store/tests/integration/vs_index.rs
```diff
@@ -444,6 +446,257 @@ async fn failed_db_index_create(#[case] config: Config) {
...
+        response.primary_keys.get(&"pk".into()).unwrap(),
+        &[Value::from(1)]
+    );
+    assert_eq!(
+        response.column_values.get(&"color".into()).unwrap(),
+        &[Some(Value::from("red"))]
+    );
+}
+
+/// Regression test: filtering by a primary-key value that was never
```
QuerthDP: Regression test what for? As I understand it's a regression test comparing to some of your previous implementations of this PR. I don't think it should be called like that

QuerthDP (dismissed):
Looks fine. Leaving some comments, I think none of those are blockers so I approve, but it would be nice to respond. Especially please verify the `slow-test-hook`.

crates/vector-store/src/vs_index/diskann/mod.rs
QuerthDP: It was used before your change in all 3 backends that you changed. Nothing surprising that the AI cannot find it anymore, since we wiped those out with the change :) 

I don't have any strong feelings about which version is better, yours looks just fine. I wanted only to mention that the `process_results` version was used before and it did their job quite well.

api/openapi.json
nyh: Done.

crates/vector-store/src/httproutes.rs
nyh: I don't remember. Does it matter, if we drop it in the next commit?

crates/vector-store/tests/integration/opensearch.rs
nyh: I don't know if it's "too much burden to maintain it", but my AI did **not** pick up on the fact that it is dead weight, and did spent a few tokens on doing everything also for it - including these tests. I don't think this is bad, considering I spent more effort discussing this comment than to actually write those tests in the first place, which the AI just did for me.

crates/vector-store/tests/integration/routing.rs
nyh: I have no idea what you are asking here :-(

My AI claims it is necessary:

Good question, but I checked and the guard is needed here. ann_return_columns_does_not_affect_routing calls assert_ann_served_by(), which is exactly the mechanism that relies on the counter fixture: it registers an ann-served-request--{keyspace}--{index} counter via internals_start_counter(), runs the request, then asserts the counter incremented by 1 to prove which index actually served it.

That counter is only ever incremented at one place, httproutes.rs's post_index_ann(), and that increment call is itself gated behind #[cfg(feature = "slow-test-hooks")]. Without the feature enabled, the increment is compiled out entirely, the counter never moves, and assert_eq!(after - before, 1, ...) inside assert_ann_served_by fails on every run — it's not that the test degrades gracefully, it can't pass at all.

So this isn't the same situation as the internal-counters case you've seen elsewhere — assert_ann_served_by is built directly on those counters, and it's the test's core assertion (which index served the request), not an unrelated leftover. I'll leave the #[cfg_attr] as-is.

crates/vector-store/tests/integration/vs_index.rs
nyh: "Regression test" doesn't mean it checks an earlier bug. Rather, it means it's a test that is supposed to be re-run after changing implementation to validates that a particular feature still works - i.e., it doesn't "regress". This is, in my opinion, the "opposite" of a unit test - a test which may become irrelevant after the code changes because the details of the code it checks changed.

In any case, I don't think the word "regression test" beginning this comment is helpful. I think my AI wrote it like that and it just didn't bother me so I left it alone. I'll edit it now and remove this "regression test" phrase and in general remove half the weird text in this comment.

crates/vector-store/tests/integration/vs_index.rs
```diff
@@ -467,6 +469,255 @@ async fn failed_db_index_create(#[case] config: Config) {
...
+        .collect();
+    assert_eq!(color_by_pk[&1].as_deref(), Some("red"));
+    assert_eq!(color_by_pk[&2], None);
+    assert_eq!(color_by_pk[&3].as_deref(), Some("blue"));
+}
+
+#[rstest]
+#[timeout(Duration::from_secs(10))]
+#[tokio::test]
+async fn ann_return_columns_null_for_column_not_tracked_by_index() {
```
ewienik, line 553: Shouldn't this test cover all three engines?
```
#[case::usearch(usearch_test_config())]
#[case::diskann(diskann_test_config())]
#[case::diskann_scylla(diskann_scylla_test_config())]
```

crates/vector-store/tests/integration/vs_index.rs
```diff
@@ -467,6 +469,255 @@ async fn failed_db_index_create(#[case] config: Config) {
...
+    assert_eq!(response.status(), StatusCode::OK);
+    let response: PostIndexAnnResponse = response.json().await.unwrap();
+    assert_eq!(
+        response.column_values.get(&"not_a_real_column".into()),
+        Some(&vec![None])
+    );
+}
+
+#[tokio::test]
+async fn ann_filtered_ann_returns_requested_column_values() {
```
ewienik, line 596: Shouldn't this test cover all three engines?
```
#[case::usearch(usearch_test_config())]
#[case::diskann(diskann_test_config())]
#[case::diskann_scylla(diskann_scylla_test_config())]
```

crates/vector-store/tests/integration/vs_index.rs
```diff
@@ -467,6 +469,255 @@ async fn failed_db_index_create(#[case] config: Config) {
...
+/// Filtering by a primary-key value that was never indexed (here, pk =
+/// 999, when only pk = 1 was ever inserted) is a well-formed query that
+/// matches zero rows - it's not the same as "no such partition exists".
+/// `column_values` must still come back present, with "color" mapped to
+/// an empty list, rather than being built as an empty map and dropped
+/// from the JSON response by serde's `skip_serializing_if` on an empty
+/// HashMap: the documented contract says `column_values` is absent only
+/// when `return_columns` is empty, which isn't the case here.
+#[tokio::test]
+async fn ann_filtered_ann_returns_empty_column_values_for_partition_with_no_rows() {
```
ewienik, line 668: Shouldn't this test cover all three engines?
```
#[case::usearch(usearch_test_config())]
#[case::diskann(diskann_test_config())]
#[case::diskann_scylla(diskann_scylla_test_config())]
```

crates/vector-store/tests/integration/vs_index.rs
```diff
@@ -467,6 +469,255 @@ async fn failed_db_index_create(#[case] config: Config) {
...
+    let response: PostIndexAnnResponse = response.json().await.unwrap();
+
+    assert!(response.primary_keys.get(&"pk".into()).unwrap().is_empty());
+    assert_eq!(response.column_values.get(&"color".into()), Some(&vec![]));
+}
+
+#[rstest]
+#[case::usearch(usearch_test_config())]
+#[case::diskann(diskann_test_config())]
+#[tokio::test]
```
ewienik, line 720: Shouldn't this test cover all three engines?
```
#[case::usearch(usearch_test_config())]
#[case::diskann(diskann_test_config())]
#[case::diskann_scylla(diskann_scylla_test_config())]
```

ewienik (dismissed):
If think tests in vs_index should cover all three vector-search engines. Shouldn't it?

crates/vector-store/tests/integration/vs_index.rs
ewienik: As @QuerthDP is on PTO IMHO it could be resolved

crates/vector-store/tests/integration/routing.rs
ewienik: I see that these tests needs slow-test-hooks as they check if the specific index was used.

crates/vector-store/src/httproutes.rs
ewienik: As @QuerthDP I'm resolving comment

crates/vector-store/tests/integration/opensearch.rs
ewienik: IMHO we can leave it

crates/vector-store/src/vs_index/diskann/mod.rs
ewienik: The issue with process_results is that we need to prepare a dynamic number of vectors for column_values. process_results must have static number of columns, so temporary allocation would be in use.

crates/vector-store/tests/integration/vs_index.rs
nyh: While you're right that for "consistency" with the other tests I should do this, I doubt this is important because what is being tested here doesn't really depend on the backend. But I guess, better safe then sorry - maybe in the future something will change that will break this feature on just one of the backends. So'll fix as you suggested.

crates/vector-store/tests/integration/vs_index.rs
nyh: No, unfortunately not,  only usearch currently implements filtered search. DiskANN answers a `FilteredAnn` message with "DiskANN index does not support filtered search" and OpenSearch with "Filtering not supported". So a `diskann`/`diskann_scylla` case
here would only assert those errors, not the column values.

Sadly, I couldn't find in your test suite any notion of xfail, so I can't add those two other cases with "xfail". So I'll leave just one case, and a comment explaining why the two others are missing.

crates/vector-store/tests/integration/vs_index.rs
nyh: Unfortunately, again, no - it's usearch-only. I'll add a comment here too.

crates/vector-store/tests/integration/vs_index.rs
nyh: Yes, this was some sort of mistake that it was missing here.

ewienik (approved):

PR #581, opened 2026-08-30
Add mechanisms to support filtering columns in Alternator

nyh:
The existing Alternator vector search API does not yet have its version of the new vector store features of filtering columns and partition columns ("local index"). We plan to add support for these features in Alternator as part of implementing the new DynamoDB vector search API (see https://github.com/scylladb/scylladb/pull/31181), so we will need the filtering/partition column code in the vector store to support Alternator tables. 

Unfortunately, there are some holes in this support - whereas the vector store generally supports Alternator tables and their weird requirements (like having "attributes" which are not real columns), some of the filtering code is missing the necessary Alternator support, and this PR adds this missing support, gated by is_alternator() checks.

### DIFF
--- a/crates/vector-store/src/db_index.rs
+++ b/crates/vector-store/src/db_index.rs
@@ -284,6 +284,10 @@ struct Statements {
     nonpk_partition_key_columns: Box<[ColumnName]>,
     filtering_columns: Arc<[ColumnName]>,
     table_columns: GetTableColumnsR,
+    /// Aligned with nonpk_partition_key_columns + filtering_columns: the
+    /// NativeType to decode each one's raw ":attrs" value as, or None for a
+    /// real CQL column. See parse_values().
+    alternator_decode_types: Box<[Option<NativeType>]>,
     st_range_scan: PreparedStatement,
     kind: IndexKind,
 }
@@ -340,18 +344,38 @@ impl Statements {
         let target_columns = metadata.target_columns.clone();
         let filtering_columns: Arc<[_]> = metadata.nonpk_filtering_columns().cloned().collect();
 
+        let real_columns: HashMap<ColumnName, NativeType> = table
+            .columns
+            .iter()
+            .filter_map(|(name, coltype)| {
+                if let ColumnType::Native(typ) = &coltype.typ {
+                    Some((ColumnName::from(name.clone()), typ.clone()))
+                } else {
+                    None
+                }
+            })
+            .collect();
+        // Merge in synthesized types for Alternator attributes with no real
+        // column (see IndexMetadata::alternator_attribute_types); real
+        // columns come last so they win on any name collision.
         let table_columns = Arc::new(
-            table
-                .columns
-                .iter()
-                .filter_map(|(name, coltype)| {
-                    if let ColumnType::Native(typ) = &coltype.typ {
-                        Some((ColumnName::from(name.clone()), typ.clone()))
-                    } else {
-                        None
-                    }
-                })
+            metadata
+                .alternator_attribute_types
+                .keys()
+                .filter_map(|name| Some((name.clone(), metadata.alternator_native_type(name)?)))
+                .chain(
+                    real_columns
+                        .iter()
+                        .map(|(name, typ)| (name.clone(), typ.clone())),
+                )
                 .collect(),
+        );
+        let alternator_decode_types = alternator_decode_types(
+            nonpk_partition_key_columns
+                .iter()
+                .chain(filtering_columns.iter()),
+            &metadata,
+            &real_columns,
         );
         let st_partition_key_list = table
             .partition_key
@@ -389,6 +413,7 @@ impl Statements {
             target_columns,
             filtering_columns,
             table_columns,
+            alternator_decode_types,
             st_range_scan,
             session_rx,
             kind: metadata.kind.clone(),
@@ -569,6 +594,7 @@ impl Statements {
         let target_columns_offset = self.primary_key_columns.len().get();
         let target_columns_len = self.target_columns.len();
         let kind = self.kind.clone();
+        let alternator_decode_types = self.alternator_decode_types.clone();
 
         // wait for an active session
         let session = {
@@ -604,6 +630,7 @@ impl Statements {
                     None,
                     target_columns_len,
                     &kind,
+                    &alternator_decode_types,
                 )
                 .inspect_err(|err| error!("Error while parsing values: {err}"))
                 .ok()?;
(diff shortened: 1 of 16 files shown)


nyh:
I verified that the patches in this PR don't break the existing "validator" tests (which use the features that the old Alternator API needed - and we still need these tests to pass until we replace the API).

crates/vector-store/src/vector.rs
```diff
@@ -69,6 +84,39 @@ const ALTERNATOR_TYPE_JSON: u8 = 4;
 /// matching the CQL `VECTOR<float, N>` on-wire encoding.
 const ALTERNATOR_TYPE_FLOAT32VECTOR: u8 = 5;
 
+/// Decodes an Alternator-encoded scalar attribute value (a filtering or
+/// partition-key column with no real CQL column) into the `CqlValue`
+/// matching `native_type`. Like the vector column above, the value is
+/// tagged; this checks the tag matches `native_type` and re-deserializes
+/// the rest via the CQL driver, rather than reimplementing e.g. decimal
+/// parsing here.
+pub(crate) fn parse_alternator_scalar(
```
ewienik: I see that `parse_alternator_scalar` is not used in `vector.rs` module and it is for parsing additional filtering or partition-key column values. What about adding a commit with refactoring vector.rs and put all alternator parsing into a separate `alternator.rs` module? If so, we won't mix vector functionality with parsing alternator column values.

crates/vector-store/src/lib.rs
```diff
@@ -714,6 +716,11 @@ pub struct IndexMetadata {
     pub target_columns: NonemptyArc<ColumnName>,
     pub partitioning: DbIndexPartitioning,
     pub filtering_columns: Arc<[ColumnName]>,
+    /// For Alternator tables: the declared DynamoDB type ("S", "N" or "B")
+    /// of each partition-key/filtering column with no real column in the
+    /// table's CQL schema (its value lives only in the ":attrs" map).
+    /// Empty for CQL-native tables and for columns that are real columns.
+    pub alternator_attribute_types: Arc<BTreeMap<ColumnName, String>>,
```
ewienik: What about using here NativeType and covert string into NativeType while parsing options? If so, it could be `Arc<BTreeMap<ColumnName, NativeType>>`? IIUC we use alternator type only by converting it into `NativeType`.

crates/vector-store/src/lib.rs
```diff
@@ -772,6 +790,8 @@ pub struct DbCustomIndex {
     pub target_columns: NonemptyArc<ColumnName>,
     pub partitioning: DbIndexPartitioning,
     pub filtering_columns: Arc<[ColumnName]>,
+    /// See IndexMetadata::alternator_attribute_types.
+    pub alternator_attribute_types: Arc<BTreeMap<ColumnName, String>>,
```
ewienik: consider using `NativeType` here instead of `String`.

ewienik (changes requested):

nyh:
I'm working now to address review comment, and also to split the long second patch into two.

crates/vector-store/src/db_index.rs
```diff
@@ -630,11 +643,68 @@ impl Statements {
...
+
+/// Builds parse_values()'s alternator_decode_types parameter: for each of
+/// `columns`, its NativeType from `metadata.alternator_native_type()`.
+pub(crate) fn alternator_decode_types<'a>(
+    columns: impl IntoIterator<Item = &'a ColumnName>,
+    metadata: &IndexMetadata,
+) -> Box<[Option<NativeType>]> {
+    columns
+        .into_iter()
+        .map(|column| metadata.alternator_native_type(column))
```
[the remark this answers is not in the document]
nyh: I'll fix this in the next version I push.

crates/vector-store/src/db_index.rs
```diff
@@ -630,11 +643,68 @@ impl Statements {
...
+        .collect()
```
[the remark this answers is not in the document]
nyh: Wow, this is a very real bug, and not at all specific to Alternator - in CQL we also forgot to normalize the local-index partition keys. I'll produce a bugfix patch and validator test that shows this is a CQL bug as well. I'll need to split that fix into a separate PR, and also create a new issue...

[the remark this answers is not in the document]
nyh: After spending a lot of time and tokens on this, I realize this is part of a bigger problem SCYLLADB-1575 found by @m-szymon - Scylla does NOT normalize numbers used as partition keys (regular base-table partition keys, unrelated to vector store), so 1.0 and 1.00 are written as separate partitions. We have a vector store unit test to check that the vector store does this too: `test_decimal_key`.

However as the Alternator test `test_number_hash_key_representation` shows, this causes an incompatibility with DynamoDB, so needs to be fixed.

What the AI found here was the same thing needs to be done for the local-index partition key. However, it can only be fixed together with SCYLLADB-1575 (for the base-table partition key), and must only be fixed for Alternator tables, otherwise the CQL behavior will also change...

crates/vector-store/src/lib.rs
nyh: I'll do that. I think it will indeed simplify things.

crates/vector-store/src/lib.rs
nyh: Yes. I'll do that. Good idea.

crates/vector-store/src/vector.rs
nyh: I'll do that. This will add another commit to this PR.

nyh:
Pushed a new version addressing the review comments, and hopefully also making review easier by breaking up the biggest patch into two. 
And as requested by a reviewer, added another final patch moving the growing number of Alternator-specific code to a new source file alternator.rs.

ewienik (approved):
Thank you for changes.

I see that we need to probably prepare PR for Rust driver to define Hash on NativeType as building Hash manually can be error-prone in the future, when we can add new fields into IndexMetadata.

crates/vector-store/src/db_index.rs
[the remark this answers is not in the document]
ewienik: As this issue is outside of this PR I'm resolving this comment.

nyh:
> Thank you for changes.
> 
> I see that we need to probably prepare PR for Rust driver to define Hash on NativeType as building Hash manually can be error-prone in the future, when we can add new fields into IndexMetadata.

Yes, my AI just worked around this lack of Hash, and I didn't give it a second thought, sorry.

PR #583, opened 2026-09-01
vector-store: fix cql_cmp() missing Blob, Boolean, Uuid and Timeuuid arms

nyh:
cql_cmp() (in table/mod.rs), used by is_valid_for() to evaluate ANN filter restrictions, had no comparison arm for these four types. Its callers treat a None result as "row does not match", not as an error, so an Eq (or Lt/Lte/Gt/Gte/In) restriction on a blob, boolean, uuid or timeuuid column always silently matched zero rows - with no error and no log line.

This bug is easily reproducable in CQL - e.g., a table with a blob filtered in a vector index - and we include in this patch CQL validator tests that demonstrate this. It was also hit while adding Alternator SearchSchema support for the "B" (binary) attribute type, which reaches the same code path in the vector store.

Fix:
* Blob: byte-wise comparison, matching Cassandra's BytesType.
* Boolean: false < true.
* Timeuuid: the driver's CqlTimeuuid: Ord already implements ScyllaDB's timeuuid ordering exactly.
* Uuid: needs its own uuid_cmp() helper - ScyllaDB's uuid_type_impl orders by version nibble first, and version-1 (time-based) UUIDs by timestamp; a plain 128-bit byte compare disagrees with ScyllaDB. 

We also add regression tests: unit tests in table/mod.rs for all four types, validator (end-to-end) tests in filtering.rs that reproduce the bug with pure CQL ANN queries filtered by each type, and additional validator (end-to-end) tests that check that the comparison order of ANN filtering in Scylla is exactly the same as the comparison order in traditional CQL filtering - by comparing ANN queries to regular queries without ANN.

Fixes VECTOR-889.

### DIFF
--- a/crates/validator/src/filtering.rs
+++ b/crates/validator/src/filtering.rs
@@ -7,10 +7,12 @@ use crate::TestActors;
 use crate::common::*;
 use httpapi::KeyspaceName;
 use scylla::client::session::Session;
+use scylla::value::CqlTimeuuid;
 use std::collections::HashSet;
 use std::net::IpAddr;
 use std::sync::Arc;
 use tracing::info;
+use uuid::Uuid;
 
 e2etest::group!(
     name = filtering,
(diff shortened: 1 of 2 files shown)


crates/validator/src/filtering.rs
```diff
@@ -662,6 +664,271 @@ async fn global_index_filter_by_filtering_columns(actors: Arc<TestActors>) {
     info!("finished");
 }
 
+/// Test ANN search filtered by equality on a BLOB filtering column.
+///
+/// Regression test for VECTOR-889: `cql_cmp()` (in `table/mod.rs`) had no
```
QuerthDP: I would not mention where the `cql_cmp()` is. It provides no information, everyone can search for it by name.

crates/vector-store/src/table/mod.rs
```diff
@@ -1315,8 +1317,27 @@ fn partition_key_from_restrictions(
         })
 }
 
+/// Orders two regular (non-time) UUIDs the way ScyllaDB's `uuid_type_impl`
+/// comparator does (`types/types.cc`'s `compare_visitor`): by version nibble
```
QuerthDP: No need for the file reference imo

crates/vector-store/src/table/mod.rs
```diff
@@ -1315,8 +1317,27 @@ fn partition_key_from_restrictions(
...
+/// first; version-1 (time-based) UUIDs then compare as timeuuids (matching
+/// `utils::timeuuid_tri_compare`); everything else compares by unsigned
+/// bytes. Plain `uuid::Uuid: Ord` is a raw 128-bit byte compare and would
+/// disagree with ScyllaDB across versions - see VECTOR-889.
+fn uuid_cmp(a: &Uuid, b: &Uuid) -> Ordering {
+    let (a, b) = (a.as_bytes(), b.as_bytes());
+    let (va, vb) = ((a[6] >> 4) & 0x0f, (b[6] >> 4) & 0x0f);
```
QuerthDP: I think this could be more legible in separate lines.

crates/vector-store/src/table/mod.rs
```diff
@@ -1315,8 +1317,27 @@ fn partition_key_from_restrictions(
...
+    if va != vb {
+        return va.cmp(&vb);
+    }
+    if va == 1 {
```
QuerthDP: Maybe we could mention that vb == 1 here as well to be more explicit.

crates/validator/src/filtering.rs
```diff
@@ -662,6 +664,271 @@ async fn global_index_filter_by_filtering_columns(actors: Arc<TestActors>) {
...
+            "SELECT pk FROM {table} WHERE f = true ORDER BY v ANN OF [0.0, 0.0, 0.0] LIMIT 10 ALLOW FILTERING"
+        ),
+        &session,
+    )
+    .await
+        .rows::<(i32,)>()
+        .expect("failed to get rows")
+        .map(|row| row.expect("failed to get row"))
+        .collect();
+    assert_eq!(results, HashSet::from([(0,)]));
```
QuerthDP: We could do the same query, but without ANN OF and that would be the expected results, then compare with the ANN OF version. Same for other tests in this patch.

QuerthDP (dismissed):
Looks mostly good. Please consider comparing the results of filtering vs a non-ANN query so that we know it's exactly the same Scylla does.

@ewienik please review this patch as well, as I know you had some objections before about ordering those types.

nyh:
> Looks mostly good. Please consider comparing the results of filtering vs a non-ANN query so that we know it's exactly the same Scylla does.

Good idea. I hope I understand you correctly: I'll add tests that exercise the order (<, not =) and check that filtering in ANN and in base-table Scylla matches the same ordering, so we didn't botch up the ordering (the equality check is much more trivial).

> @ewienik please review this patch as well, as I know you had some objections before about ordering those types.

QuerthDP:
> > Looks mostly good. Please consider comparing the results of filtering vs a non-ANN query so that we know it's exactly the same Scylla does.
> 
> Good idea. I hope I understand you correctly: I'll add tests that exercise the order (<, not =) and check that filtering in ANN and in base-table Scylla matches the same ordering, so we didn't botch up the ordering (the equality check is much more trivial).

Correct. You can take a look how I did that for `inet` type: https://github.com/scylladb/vector-store/pull/574/changes/34d25bb81aa2b9e6299cd70db2585130a0b6da9d#diff-f5651486a11468755107f33078fbc1edbc3ab88575273d7c7955ffdf8ab97319R312-R401

> 
> > @ewienik please review this patch as well, as I know you had some objections before about ordering those types.

ewienik:
> @ewienik please review this patch as well, as I know you had some objections before about ordering those types.

My objections are connected with Rust driver. We should have comparision of these values implemented in driver, shouldn't we? What do you think @Lorak-mmk or @wprzytula ?

nyh:
> > @ewienik please review this patch as well, as I know you had some objections before about ordering those types.
> 
> My objections are connected with Rust driver. We should have comparision of these values implemented in driver, shouldn't we? What do you think @Lorak-mmk or @wprzytula ?

I think that @QuerthDP nailed it: The order that the ANN uses for its filtering must be the same order that **Scylla**, in CQL, uses for its filtering on the same data. So I'm writing now tests that will confirm that.

If the Rust driver got something wrong in this ordering considering the Scylla "golden truth", then it's a Rust driver bug we need to fix, not something we need to be compatible with. No?

wprzytula:
> > @ewienik please review this patch as well, as I know you had some objections before about ordering those types.
> 
> My objections are connected with Rust driver. We should have comparision of these values implemented in driver, shouldn't we? What do you think @Lorak-mmk or @wprzytula ?

Agreed. In case of Timeuuid, we've already added it years ago in the form of `CqlTimeuuid` dedicated wrapper: https://github.com/scylladb/scylla-rust-driver/pull/894.

As for Boolean, Uuid and Blob, we have no dedicated wrappers for them - we use plain native Rust types, which have comparison semantics either already defined differently or purposely no comparison semantics.
So we would probably have to implement wrappers for those types too (`CqlBoolean`, `CqlUuid`, `CqlBlob`).

On a second thought, more elegantly, we could introduce a new trait in the driver - `ScyllaOrd` - which would provide ordering as in the DB. WDYT @Lorak-mmk?

QuerthDP:
> > @ewienik please review this patch as well, as I know you had some objections before about ordering those types.
> 
> My objections are connected with Rust driver. We should have comparision of these values implemented in driver, shouldn't we? What do you think @Lorak-mmk or @wprzytula ?

I would agree. Same for other types (e.g. `inet`). The default Ord should be same as CQL ordering IMO

nyh:
> > > @ewienik please review this patch as well, as I know you had some objections before about ordering those types.
> > 
> > 
> > My objections are connected with Rust driver. We should have comparision of these values implemented in driver, shouldn't we? What do you think @Lorak-mmk or @wprzytula ?
> 
> I would agree. Same for other types (e.g. `inet`). The default Ord should be same as CQL ordering IMO

You're talking above my head ;-) If there's something concrete you want me to change in the approach in this patch, please tell me. Thanks.

Lorak-mmk:
> On a second thought, more elegantly, we could introduce a new trait in the driver - ScyllaOrd - which would provide ordering as in the DB. WDYT @Lorak-mmk?

In other cases we went with wrappers (CqlTimeuuid, CqlTime etc). I would like to be consistent and define new wrappers where necessary.

wprzytula:
@QuerthDP Feel free to open an issue in Rust Driver specifying your needs. Provide example usage from VS that will help us design the solution.

nyh:
> Agreed. In case of Timeuuid, we've already added it years ago in the form of `CqlTimeuuid` dedicated wrapper: [scylladb/scylla-rust-driver#894](https://github.com/scylladb/scylla-rust-driver/pull/894).

I think I'm beginning to understand how the Rust driver is related. You are saying that you are already **using** the Rust driver in the code, so you were hoping that it would give you the comparison functions and you don't need to implement them. But if it doesn't? And **must** it provide these functions - why does the driver even need to provide client-side comparison functions, when its main job is to get the server to return the right results, for which no client-side comparison functions are needed?

crates/validator/src/filtering.rs
nyh: Will do.

wprzytula:
> > Agreed. In case of Timeuuid, we've already added it years ago in the form of `CqlTimeuuid` dedicated wrapper: [scylladb/scylla-rust-driver#894](https://github.com/scylladb/scylla-rust-driver/pull/894).
> 
> I think I'm beginning to understand how the Rust driver is related. You are saying that you are already **using** the Rust driver in the code, so you were hoping that it would give you the comparison functions and you don't need to implement them. But if it doesn't? And **must** it provide these functions - why does the driver even need to provide client-side comparison functions, when its main job is to get the server to return the right results, for which no client-side comparison functions are needed?

A part of the driver's responsibility is modeling the DB types in the language it's written for. Those models should behave as close as possible to the DB types, to ensure logic portability (language-agnosticity).

crates/validator/src/filtering.rs
nyh: AI loves to brag about it's ability to find code in obscure places and also re-find it if it's "lost". I agree we shouldn't let it brag in our comments ;-)

crates/validator/src/filtering.rs
nyh: Done, but not in these tests (that use "=") but in better tests involving inequality (< or >), where this becomes significantly more interesting.

crates/vector-store/src/table/mod.rs
nyh: Agree. Removed.

crates/vector-store/src/table/mod.rs
nyh: I agree. Done.

crates/vector-store/src/table/mod.rs
nyh: Done (as a comment, I hope that's what you meant).

nyh: By the way, rustfmt really #$@!ed up that comment. So sad. So the result is uglier than I wanted it to be.

crates/vector-store/src/table/mod.rs
```diff
@@ -1315,8 +1317,27 @@ fn partition_key_from_restrictions(
...
+        return CqlTimeuuid::from_bytes(*a).cmp(&CqlTimeuuid::from_bytes(*b));
+    }
```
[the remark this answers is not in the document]
nyh: Fantastic catch!
I thought that CqlTimeuuid was specifically written by us to use the right comparison, but... wow, it turns out that Scylla does *not* use the timeuuid comparator when comparing time uuids stored in it. I'll update the validator test to catch this case.

nyh:
I pushed a new version, fixing a subtle (but serious) bug in the uuid order caught by copilot, but more importantly: Implements @QuerthDP's end-to-end ("validator") test idea: Tests inequality comparisons of different values of these types in both ANN and Scylla's base table, and validate the comparison are the same. These tests reproduce, among other things, the uuid ordering bug that copilot found. But more importantly - they are bona-fide **regression tests** - if we ever want to switch the implementation of the comparison functions - perhaps using something provided by the Rust driver as suggested by some reviewers - these tests will (or should) shout loudly if we replaced them by bad implementations. 

I also had to rebase my patch to resolve conflicts, because a lot of code moved around :-( Luckily these days the AI can do these complex rebases - something that would have taken me an hour in the past - in minutes while I'm leaning back and watching :-)

I believe we can accept this patch with its own home-grown comparison functions and **later** switch to some functions provided by the driver, and these new tests will make us feel safe to do it.

ewienik:
> I believe we can accept this patch with its own home-grown comparison functions and **later** switch to some functions provided by the driver, and these new tests will make us feel safe to do it.

IMHO it seems worth to merge this PR and in the future replace with driver version. Driver can see what we think is needed for comparison.

crates/validator/src/filtering.rs
```diff
@@ -753,6 +755,609 @@ async fn global_index_filter_by_filtering_columns(actors: Arc<TestActors>) {
...
+        .expect("failed to insert data");
+
+    let index =
+        create_index(CreateIndexQuery::new(&session, &clients, &table, "v").filter_columns(["f"]))
+            .await;
+
+    for client in &clients {
+        let index_status = wait_for_index(client, &index).await;
+        assert_eq!(index_status.count, 2, "Expected 2 vectors to be indexed");
+    }
```
ewienik, line 802: nit: this pattern can be replaced by wait_for_index_count here and following tests

crates/vector-store/src/cql_types.rs
```diff
@@ -274,8 +275,58 @@ pub(crate) fn from_json(value: Value, cql_type: &NativeType) -> anyhow::Result<C
...
+    let b = b.as_bytes();
+    let va = (a[6] >> 4) & 0x0f;
+    let vb = (b[6] >> 4) & 0x0f;
+    if va != vb {
+        return va.cmp(&vb);
+    }
+    if va == 1 {
+        // && vb == 1
+        return uuid_timeuuid_msb(a)
+            .cmp(&uuid_timeuuid_msb(b))
```
ewienik, line 307: As uuid_timeouuid_msb produce strict u64 it could be wrong in different endianness. Am I right?

ewienik (dismissed):

crates/vector-store/src/cql_types.rs
```diff
@@ -274,8 +275,58 @@ pub(crate) fn from_json(value: Value, cql_type: &NativeType) -> anyhow::Result<C
...
+fn uuid_cmp(a: &Uuid, b: &Uuid) -> Ordering {
+    let a = a.as_bytes();
+    let b = b.as_bytes();
+    let va = (a[6] >> 4) & 0x0f;
+    let vb = (b[6] >> 4) & 0x0f;
+    if va != vb {
+        return va.cmp(&vb);
+    }
+    if va == 1 {
+        // && vb == 1
```
ewienik, line 305: nit: leftover?

nyh: Ironically, it's the opposite of leftover:

The code has if va = 1, and because we didn't return above, we know va == vb, so vb is also 1. The original code just had `if va == 1` but the AI reviewer asked me to aks add vb == 1 as a hint to readers what this case really means.
The @$#$ing rustfmt simply refused to allow me to put this comment on the same line as the if (I tried `/* .. */` as well as '//' comments) so it ended up on the next line. Ugly.

Maybe I should have just refused the reviewer's request...

crates/validator/src/filtering.rs
nyh: I'll do it, even though in general I favor the pytest approach where asserts should look like `assert something()`, not a bazillion utility functions `assert_that_something_happened()`

Note that this change will affect more than the tests that I added. So I'll do it in a second patch in this PR.

crates/vector-store/src/cql_types.rs
nyh: I'll let Claude answer:

No — the reviewer's concern doesn't apply here, and the function is correct on any host endianness.

The key is how the u64 is built at [cql_types.rs:316-325](vscode-webview://0oh2h8ra67pih7u5iq9reisi831235hm420nnruj9qu209kb8cib/crates/vector-store/src/cql_types.rs#L316-L325):

```
fn uuid_timeuuid_msb(bytes: &[u8; 16]) -> u64 {
    u64::from(bytes[6] & 0x0f) << 56
        | u64::from(bytes[7]) << 48
        | u64::from(bytes[4]) << 40
        | u64::from(bytes[5]) << 32
        | u64::from(bytes[0]) << 24
        | u64::from(bytes[1]) << 16
        | u64::from(bytes[2]) << 8
        | u64::from(bytes[3])
```

This reads each byte individually as a u8, widens it to u64, and shifts it into an explicit bit position by hand. There's no reinterpretation of a multi-byte memory region as an integer — no from_ne_bytes/from_le_bytes, no pointer cast, no transmute. Since each byte's position in the resulting integer's value (not its in-memory representation) is chosen explicitly by the shift amount, the resulting u64 numeric value is identical regardless of whether the host CPU is little-endian or big-endian. Endianness only matters when you reinterpret a byte buffer as a native integer type (e.g. u64::from_ne_bytes(...)) — that's not happening here.

And then .cmp() on two u64s compares by numeric value, which is likewise endianness-independent — Rust integer comparison isn't defined in terms of byte layout.

So the reviewer's specific worry ("produces a strict u64 that could be wrong in different endianness") doesn't hold: this would only be a real risk if the code used a native-endian byte-reinterpretation shortcut instead of the explicit per-byte shift/OR composition it actually uses. You can reply to the thread explaining this — worth pointing out the `u64::from(bytes[N]) << shift` pattern is the idiomatic Rust way to do endianness-agnostic manual byte assembly, matching how ScyllaDB's own timeuuid_read_msb in UUID.hh does the same explicit reassembly (also required there, since UUID byte layout is RFC 4122 big-endian while x86 is little-endian).

crates/validator/src/filtering.rs
```diff
@@ -753,6 +755,609 @@ async fn global_index_filter_by_filtering_columns(actors: Arc<TestActors>) {
...
+        None,
+    )
+    .await;
+
+    let u_min = Uuid::parse_str("00000000-0000-1000-8000-000000000000").unwrap(); // v1, timestamp = 0
+    let u_a = Uuid::parse_str("ffffffff-0000-1000-8000-000000000001").unwrap(); // v1, timestamp ~ 4.3e9
+    let u_b = Uuid::parse_str("00000000-0000-1fff-8000-000000000002").unwrap(); // v1, timestamp ~ 1.15e18
+    let u_max = Uuid::parse_str("ffffffff-ffff-1fff-bfff-ffffffffffff").unwrap(); // v1, max timestamp
+    let u_v4 = Uuid::parse_str("80000000-0000-4000-8000-000000000000").unwrap(); // v4
+    let u_lo = Uuid::parse_str("00000000-0000-1000-7f00-000000000003").unwrap(); // v1, timestamp = 0, same as u_min
```
QuerthDP: How is that possible that both `u_min` and `u_lo` share the same representation in the same version?

crates/vector-store/src/cql_types.rs
```diff
@@ -386,12 +446,7 @@ mod tests {
         }
     }
 
-    pub(crate) const CMP_UNSUPPORTED: &[NativeType] = &[
-        NativeType::Blob,
-        NativeType::Boolean,
-        NativeType::Timeuuid,
-        NativeType::Uuid,
-    ];
+    pub(crate) const CMP_UNSUPPORTED: &[NativeType] = &[];
```
QuerthDP: Why do we keep it? I guess that in case when we would support some type as PK, but not implement CMP, but why would we defer implementing it?

QuerthDP (dismissed):
Please address the remaining comments or resolve them, so that we can merge.

crates/validator/src/filtering.rs
nyh: Done in a second patch (I'll push in a minute) because it modifies several tests, not just the new test, to use the same pattern.

crates/validator/src/filtering.rs
nyh: The comment is indeed bad. u_lo and u_min are different raw UUIDs but their "v1 timestamp" is the same.

Where they differ is in parts outsie the timestamp - and we are trying to test the tie-breaking" in this case. When the timestamps tie, ScyllaDB's uuid_cmp() falls back to comparing bytes 8..16 as plain unsigned bytes, so 0x7f < 0x80 puts u_lo before u_min. CqlTimeuuid's Ord impl uses a different, sign-flipped tie-break on that same byte range, which would order them the other way - which is a bug that copilot caught in my original implementation.

The longer doc comment above already explains this, but I agree the short comment is bad. I'll fix it.

crates/vector-store/src/cql_types.rs
nyh: Good question. Instead of deleting this and the test it drives, I thought maybe in the future we'll have more types that needs to be on this list again. But... Why... Why would we repeat this mistake of half-implementing a type - supporting it in PK but not supporting CMP in the same patch? That would just be a mistake, not something we should encourage future developers to do. I'll fix it. Thanks.

nyh:
> Please address the remaining comments or resolve them, so that we can merge.

Addressed all the review comments and pushed a new version.

crates/validator/src/filtering.rs
QuerthDP, line 1: I think the commits should be in inverse order. Why introduce tests with `wait_for_index` just to change it to `wait_for_index_count` in the next commit.

nyh: I can do that, though I think it really doesn't matter. The end result is exactly the same...

QuerthDP (approved):

crates/validator/src/filtering.rs
QuerthDP: Let it be

crates/vector-store/src/cql_types.rs
QuerthDP: I requested for that. I think it looks better with it.

PR #585, opened 2026-09-03
ci: trigger siren add-vectorstore-ver workflow after cloud images are built

swasik:
### Motivation

The last step of a Vector Store release is still manual: after the `build-cloud-images` job uploads `packer-manifest.json`, someone has to open siren's [add-vectorstore-ver](https://github.com/scylladb/siren/actions/workflows/add-vectorstore-ver.yml) workflow, copy the AMI IDs and the GCP image name out of the manifest, and start it via `workflow_dispatch`. The siren workflow already supports being triggered remotely (`repository_dispatch` with `types: [vectorstore-release]`), and the `SIREN_REPO_TOKEN` secret is now provisioned (VECTOR-854), so the release workflow can send that dispatch itself.

### What this does

**`packer: tag manifest entries with the target architecture`** — both AWS AMIs come from the same `amazon-ebs` builder block (architecture is only a user variable), so their manifest entries were indistinguishable except by build ordering. The manifest post-processor now records `custom_data.arch`, letting consumers select AMIs by architecture instead of position.

**`ci: trigger siren add-vectorstore-ver workflow after cloud images`** — adds a `trigger-siren` job with `needs: [build-cloud-images]` that:

- downloads the packer manifest artifact and extracts the bare `us-east-1` amd64/arm64 AMI IDs and the GCP image name, selecting AWS entries by `custom_data.arch` and taking the **last** matching entry (re-running a build appends to `.builds[]` rather than replacing);
- sends the `vectorstore-release` `repository_dispatch` to scylladb/siren via `peter-evans/repository-dispatch` (pinned by SHA) using `SIREN_REPO_TOKEN`;
- **skips** (not fails) when `VECTOR_VERSION` is not a GA `X.Y.Z` version — siren validates against `^X.Y.Z$` and unconditionally moves `defaults.vectorstore`, and while the `release: [released]` trigger already filters prereleases, `workflow_dispatch` takes an arbitrary version;
- fails loudly if any image ID cannot be extracted or the dispatch itself fails, and logs the dispatched IDs plus a link to the siren workflow page in the job summary (`repository_dispatch` returns 204 with no run id, so the siren run status cannot be reported directly).

Being a separate job consuming the artifact (the `push-docker` pattern), the dispatch can be re-run without rebuilding the AMIs.

`docs/releasing.md` documents the automated step, the GA-only guard, the out-of-order-release caveat (a patch release on an older series still moves siren's `defaults.vectorstore` back — review the siren PR before merging in that case), and the manual `workflow_dispatch` fallback.

### Test plan

- [x] Workflow YAML and packer template JSON parse cleanly
- [x] jq extraction verified against a simulated manifest: re-run append picks the last entry, AMIs selected by arch (not position), bare `us-east-1` AMI extracted from the `region:ami` list, missing arch fails the job with a clear error
- [ ] Dry-run: dispatch a `workflow_dispatch` release run with a non-GA version (e.g. `X.Y.Z-dev`) and confirm `trigger-siren` skips cleanly
- [ ] First GA release: confirm the `feat(vectorstore): add version X.Y.Z [automation]` PR appears in scylladb/siren with the correct image IDs

Fixes: VECTOR-853
Refs: VECTOR-854

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -195,6 +195,107 @@ jobs:
           asset_name: vector-store-cloud-images-${{ env.VECTOR_VERSION }}.json
           asset_content_type: application/json
 
+  # Hand the freshly built cloud images over to siren: extract the image IDs
+  # from the packer manifest and send the repository_dispatch that starts
+  # https://github.com/scylladb/siren/actions/workflows/add-vectorstore-ver.yml,
+  # which opens a versions.yaml PR in scylladb/siren. A separate job consuming
+  # the manifest artifact (the push-docker pattern) keeps the dispatch
+  # re-runnable without rebuilding the images.
+  trigger-siren:
+    needs: [build-cloud-images]
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - name: Check the version is a GA release
+        id: ga
+        run: |
+          set -euo pipefail
+          # siren validates the version against ^X.Y.Z$ and unconditionally
+          # moves its defaults.vectorstore to it, so pre-release and dev tags
+          # (1.6.1-rc0, X.Y.Z-dev) must not be dispatched. The release trigger
+          # already filters prereleases out, but workflow_dispatch takes an
+          # arbitrary VECTOR_VERSION — skip (not fail) the job on those.
+          if [[ "${VECTOR_VERSION}" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]] ; then
+            echo "ga=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "ga=false" >> "$GITHUB_OUTPUT"
+            echo "Version \`${VECTOR_VERSION}\` is not a GA X.Y.Z release — skipping the siren dispatch" >> "$GITHUB_STEP_SUMMARY"
+          fi
+
+      - name: Download the packer manifest artifact
+        if: steps.ga.outputs.ga == 'true'
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          name: vector-store-cloud-images-${{ env.VECTOR_VERSION }}
+
+      - name: Extract the image IDs from the packer manifest
+        if: steps.ga.outputs.ga == 'true'
+        id: images
+        run: |
+          set -euo pipefail
+          # All three sequential packer builds append to the one .builds[]
+          # array, and a re-run appends rather than replaces — take the last
+          # matching entry. The AWS builds are told apart by the arch recorded
+          # in custom_data by the manifest post-processor; their artifact_id
+          # is a comma-separated region:ami list and siren expects the bare
+          # us-east-1 AMI (it prefixes the value with us-east-1/ itself).
+          # The GCP image needs no such disambiguation: only one architecture
+          # (amd64) is built there, so builder_type alone identifies it.
+          aws_ami() {
+            jq -r --arg arch "$1" \
+              '[.builds[] | select(.builder_type=="amazon-ebs" and .custom_data.arch==$arch)] | last | .artifact_id // empty' \
+              packer-manifest.json \
+              | tr ',' '\n' | { grep '^us-east-1:' || true ; } | cut -d: -f2
+          }
+          aws_amd64_ami=$(aws_ami amd64)
+          aws_arm64_ami=$(aws_ami arm64)
+          gcp_amd64_image=$(jq -r \
+            '[.builds[] | select(.builder_type=="googlecompute")] | last | .artifact_id // empty' \
+            packer-manifest.json)
+          for var in aws_amd64_ami aws_arm64_ami gcp_amd64_image ; do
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yaml
```diff
@@ -195,6 +195,105 @@ jobs:
...
+        if: steps.ga.outputs.ga == 'true'
+        id: images
+        run: |
+          set -euo pipefail
+          # All three sequential packer builds append to the one .builds[]
+          # array, and a re-run appends rather than replaces — take the last
+          # matching entry. The AWS builds are told apart by the arch recorded
+          # in custom_data by the manifest post-processor; their artifact_id
+          # is a comma-separated region:ami list and siren expects the bare
+          # us-east-1 AMI (it prefixes the value with us-east-1/ itself).
```
swasik, line 242: Add a comment that this issue does not exist for GCP as we support just one architecture there.

swasik: Done in d533ff1 (amended into the ci commit) — the extraction-step comment now notes the GCP image needs no arch disambiguation, since only amd64 is built there and builder_type alone identifies it.

swasik:
@coderabbitai The docs/releasing.md finding about the manual fallback is addressed in d533ff1: the release asset only exists for release-event runs, so the docs now direct workflow_dispatch runs to download the vector-store-cloud-images-X.Y.Z workflow artifact from the run page instead.

docs/releasing.md
```diff
@@ -87,6 +87,45 @@ This will upload all tar.gz archives and SBOM files to the GitHub release
...
+a GA release.
+
+Two caveats:
+
+- Only GA `X.Y.Z` versions are dispatched. Pre-release and dev tags
+  (`1.6.1-rc0`, `X.Y.Z-dev`) skip the dispatch, because siren validates the
+  version against `^X.Y.Z$` and unconditionally moves its
+  `defaults.vectorstore` to the dispatched version. For the same reason,
+  releasing a patch on an *older* series still moves `defaults.vectorstore`
+  back to it — review the siren PR before merging in that case.
```
ewienik, line 109: I think this behavior is bad as not always we need to move default vector-store version. Could it be split into two workflows - one for automatic registering new vector-store version while releasing, second for manually triggering workflow for setting new default vector-store version?

ewienik (changes requested):

docs/releasing.md
swasik: I want to reduce the complexity and number of manual steps here. I would like to follow the services oriented design here - each production version of service should be automatically released. And the QA pipeline should guarantee that the risk is low. If we don't want to release, we should create -dev or -rc version.

ewienik: Will we have always HEAD version without branching? We should have some way (labeling or anything) to say that we release this version as GA but it is not default, shouldn't we?

ewienik: What about provide such information in git tag annotation? If this version is not for default that it should be pointed in tag commit description.

swasik: No, we do branch — but only one series is alive at a time, and that is what makes the unconditional default move correct rather than accidental.

The policy is: after X.Y.0 ships we branch, and any fixes on it go out as X.Y.1, X.Y.2, … Each of those is the new production version, so it should become the default. Once X.(Y+1).0 ships, it goes to production and we stop developing X.Y.Z. So there is never a GA tag that is not the newest GA version, and no per-release "is this the default?" decision to express. If we need to go back to an older version, that is a rollback, not a release — it gets configured by hand in siren.

That is also why I would rather not put it in the tag annotation: there is nothing left for the annotation to carry, and it would mean CI parsing prose to decide release behaviour, immutably, invisibly from the release page.

You were right that the docs invited exactly this question, though. The caveat you commented on described the default move as a hazard to watch out for, which is backwards — it is the intended outcome. Rewritten in e402fb9 (amended into the ci commit): the section now states the one-version-in-production policy outright and says a rollback is a manual siren edit rather than a release.

One real gap left, which is ordering rather than intent: a patch still in flight when the next minor ships (1.9.2 tagged a day after 1.10.0) would dispatch an older version and propose a default downgrade. That is a process mistake rather than a supported case, and siren's automation opens a PR instead of merging one, so it is catchable at review — but the guard belongs in siren, next to the file that holds the default, so that it covers manual workflow_dispatch runs too. Filed as CLOUD-3900: register the version but leave `defaults.vectorstore` alone when the dispatched version is older than the current default.

VAveryanov8 (approved):
LGTM! Let's test it! 

> docs/releasing.md documents the automated step, the GA-only guard, the out-of-order-release caveat (a patch release on an older series still moves siren's defaults.vectorstore back — review the siren PR before merging in that case), and the manual workflow_dispatch fallback.

Yes, it's fine - I will still manually review the PR. Additionally, if we see that ordering issues are common, then we can adjust the pipeline on the siren side, so it will have more validations and etc.

VAveryanov8 (approved):

docs/releasing.md
ewienik: ok. let's check how it works

ewienik (approved):

PR #590, opened 2026-09-07
httpapi: report the index status as part of the index info

QuerthDP:
Rendering the Vector Store index build status of a whole cluster took M x N requests. Listing the indexes of a node needs an `/indexes` call, then every index needs an `/indexes/{keyspace}/{index}/status` call. Such polling grows with the size of the cluster.

The status now belongs to the description of an index: `IndexInfo` carries `status`, `count` and `build_progress` next to the creation options, and every endpoint serving an index answers with it. `/indexes` lists one such object per index, so a node takes one request:

```json
{
  "keyspace": "my_keyspace",
  "index": "my_vector_index",
  "options": { "type": "vector", "dimensions": 384, "...": "..." },
  "status": "SERVING",
  "count": 12345,
  "build_progress": 100.0
}
```

`/indexes/{keyspace}/{index}` answers with the same object for one index, and so does `/indexes/{keyspace}/{index}/status`, which leaves `IndexStatusResponse` with nothing to describe. That path is marked deprecated in the OpenAPI document and stays served for the clients reading the status from it; its body only gains fields, so those clients keep working.

The counts are gathered concurrently, once the `Indexes` read lock has been released. The listing skips an index whose count cannot be obtained, for example while the index is being removed. A lookup of a single index counts the items as well, and answers 500 when the count fails.

Worth a reviewer's attention: `/indexes` and `/indexes/{keyspace}/{index}` used to be pure in-memory reads, and each of them now issues one count per index it reports. On the OpenSearch provider that count is a remote `_count` request.

The API version goes to 3.1.0 and `openapi.json` is regenerated. Integration tests cover a serving index, one whose initial scan still runs, and a full-text index.

Fixes: VECTOR-887

### DIFF
--- a/api/openapi.json
+++ b/api/openapi.json
@@ -427,7 +427,7 @@
         "tags": [
           "scylla-vector-store-index"
         ],
-        "description": "Returns the same information about a single index as `/api/v1/indexes/{keyspace}/{index}` returns.",
+        "description": "Returns the same information about a single index as `/api/v1/indexes/{keyspace}/{index}`. Deprecated. Use that endpoint instead.",
         "operationId": "get_index_status",
         "parameters": [
           {
@@ -496,7 +496,8 @@
               }
             }
           }
-        }
+        },
+        "deprecated": true
       }
     },
     "/api/v1/info": {
--- a/api/openapi.json
+++ b/api/openapi.json
@@ -6,7 +6,7 @@
     "license": {
       "name": "LicenseRef-ScyllaDB-Source-Available-1.1"
     },
-    "version": "3.0.0"
+    "version": "3.1.0"
   },
   "paths": {
     "/api/v1/indexes": {
@@ -14,7 +14,7 @@
         "tags": [
           "scylla-vector-store-index"
         ],
-        "description": "Returns the list of indexes managed by the Vector Store indexing service. The list includes both vector and fulltext indexes in any state (initializing, available/built, destroying). Due to synchronization delays, it may temporarily differ from the list of indexes inside ScyllaDB.",
+        "description": "Returns the list of indexes managed by the Vector Store indexing service, each with its current operational status, the total number of items indexed and the progress of the initial table scan that backfills it. The list includes both vector and fulltext indexes in any state (initializing, available/built, destroying). Due to synchronization delays, it may temporarily differ from the list of indexes inside ScyllaDB. An index whose item count cannot be obtained, e.g. because it is being removed while the request is served, is omitted from the list.",
         "operationId": "get_indexes",
         "responses": {
           "200": {
@@ -29,6 +29,8 @@
                 },
                 "example": [
                   {
+                    "build_progress": 100.0,
+                    "count": 12345,
                     "index": "my_vector_index",
                     "keyspace": "my_keyspace",
                     "options": {
@@ -39,16 +41,20 @@
                       "search_beam_width": 64,
                       "similarity_function": "COSINE",
                       "type": "vector"
-                    }
+                    },
+                    "status": "SERVING"
                   },
                   {
+                    "build_progress": 37.5,
+                    "count": 42,
                     "index": "my_fulltext_index",
                     "keyspace": "my_keyspace",
                     "options": {
                       "analyzer": "standard",
                       "positions": true,
                       "type": "fulltext"
-                    }
+                    },
+                    "status": "BOOTSTRAPPING"
                   }
                 ]
               }
@@ -62,7 +68,7 @@
         "tags": [
           "scylla-vector-store-index"
         ],
-        "description": "Retrieves information about a specific index, including its search type and the options it was created with. This is the same information reported per-entry by the `/api/v1/indexes` listing, scoped to a single index.",
+        "description": "Retrieves information about a specific index: its search type, the options it was created with, its current operational status, the total number of items indexed and the progress of the initial table scan that backfills it. Each entry of the `/api/v1/indexes` listing carries the same information.",
         "operationId": "get_index_info",
         "parameters": [
           {
@@ -86,13 +92,15 @@
         ],
         "responses": {
           "200": {
-            "description": "Successful operation. Returns the index's type and creation options.",
+            "description": "Successful operation. Returns the index's type, creation options, and the status it currently reports.",
             "content": {
               "application/json": {
                 "schema": {
                   "$ref": "#/components/schemas/IndexInfo"
                 },
                 "example": {
+                  "build_progress": 100.0,
+                  "count": 12345,
                   "index": "my_vector_index",
                   "keyspace": "my_keyspace",
                   "options": {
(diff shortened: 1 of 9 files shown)


crates/vector-store/src/httproutes.rs
```diff
@@ -304,29 +310,68 @@ impl From<crate::SimilarityScore> for httpapi::SimilarityScore {
...
-        })
-        .chain(indexes_guard.iter_fts().map(|(key, entry)| IndexInfo {
-            keyspace: key.keyspace().into(),
-            index: key.index().into(),
-            options: IndexOptions::Fulltext(entry.options().into()),
-        }))
-        .collect();
+    let indexes: Vec<_> = future::join_all(entries.into_iter().map(
+        |(key, options, index, status, progress)| async move {
+            match index.count(key.clone()).await {
```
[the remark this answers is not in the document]
QuerthDP: Not fixed. Correct observation, but the swallowing lives in `opensearch.rs` and predates this PR, and preserving those errors would also change `/indexes/{keyspace}/{index}/status` from `count: 0` to a 500. That is a separate change with its own blast radius, so it gets its own ticket. The documented skip stays reachable: usearch returns `Err` when the index id is missing from the table, and a closed actor channel errors too.

_Reply generated by Claude Code._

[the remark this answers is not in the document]
QuerthDP: Not fixed. One request fans out to exactly the number of indexes on the node, which the schema sets rather than the caller, and a per-request cap would not bound the aggregate anyway (N concurrent callers still give N x M in-flight counts). On the deadline: no handler in this crate applies a request-path timeout, and `/indexes/{keyspace}/{index}/status` already awaits a count unbounded, so this PR does not introduce that exposure. The real gap is the OpenSearch client being built without a request timeout, which would bound both endpoints; that belongs with the item above.

_Reply generated by Claude Code._

crates/vector-store/src/httproutes.rs
```diff
@@ -304,29 +310,68 @@ impl From<crate::SimilarityScore> for httpapi::SimilarityScore {
...
+                    entry.progress(),
+                )
+            })
+            .chain(indexes.iter_fts().map(|(key, entry)| {
+                (
+                    key.clone(),
+                    IndexOptions::Fulltext(entry.options().into()),
+                    IndexSender::Fts(entry.index().clone()),
+                    entry.status(),
+                    entry.progress(),
```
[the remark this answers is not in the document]
QuerthDP: Fixed. Added `indexes_listing_reports_status_of_a_fulltext_index`, which indexes one document and asserts the listed entry reports `SERVING`, `count == 1` through the `IndexSender::Fts` branch, `build_progress == 100.0` and `IndexOptions::Fulltext`.

_Reply generated by Claude Code._

ewienik:
> /indexes now reports the status, item count and build progress of every index it lists, so a node takes one request. Each entry is the description it already carried, plus the body of the per-index status endpoint

> /indexes used to be a pure in-memory read, and now issues one count per listed index. /indexes/{keyspace}/{index} returns a plain IndexInfo as before, so a single-index lookup stays free of counting, and /indexes/{keyspace}/{index}/status is unchanged.

To be aligned with standard JSON API I would rather combine IndexStatus withing IndexInfo and remove `/indexes/{keyspace}/{index}/status` endpoint. The two endpoint would be aligned: `/indexes` can return a list of IndexInfo and `/indexes/{keyspace}/{index}` can return single IndexInfo. If there are some backward compatibility issues we can stay for a while with `/indexes/{keyspace}/{index}/status` returning IndexInfo also.

QuerthDP:
Changelog:
- deprecate `/{index}/status` endpoint
- adjust `IndexInfo` to contain `IndexStatusResponse`
- make `/{index}` return `IndexInfo` (with status) matching `/indexes` endpoint

QuerthDP:
Changelog:
- split `/indexes` and `/{index}` changes from `/status` changes

QuerthDP:
Changelog:
- rebased on master

QuerthDP:
Changelog:
- remove the invalid compatibility according `build_progress`

crates/httpapi/src/lib.rs
```diff
@@ -82,11 +82,24 @@ impl Serialize for Distance {
...
+    pub count: usize,
+    /// Progress of the initial full table scan that backfills the index,
+    /// expressed as a percentage in the range `0.0..=100.0`. This reflects
+    /// only how far the full scan has advanced; it is not a source of truth
+    /// for the index state. The `status` field is authoritative: an index
+    /// reporting `SERVING` is serving regardless of `build_progress`, which
+    /// may not reach `100.0` (e.g. if some scan ranges failed). Defaults to
+    /// `100.0` when omitted, so a new client talking to an older server that
+    /// does not report the field falls back to relying solely on `status`.
+    #[serde(default = "default_build_progress")]
```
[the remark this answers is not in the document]
QuerthDP: Fixed. `status` and `count` have no default, so an older response fails to parse before `build_progress`'s default is ever reached — the sentence and the `#[serde(default)]` are gone, and the schema now marks `build_progress` required, which is what the server always sends. `default_build_progress` is dropped together with `IndexStatusResponse`, its last user, in the next commit.

_Reply generated by Claude Code._

api/openapi.json
```diff
@@ -427,7 +427,7 @@
         "tags": [
           "scylla-vector-store-index"
         ],
-        "description": "Returns the same information about a single index as `/api/v1/indexes/{keyspace}/{index}` returns.",
+        "description": "Returns the same information about a single index as `/api/v1/indexes/{keyspace}/{index}`. It is kept for the clients that read the status of an index from this path; use the other endpoint instead.",
```
knowack1: Could be simply said: "Deprecated. Use /api/v1/indexes/{keyspace}/{index} instead."

crates/vector-store/src/httproutes.rs
```diff
@@ -446,6 +448,7 @@ impl From<crate::node_state::IndexStatus> for httpapi::IndexStatus {
         )
     )
 )]
+#[deprecated(note = "use get_index_info, serving /api/v1/indexes/{keyspace}/{index}")]
```
knowack1: "serving" confuses with index "serving" state.. Consider just "use get_index_info at /api/v1/indexes/{keyspace}/{index}" or some small rewording.

crates/vector-store/tests/integration/index_status.rs
knowack1, line 1: We test client.indexes() in index_info.rs, however this file is also fine for me.

knowack1 (dismissed):

crates/vector-store/tests/integration/index_status.rs
```diff
@@ -115,3 +119,143 @@ async fn index_status_reports_full_progress_when_serving() {
...
+                .indexes()
+                .await
+                .first()
+                .is_some_and(|entry| entry.status == IndexStatus::Serving && entry.count == 1)
+        },
+        "the listed index to be serving with a single item indexed",
+    )
+    .await;
+
+    let entries = client.indexes().await;
```
knowack1: Consider wait_for_value(). Also in other places.

QuerthDP:
Changelog:
- reword the deprecation endpoint description
- moved tests to `index_info.rs`
- added `wait_for_value()` instead of `wait_for()` while waiting on status

crates/vector-store/tests/integration/index_status.rs
QuerthDP: Moved to `index_info.rs` as it fits better here.

QuerthDP: The whole `index_status.rs` could possibly be removed when deprecating the `/status` endpoint.

crates/vector-store/src/httproutes.rs
QuerthDP: Reworded

api/openapi.json
QuerthDP: Changed

crates/vector-store/tests/integration/index_status.rs
QuerthDP: Changed

ewienik (approved):

PR #592, opened 2026-09-08
vector-store: reduce allocations on fetched Vector rows

Akvear:
Reading an embedding through CqlValue materializes one 72-byte enum per
4-byte float, then walks it again to build the Vec<f32>. This PR removes this for both 
fetching vectors during the DiskANN graph traversal and during full-scan and CDC.
Other CQL types are still decoded the same way as before, including Alternator embeddings
which arrive as blobs.

Fixes: VECTOR-934

### DIFF
--- a/crates/vector-store/src/db_value.rs
+++ b/crates/vector-store/src/db_value.rs
@@ -1,0 +1,246 @@
+/*
+ * Copyright 2026-present ScyllaDB
+ * SPDX-License-Identifier: LicenseRef-ScyllaDB-Source-Available-1.1
+ */
+
+//! Row and value types for reads of an indexed base table.
+//!
+//! The rows read by a full scan ([`crate::db_index`]) and by a CDC upsert's
+//! point read ([`crate::db_cdc`]) have a column layout that is only known at
+//! runtime, which is why they used to be read as [`CqlValue`]s. That is
+//! needlessly expensive for the embedding column: a `vector<float, N>`
+//! becomes a `Vec<CqlValue>`, one 72-byte enum per 4-byte element, which is
+//! then walked a second time to unwrap the floats.
+//!
+//! [`DbRow`] keeps the dynamic layout, but dispatches per value on the
+//! column's CQL type, so the embedding is deserialized straight into a
+//! [`Vector`] while every other value keeps its [`CqlValue`]
+//! representation.
+
+use crate::Vector;
+use scylla::cluster::metadata::ColumnType;
+use scylla::cluster::metadata::NativeType;
+use scylla::deserialize::DeserializationError;
+use scylla::deserialize::FrameSlice;
+use scylla::deserialize::TypeCheckError;
+use scylla::deserialize::row::ColumnIterator;
+use scylla::deserialize::row::DeserializeRow;
+use scylla::deserialize::value::DeserializeValue;
+use scylla::frame::response::result::ColumnSpec;
+use scylla::value::CqlValue;
+
+/// A single value of a [`DbRow`].
+#[derive(Clone, Debug, PartialEq)]
+pub(crate) enum DbValue {
+    /// A value of a native CQL `vector<float, N>` column.
+    Vector(Vector),
+    /// Any other value, in the dynamic representation its runtime-only CQL
+    /// type requires. For an Alternator table this includes the embedding
+    /// column, which arrives as a tagged `CqlValue::Blob`.
+    Value(CqlValue),
+}
+
+impl<'frame, 'metadata> DeserializeValue<'frame, 'metadata> for DbValue {
+    fn type_check(typ: &ColumnType) -> Result<(), TypeCheckError> {
+        match typ {
+            // Only a vector<float, N> is an embedding. A vector of any other
+            // element type is an ordinary column and keeps its CqlValue
+            // representation, instead of failing the type check of the whole
+            // row.
+            ColumnType::Vector { typ: element, .. }
+                if matches!(**element, ColumnType::Native(NativeType::Float)) =>
+            {
+                <Vector>::type_check(typ)
+            }
+            _ => <CqlValue>::type_check(typ),
+        }
+    }
+
+    fn deserialize(
+        typ: &'metadata ColumnType<'metadata>,
+        v: Option<FrameSlice<'frame>>,
+    ) -> Result<Self, DeserializationError> {
+        match typ {
+            ColumnType::Vector { typ: element, .. }
+                if matches!(**element, ColumnType::Native(NativeType::Float)) =>
+            {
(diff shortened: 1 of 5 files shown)


Akvear:
@coderabbitai review

ewienik:
@Lorak-mmk @wprzytula We are preparing here a custom serialization for vector of floats. Do you think it make sense to update driver with something similar to `CqlValue::VectorFloat`?

Lorak-mmk:
> @Lorak-mmk @wprzytula We are preparing here a custom serialization for vector of floats. Do you think it make sense to update driver with something similar to CqlValue::VectorFloat?

Not really.

CqlValue / Row are dynamic types, that should be used only when you don't know the type. 
If you are deserializing / serializing "something" and you have no idea what that "something" (because it can be anything) then CqlValue / Row are right choices.

If you are serializing / deserializing a vector of float specifically, then you know what it is, so the above doesn't apply, so you should use a proper type (`Vec<f32>` in this case). If you want your own newtype (`Vector`) then you need to impl DeserializeValue / SerializeValue for it that delegate to the underlying Vec, which this PR correctly does.

Adding `CqlValue::VectorFloat` also doesn't work because of backwards compatibility issues.

crates/vector-store/src/vector.rs
```diff
@@ -60,6 +66,42 @@ impl TryFrom<CqlValue> for Vector {
...
+struct EmbeddingColumnError(String);
+
+impl<'frame, 'metadata> DeserializeValue<'frame, 'metadata> for Vector {
+    fn type_check(typ: &ColumnType) -> Result<(), TypeCheckError> {
+        match typ {
+            ColumnType::Vector { .. } => <Vec<f32>>::type_check(typ),
+            ColumnType::Native(NativeType::Blob) => Ok(()),
+            other => Err(TypeCheckError::new(EmbeddingColumnError(format!(
+                "unsupported CQL type for an embedding column: {other:?}"
+            )))),
```
wprzytula: 💭 Is stringification really the best way to represent inner error kinds?

crates/vector-store/src/indexed_row.rs
```diff
@@ -0,0 +1,235 @@
...
+/// A single column of an [`IndexedRow`].
+#[derive(Clone, Debug, PartialEq)]
+pub(crate) enum IndexedColumn {
+    /// A native CQL `vector<float, N>` column.
+    Vector(Vector),
+    /// Any other column, in the dynamic representation its runtime-only CQL
+    /// type requires. For an Alternator table this includes the embedding
+    /// column, which arrives as a tagged `CqlValue::Blob`.
+    Value(CqlValue),
+}
```
wprzytula: 🤔 How does this reduce allocations?

The enum itself has even bigger size than `CqlValue`, because it adds the discriminant's size to its biggest variant (`CqlValue`).

Lorak-mmk: This enum has bigger size, but the memory usage will be lower.
In CqlValue, each element of the Vector is itself a CqlValue. Here, you have just `Vec<f32>`

crates/vector-store/src/vector.rs
ewienik: IMHO you can consider refactoring by using enum subtype as error.

crates/vector-store/src/db_value.rs
ewienik, line 1: Consider refactoring this PR in a way to in the first commit introduce db_value.rs (instead of indexed_row.s) and put there implementation of DbRow (instead of IndexedRow), DbValue (instead of IndexedValue) and `impl DeserializeValue` for Vector. IMHO DbRow and DbValue are better names that point at the scope of values.
In the second commit refactor other modules to use implemented types. In this way it will be easier to review and distinguish new types from logic changes.

ewienik (changes requested):

Akvear:
Changes:
- changed order of commits
- rewrote commit messages
- changed String error to enum

crates/vector-store/src/db_value.rs
```diff
@@ -0,0 +1,227 @@
...
+    /// Any other value, in the dynamic representation its runtime-only CQL
+    /// type requires. For an Alternator table this includes the embedding
+    /// column, which arrives as a tagged `CqlValue::Blob`.
+    Value(CqlValue),
+}
+
+impl<'frame, 'metadata> DeserializeValue<'frame, 'metadata> for DbValue {
+    fn type_check(typ: &ColumnType) -> Result<(), TypeCheckError> {
+        match typ {
+            ColumnType::Vector { .. } => <Vector>::type_check(typ),
```
ewienik: consider checking if this is vector of floats

crates/vector-store/src/db_value.rs
```diff
@@ -0,0 +1,227 @@
...
+            _ => <CqlValue>::type_check(typ),
+        }
+    }
+
+    fn deserialize(
+        typ: &'metadata ColumnType<'metadata>,
+        v: Option<FrameSlice<'frame>>,
+    ) -> Result<Self, DeserializationError> {
+        match typ {
+            ColumnType::Vector { .. } => <Vector>::deserialize(typ, v).map(Self::Vector),
```
ewienik: consider checking if this is vector of floats

ewienik (changes requested):

crates/vector-store/src/db_value.rs
Akvear: I believe this already checks this, there is a test to confirm

```rust
 #[test]
    fn deserialize_rejects_a_wrong_vector_element_type() {
        let typ = ColumnType::Vector {
            typ: Box::new(ColumnType::Native(NativeType::Int)),
            dimensions: 1,
        };
        assert!(<Vector>::type_check(&typ).is_err());
    }
```

crates/vector-store/src/db_value.rs
Akvear: The type check should rule this out.

crates/vector-store/src/db_value.rs
ewienik: I mean that for correctness of parsing we need to parse only Vec<Float> into our Vector. Other vector types should be parsed into CqlValue::Vector.

crates/vector-store/src/db_value.rs
ewienik: If you want to deserilize only Vector<Float> into our Vector, other vectors should deserialize into CqlValue::Vector.

Akvear:
Changes:
- only deserialize `Vec<f32>` into Vector, other `Vec<T>` are deserialized into CqlValue::Vector
- changed test to validate this happens

ewienik (approved):

PR #595, opened 2026-09-08
vector-store: fix concurrent reading schema version and index list

ewienik:
Currently, monitor_indexes reads schema version and index list in concurrent approach. Reading schema version is done using state_id from system.group0_history.  Reading indexes is done using system_schema.indexes which is guarded by checking if schema version is agreed in available nodes. So there are 3 tables, and there is some time to propagate schema changes (adding/removing index) around scylla cluster. We are using concurrent reads which can interlace. For this reason there is a possibility (and it happens in validator tests) that index list is read from the node that has no actual index list. In that case the number of indexes can be wrong up to the next schema change.

This PR removes usage of group0_history.state_id and use schema_version instead. It facilitates possibility to additional validate of index list by providing schema_version retrieved before. In db.is_valid_schema we check if agreed schema_version is equal to the previous one. If not, we will repeat reading indexes again in the next tick interval. We can read schema_version from only one node for triggering indexes check as it will be validated over agreed value further.

This PR removes guard of checking agreed schema_version in db.is_valid_index as it needs to be done in caller. Therefore, this check is done using db.is_valid_schema in monitor_indexes actor.

This PR refactors get_latest_schema_version and adds is_valid_schema messages to the db actor.

Fixes: VECTOR-905
Fixes: VECTOR-906
Fixes: VECTOR-908

### DIFF
--- a/crates/vector-store/src/monitor_indexes.rs
+++ b/crates/vector-store/src/monitor_indexes.rs
@@ -26,7 +26,6 @@ use crate::perf;
 use anyhow::bail;
 use futures::StreamExt;
 use futures::stream;
-use scylla::value::CqlTimeuuid;
 use std::collections::HashSet;
 use std::sync::Arc;
 use std::sync::atomic::AtomicBool;
@@ -42,6 +41,7 @@ use tracing::debug;
 use tracing::error_span;
 use tracing::info;
 use tracing::warn;
+use uuid::Uuid;
 
 pub(crate) enum MonitorIndexes {}
 
@@ -152,8 +152,7 @@ pub(crate) async fn new(
     Ok(tx)
 }
 
-#[derive(PartialEq)]
-struct SchemaVersion(Option<CqlTimeuuid>);
+struct SchemaVersion(Option<Uuid>);
 
 impl SchemaVersion {
     fn new() -> Self {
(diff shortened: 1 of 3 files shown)


crates/vector-store/src/monitor_indexes.rs
```diff
@@ -152,28 +152,52 @@ pub(crate) async fn new(
...
         let schema_version = db.latest_schema_version().await.unwrap_or_else(|err| {
             warn!("unable to get latest schema change from db: {err}");
             None
         });
-        if self.0 == schema_version {
+        if schema_version.is_none() {
+            self.propagation_count = 0;
+        }
+        if self.last_version == schema_version {
+            if self.propagation_count > 0 {
```
knowack1: Why return true while the counter is above 0? That re-reads the index list on every tick as the counter is decremented, instead of re-reading it once, when the counter reaches 0.

knowack1:
I think we can include VECTOR-908 and  VECTOR-906  to the fixes list in cover latter.

crates/vector-store/src/monitor_indexes.rs
ewienik: The PR was redesigned

ewienik:
I've rewritten PR

knowack1 (approved):

knowack1:
Looks good. Approved.

PR #596, opened 2026-09-09
vector-store: pin the index listing to the schema-version coordinator

knowack1:
## Problem

`fts::bm25_boolean_or_query` (VECTOR-905) and `fts::bm25_stop_word_filtering`
(VECTOR-908) fail in the daily validator job with the same panic:

```
Timeout on: the index idx_5 at http://127.0.2.21:6080/api/v1 must be created
```

Both are the same bug. Counting the discovery lines each vector-store node
emits shows the index was never picked up by two of the three nodes — it was
not a slow build:

```
$ grep -ao 'creating the index ksp_[0-9]*\.idx_[0-9]*' ci.log | sort | uniq -c
      3 creating the index ksp_4.idx_4
      1 creating the index ksp_5.idx_5     <-- VECTOR-905
      3 creating the index ksp_6.idx_6
```

VECTOR-908 shows `1 creating the index ksp_9.idx_9`, otherwise identical.

## Root cause

`monitor_indexes` reads the cluster schema version once per tick and skips the
round while that version has not changed, storing the version as soon as it
reads it. The version comes from `system.group0_history`, the listing of
custom indexes from `system_schema.indexes`. Both are node-local, and the
driver routes the two reads independently, so a single round can read them
from two different coordinators.

A group0 command is applied on each node on its own. Under CI load the nodes
drift noticeably — the index-creation transaction in the VECTOR-905 run landed
234 ms apart:

```
00:10:28,396  schema_tables - Altering ksp_5.tbl_5 ...   node 1
00:10:28,401  schema_tables - Altering ksp_5.tbl_5 ...   node 2
00:10:28,630  schema_tables - Altering ksp_5.tbl_5 ...   node 3   (+234 ms)
```

(VECTOR-908: +258 ms.)

A round that reads the version from a node that has applied the `CREATE CUSTOM
INDEX` and the listing from a node that has not completes without error and
without the new index. The stored version is already the post-change one, so
no later tick reads the listing again. The fts tests issue no further DDL
between `create_fts_index` and the wait, so the index stays undiscovered for
the full 60 s timeout.

This is the same failure shape as #568 (`61b90592`, `9e01770a`), which fixed
the *per-index follow-up* reads by failing the round so the caller resets the
version and retries. The top-level listing was never guarded.

## Fix

Report the coordinator that served the schema version and send the index
listing to that same node, so a discovery round is consistent by construction
rather than detectable after the fact.

The invariant becomes: a stored version `V` implies a listing taken at `V` on
some node has been processed. A lagging node now reports its own older
version, which either equals the stored one (round skipped; a later tick finds
the change on another node) or differs from it (round runs against that node's
own consistent listing).

`SingleTargetLoadBalancingPolicy` — which the driver documents for exactly
this case, "useful for queries to node-local system tables" — has an empty
fallback plan, so a pinned node that went away fails the round; the version is
reset and the next tick picks a live node. Failover is preserved.

## Test plan

- [x] New unit test `index_is_discovered_when_the_listing_node_lags_behind_the_version`
      models a fresh/stale node pair. Verified it times out without the fix
      (by reverting the one-line change that threads the coordinator through)
      and passes with it.
- [x] `cargo test --workspace` — 306 + 102 pass, 0 failed.
- [x] `cargo fmt --all --check` and
      `cargo clippy --workspace --all-targets -- -Dwarnings` clean.
- [x] Full validator suite against `scylladb/scylla-nightly:latest` —
      166/166 passed, no timeouts, and every index in the run was discovered
      by all three nodes.

Note on reproduction: the end-to-end race could not be triggered on an
unloaded local cluster (a 40-round probe with per-node pinned sessions
observed no divergence window — schema propagates too fast there). The
diagnosis rests on the CI logs above, and the regression guard is the
deterministic unit test.

## Follow-up, not in this patch

The per-index follow-up reads (`get_index_version`, `get_index_target_type`,
`get_fts_index_params`) still go to driver-chosen nodes. They are safe today
because a missing row fails the round and retries, but pinning them to the
same coordinator would make the whole round coherent and turn those bail paths
into pure safety nets.

Fixes: VECTOR-905
Fixes: VECTOR-908

https://claude.ai/code/session_019zc5zEpHyugPXaLy6CXCXb

### DIFF
--- a/crates/vector-store/src/db.rs
+++ b/crates/vector-store/src/db.rs
@@ -53,6 +53,8 @@ use scylla::client::session_builder::SessionBuilder;
 use scylla::cluster::metadata::ColumnType;
 use scylla::cluster::metadata::NativeType;
 use scylla::cluster::metadata::Table;
+use scylla::policies::load_balancing::NodeIdentifier;
+use scylla::policies::load_balancing::SingleTargetLoadBalancingPolicy;
 use scylla::statement::prepared::PreparedStatement;
 use scylla::value::CqlTimeuuid;
 use secrecy::ExposeSecret;
@@ -80,7 +82,7 @@ type GetDbIndexR = anyhow::Result<(
     mpsc::Sender<DbIndex>,
     mpsc::Receiver<(DbIndexedRow, AsyncInProgress)>,
 )>;
-pub(crate) type LatestSchemaVersionR = anyhow::Result<Option<CqlTimeuuid>>;
+pub(crate) type LatestSchemaVersionR = anyhow::Result<SchemaVersionSnapshot>;
 type GetIndexesR = anyhow::Result<Vec<DbCustomIndex>>;
 type GetIndexVersionR = anyhow::Result<Option<IndexVersion>>;
 type GetIndexTargetTypeR = anyhow::Result<Option<Dimensions>>;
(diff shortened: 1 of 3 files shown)


ewienik:
It duplicates #595

knowack1:
> It duplicates #595

Thanks for pointing out.

knowack1:
Closing as fixed in https://github.com/scylladb/vector-store/pull/595

PR #601, opened 2026-09-11
ci: start siren's add-vectorstore-ver via workflow_dispatch

swasik:
## Motivation

The 1.11.0 release could not hand its cloud images to siren. The dispatch was refused ([run](https://github.com/scylladb/vector-store/actions/runs/34469090804/job/102883602170#step:5:20)):

```
Resource not accessible by personal access token
```

siren's `add-vectorstore-ver` can be started two ways, and both take the same four values — but they are reached through different endpoints, and each endpoint checks a different permission of `SIREN_REPO_TOKEN`:

| trigger | endpoint | permission |
|---|---|---|
| `repository_dispatch` (before) | `POST /repos/{owner}/{repo}/dispatches` | `Contents: write` |
| `workflow_dispatch` (after) | `POST /repos/{owner}/{repo}/actions/workflows/{id}/dispatches` | `Actions: write` |

We were using the first, so the token was being asked for write access to siren's *contents* — the right to push commits to siren — in order to start one workflow there. This asks for the narrower grant instead: announcing a release should not carry the ability to rewrite the repository being announced to.

It is also the likely repair for 1.11.0. `Actions: write` is the intuitive grant to provision for "trigger a workflow in another repository", and a `403` (rather than `404`) tells us the token does reach `scylladb/siren` but lacks what the dispatch endpoint wanted. If that is what it holds, this gets releases dispatching again with no change to the token.

## Changes

- `trigger-siren` now calls the `workflow_dispatch` endpoint with `gh`, which is preinstalled on the runners. No other action covers `workflow_dispatch`, so calling the API directly leaves nothing new to pin and audit, and drops `peter-evans/repository-dispatch`.
- The ref to run on — which `repository_dispatch` did not need — is read from siren's `default_branch` rather than hardcoded, so siren renaming its default branch cannot silently break our releases.
- The four values sent are unchanged, so the GA-only guard and the `defaults.vectorstore` caveat still hold, and a failed dispatch is still re-runnable on its own from the packer manifest artifact.
- `docs/releasing.md` and the job summary say `workflow_dispatch` where they said `repository_dispatch`.

## Needs confirming before merge

**The four `workflow_dispatch` input names are assumed to match the old `client_payload` keys** (`version`, `aws_amd64_ami`, `aws_arm64_ami`, `gcp_amd64_image`). `scylladb/siren` was not readable from the session this was written in, so its `on: workflow_dispatch: inputs:` block could not be checked. If the names differ, the API answers `422 Unexpected inputs provided` — please confirm them against siren's workflow.

## Test plan

- [x] `release.yaml` parses; the step's shell passes `bash -n`
- [x] Step executed locally against a stubbed `gh`: builds the expected body — `{"ref": "master", "inputs": {"version": "1.11.0", "aws_amd64_ami": "ami-0e8e5a026e0ae703a", "aws_arm64_ami": "ami-0fa32fd78fdc8bc56", "gcp_amd64_image": "vector-store-1-11-0-amd64-2026-09-10t01-07-04z"}}` (the real IDs from the 1.11.0 manifest)
- [x] Failure propagation: a rejected dispatch and a failed default-branch lookup each fail the step (`pipefail`), rather than passing silently
- [ ] Input names confirmed against siren's `add-vectorstore-ver.yml`
- [ ] Re-run `trigger-siren` on 1.11.0 and confirm the dispatch returns 204 and the siren PR appears

## Relationship to #600

Independent of [#600](https://github.com/scylladb/vector-store/pull/600), which leaves the call alone and only adds diagnostics and token documentation. Both edit the same region of `release.yaml`, so whichever merges second needs a rebase. If this one lands, #600's text needs `Contents: read and write` changed to `Actions: write` to match.

Refs: VECTOR-854
Refs: VECTOR-853

https://claude.ai/code/session_01JPFzonL9RQCXzPo7SeHkfc

---
_Generated by [Claude Code](https://claude.ai/code/session_01JPFzonL9RQCXzPo7SeHkfc)_

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -278,7 +278,7 @@ jobs:
           asset_content_type: application/json
 
   # Hand the freshly built cloud images over to siren: extract the image IDs
-  # from the packer manifest and send the repository_dispatch that starts
+  # from the packer manifest and start
   # https://github.com/scylladb/siren/actions/workflows/add-vectorstore-ver.yml,
   # which opens a versions.yaml PR in scylladb/siren. A separate job consuming
   # the manifest artifact (the push-docker pattern) keeps the dispatch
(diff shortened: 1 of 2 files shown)


ewienik (approved):

ewienik:
ci failed test: VECTOR-943. restarting

PR #602, opened 2026-09-11
ci: attach the release tarballs and SBOMs to the GitHub release

swasik:
### Motivation

The release workflow builds the per-architecture tarballs and the SBOMs but only ever keeps them as **workflow artifacts**. The one thing it attaches to the release page is the packer manifest — putting the tarballs and SBOMs there is still the manual `scripts/upload-release` step.

That breaks `build-cloud-images`. The packer provisioner installs the service by downloading the tarball from the release page (`packer/files/vector_store_install_image`):

```
https://github.com/scylladb/vector-store/releases/download/$VECTOR_VERSION/vector-store-$VECTOR_VERSION-$arch.tar.gz
```

With nothing attached that URL 404s, and since the download uses `curl -L` without `-f`, curl exits 0 and saves GitHub's error page as `vector-store.tar.gz`. The build then dies one command later in `tar -xvf` with `returned non-zero exit status 2` — an error that says nothing about the missing asset.

So a release can only finish if somebody uploads the tarballs by hand while the workflow is running. That is exactly what happened for [1.11.0](https://github.com/scylladb/vector-store/actions/runs/34469090804): `build-cloud-images` failed on attempts 1 and 2, the assets were uploaded manually at 12:19, and only the third attempt got past packer.

### What this does

**`ci: attach the release tarballs and SBOMs to the GitHub release`** — adds an `upload-release-assets` job that downloads the run's tarball and SBOM artifacts and attaches them to the release, and makes `build-cloud-images` depend on it so the assets are in place before packer needs them. It uploads exactly the set `scripts/upload-release` uploads by hand:

- `vector-store-{version}-{amd64,arm64}.tar.gz`
- `vector-store-{version}.cdx.json`
- `vector-store-docker-{version}-{amd64,arm64}.cdx.json`

The job addresses the release by tag with `gh release upload` rather than through `github.event.release.upload_url`, so a `workflow_dispatch` re-release attaches its assets too, and `--clobber` replaces the assets of an earlier attempt instead of failing on them. A tag with no GitHub release at all fails the job with an explicit message — such a run cannot produce cloud images either way, and failing here says why instead of leaving it to the tar error four minutes into the packer build.

**`ci: upload the packer manifest with gh release upload`** — `actions/upload-release-asset` is archived upstream, takes the `upload_url` only a `release` event carries, and refuses an asset name already attached to the release. Re-running `build-cloud-images` once the manifest is up would fail with `already_exists`, and that job is exactly the one a release re-run repeats. The manifest now goes up the same way as the other assets.

**`packer: fail the image build when the release tarball is missing`** — adds `-f` (plus `--retry 3 --retry-delay 5`) to the provisioner's `curl`, so a missing asset fails at the download with `curl: (22) The requested URL returned error: 404` and the failing URL in the provisioner output, instead of in `tar`.

`docs/releasing.md` documents the new job and why it has to precede the cloud image build; the note saying a `workflow_dispatch` release uploads no release asset is now stale and was corrected.

### Test plan

- [x] Workflow YAML parses; job graph resolves in topological order (`upload-release-assets` after `release-build-docker-archive-sbom`, `build-cloud-images` after both it and `push-docker`)
- [x] Both new `run:` blocks pass `bash -n`; the asset-collection logic simulated for the complete case and for a missing file (fails with the `::error::` line naming the file)
- [x] Artifact name patterns checked against the 13 real artifacts of run 34469090804: `vector-store-<version>-*` matches only the two tarball artifacts — not `vector-store-{amd64,arm64}`, `vector-store-image-<version>-*` or the merged `vector-store-sboms-<version>`
- [x] `curl` behaviour reproduced against a local server: without `-f` exit 0 + a 335-byte HTML file + `tar` exit 2 (the reported failure), with `-f` exit 22 and `The requested URL returned error: 404`
- [ ] Next release: confirm the tarballs and both SBOM kinds appear on the release page before `build-cloud-images` starts, and that packer installs from them

### Note

Not fixed here: attempt 3 of the 1.11.0 run failed in `trigger-siren` with `Resource not accessible by personal access token` — the `SIREN_REPO_TOKEN` secret cannot send the `repository_dispatch` to scylladb/siren (see VECTOR-854). That needs the token's permissions restored, not a workflow change.

Fixes: VECTOR-942

https://claude.ai/code/session_01J77sURMZDAVb8pMXKYhHmb

---
_Generated by [Claude Code](https://claude.ai/code/session_01J77sURMZDAVb8pMXKYhHmb)_

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -217,8 +217,72 @@ jobs:
             sboms/vector-store-docker-${{ env.VECTOR_VERSION }}-*.cdx.json
           if-no-files-found: warn
 
+  # Attach the per-arch tarballs and the SBOMs to the GitHub release. This has
+  # to run before build-cloud-images: the packer provisioner installs the
+  # service by downloading
+  # releases/download/$VECTOR_VERSION/vector-store-$VECTOR_VERSION-$arch.tar.gz
+  # (see packer/files/vector_store_install_image), so with the tarballs missing
+  # from the release page the image build fails.
+  #
+  # Only a run started by publishing a release uploads: there the release the
+  # assets belong to is the one being published, while workflow_dispatch takes
+  # an arbitrary VECTOR_VERSION and would attach freshly built binaries to an
+  # already-published release. Re-running a failed release run keeps its
+  # `release` event, so that recovery path still uploads. The guard is on the
+  # steps rather than the job because a skipped job would skip
+  # build-cloud-images with it.
+  upload-release-assets:
+    needs: [release-build-docker-archive-sbom]
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write     # to attach the assets to the release
+    steps:
+      - name: Download the release tarball artifacts
+        if: github.event_name == 'release'
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          pattern: vector-store-${{ env.VECTOR_VERSION }}-*
+          merge-multiple: true
+          path: assets
+
+      - name: Download the SBOM artifacts
+        if: github.event_name == 'release'
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          pattern: vector-store-sboms-${{ env.VECTOR_VERSION }}-*
+          merge-multiple: true
+          path: assets
+
+      - name: Upload the tarballs and SBOMs to the release
+        if: github.event_name == 'release'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_REPO: ${{ github.repository }}
+        run: |
+          set -euo pipefail
+          # both arch legs generate the same Cargo SBOM, so merge-multiple
+          # leaves a single copy of it beside the per-arch tarballs and docker
+          # SBOMs
+          assets=("assets/vector-store-${VECTOR_VERSION}.cdx.json")
+          for arch in amd64 arm64 ; do
+            assets+=(
+              "assets/vector-store-${VECTOR_VERSION}-${arch}.tar.gz"
+              "assets/vector-store-docker-${VECTOR_VERSION}-${arch}.cdx.json"
+            )
+          done
+          for asset in "${assets[@]}" ; do
+            if [[ ! -f "${asset}" ]] ; then
+              echo "::error::${asset} is missing from the build artifacts"
+              exit 1
+            fi
+          done
+          # the release being published is addressed by its tag rather than
+          # through its upload_url; --clobber replaces assets left by an
+          # earlier attempt of the same run instead of failing on them
+          gh release upload "${VECTOR_VERSION}" "${assets[@]}" --clobber
+
   build-cloud-images:
-    needs: [push-docker]
+    needs: [push-docker, upload-release-assets]
     runs-on: ubuntu-latest
     permissions:
       contents: write     # to attach the packer manifest to the release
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yaml
```diff
@@ -217,8 +217,69 @@ jobs:
...
+            fi
+          done
+          # a workflow_dispatch run has no github.event.release, so address the
+          # release by tag rather than through its upload_url; fail with a
+          # readable message when the tag has no release at all, because the
+          # cloud image build installs the tarball from the release page and
+          # would otherwise fail later with an unrelated tar error
+          if ! gh release view "${VECTOR_VERSION}" > /dev/null 2>&1 ; then
+            echo "::error::no GitHub release for tag ${VECTOR_VERSION}, create it before running the release workflow"
+            exit 1
```
ewienik: Uploading should be possible only when it is release trigger. Shouldn't it?

.github/workflows/release.yaml
```diff
@@ -327,16 +327,23 @@ jobs:
           name: vector-store-cloud-images-${{ env.VECTOR_VERSION }}
           path: packer/packer-manifest.json
 
+      # `gh release upload` rather than actions/upload-release-asset: that
+      # action is archived, takes the upload_url only a release event provides,
+      # and rejects an asset name that is already attached — so re-running this
+      # job after the manifest went up once would fail with `already_exists`.
       - name: Upload Packer manifest to the release
-        if: github.event_name == 'release' && github.event.release
```
ewienik, line 270: uploading should be available only when 'release' trigger. Shouldn't it?

ewienik (changes requested):

.github/workflows/release.yaml
swasik: You're right — done in 6d2deb7.

My reasoning for uploading unconditionally was that `build-cloud-images` can't succeed without the assets on the release page, so a `workflow_dispatch` run would need them too. But that argument doesn't hold: re-running a failed release run keeps its `release` event, so the recovery path that actually matters still uploads. What's left for `workflow_dispatch` is building an arbitrary `VECTOR_VERSION`, and there `--clobber` would replace the published tarballs of an already-shipped release with freshly built binaries. Not worth the risk for a path that has the manual `scripts/upload-release` anyway.

One implementation detail: the guard is on the steps, not on the job. A job-level `if` that evaluates false makes the job `skipped`, and `build-cloud-images` — which now `needs` it for ordering — would be skipped along with it, so a manual dispatch would silently build no cloud images. With step-level guards the job still runs and succeeds as a no-op, and the ordering dependency holds.

This also removes the `gh release view` existence check, since a `release` event always has a release — which resolves the [qodo finding](https://github.com/scylladb/vector-store/pull/602#discussion_r3988692785) on the same lines.

`docs/releasing.md` now states that only a release-triggered run attaches assets, and that a dispatch run builds its images from whatever the release page already carries (failing at the download, naming the URL, if the tarball isn't there).

---
_Generated by [Claude Code](https://claude.ai/code)_

.github/workflows/release.yaml
swasik: Agreed — restored in 6d2deb7. This step keeps the `if: github.event_name == 'release'` guard it had before; dropping it was an unintended consequence of switching the step from `actions/upload-release-asset` to `gh release upload`, since the old action needed `github.event.release.upload_url` and so was implicitly release-only.

The only behaviour change left in that patch is the one it's meant to make: `actions/upload-release-asset` is archived and refuses an asset name already attached, so re-running `build-cloud-images` after the manifest went up once would fail with `already_exists` — and that job is exactly the one a release re-run repeats. `gh release upload --clobber` replaces it instead.

I also reverted the `docs/releasing.md` edit that went with it, so the note about a `workflow_dispatch` release uploading no release asset stands as before.

---
_Generated by [Claude Code](https://claude.ai/code)_

.github/workflows/release.yaml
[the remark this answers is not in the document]
swasik: Valid — `> /dev/null 2>&1` did fold auth, permission and rate-limit failures into "no GitHub release for this tag", which would have sent a release operator off to create a release that already exists.

Fixed in 6d2deb7 by removing the check rather than capturing its stderr. The upload is now guarded to `github.event_name == 'release'` (see [this thread](https://github.com/scylladb/vector-store/pull/602#discussion_r3988977941)), and a `release` event always has a release, so there is nothing to probe for. Any remaining failure surfaces as `gh release upload`'s own error with its diagnostic intact.

---
_Generated by [Claude Code](https://claude.ai/code)_

ewienik (approved):

PR #604, opened 2026-09-16
vector-store: decode a CQL vector off the frame instead of via Vec<f32>

Akvear:
`<Vec<f32>>::deserialize` collects an iterator of `Result` into `Result<Vec<_>, _>`. Such a collection's size hint has a lower bound of 0, because it may stop early on an error, even though the driver's `VectorIterator` reports an exact one. Vec sizes its first allocation from the lower bound, so a 1536-dimension embedding is grown geometrically rather than allocated once, on every vector of every hop of every query.

Decode the frame bytes with chunks_exact, an ExactSizeIterator, so the Vec is sized once. Worth 2.1x QPS on the DiskANN backend that reads vectors back from the base table (36.1 -> 75.0 at 50k).

This was found when benchmarking for the VECTOR-830 [ADR](https://github.com/scylladb/vector-search-docs/pull/33).


crates/vector-store/src/vector.rs
```diff
@@ -74,6 +74,28 @@ enum EmbeddingColumnError {
...
+            dimensions,
+        });
+    }
+
+    Ok(chunks
+        .map(|chunk| {
+            let bytes = chunk.try_into().expect("chunks_exact yields 4 bytes");
+            f32::from_be_bytes(bytes)
+        })
+        .collect())
```
Lorak-mmk, line 98: You could use https://doc.rust-lang.org/std/primitive.slice.html#method.as_chunks , which yields a slice of arrays, eliminating the `try_into` and `expect`.

Lorak-mmk:
Adding a new CqlValue variant for float vector is not something we can do in Rust Driver, as mentioned previously.

However, optimizing our deserialization is an entirely different matter. I opened a PR that significantly improves performance of Vector deser in Rust Driver: https://github.com/scylladb/scylla-rust-driver/pull/1924

Would you mind running your benchmark without this VS PR, but with version of Rust Driver from my PR?

Akvear:
> Would you mind running your benchmark without this VS PR, but with version of Rust Driver from my PR?

Works great! The driver fix fully replaces this change.

| build                         | QPS  |
|-------------------------------|------| 
| no fix, stock driver          | 36.1 |
| this PR's workaround          | 77.3 |
| patched driver, no workaround | 77.2 |

Basically identical performance.

PR #606, opened 2026-09-16
vector-store: size the index capacity threshold for adds in flight

swasik:
`PartitionState::needs_more_capacity()` decides whether to grow the usearch index by
comparing the reserved capacity against `size`, the number of vectors already in the
index, and reserves once fewer than `free_threshold` slots are left. `free_threshold`
was `perf::channel_size()`, which is smaller than the number of adds that can be
missing from `size` when the check runs: the check runs in the index actor, but `size`
is bumped by a worker only once the add has finished, so in between an add is either
waiting on the worker channel (`perf::channel_size()` slots) or running on a worker
(`perf::num_workers()` of them). Approaching a capacity boundary the actor kept
dispatching, usearch rejected the overflow with `Reserve capacity ahead of
insertions!`, and `add()` dropped those vectors with nothing but a warning: no metric,
no retry, and the index still reported `SERVING` at 100% because readiness comes from
full-scan progress rather than from how many vectors were actually indexed.

Size the threshold for that worst case instead: `channel_size + num_workers` covers
everything that can already be in flight, and the `+ 1` covers the add the check itself
is about to let through. Reserving that early keeps every dispatched add inside the
reservation, with no per-add bookkeeping on the hot path.

The free-space subtraction also becomes `saturating_sub`. Release builds do not check
overflow, so a single overshoot wrapped `capacity - size` to `usize::MAX`, leaving
`needs_more_capacity()` returning `None` forever and failing every later add.

### Why now

The race predates the usearch 2.26 upgrade — 2.22 has the same guard in
`index_gt::add` — but 2.22 hid it. Its `reserve(N)` rounded the member count up to
`ceil2(N * 3 / 2)`, so loading 10M rows crossed roughly three capacity boundaries
instead of ten and each one had room to spare:

| requested | usearch 2.22.0 | usearch 2.26.2 |
| --- | --- | --- |
| `reserve(200_000)` | 524_288 | 200_000 |
| `reserve(2_000_000)` | 4_194_304 | 2_000_000 |

Reserving exactly what was asked for (usearch PR #777, released in 2.26.1) makes every
boundary tight. Loading the 10M-vector Cohere dataset on 2.26.2 lost 69 vectors this
way, measured as the permanent step in `index_modified{operation="insert"}` minus
`index_size` at each 1,000,000 boundary.

The fix is independent of usearch's reservation policy — `reserve()` already reads the
real capacity back via `idx.capacity()` — so it is correct on 2.22 and 2.26 alike.

### Testing

Two regression tests. `capacity_check_leaves_room_for_every_add_in_flight` drives the
check with the workers lagging by the full in-flight window, for both a single worker
and an eight-worker runtime; it fails on the old threshold with `dispatched 4097 adds
against a capacity of 4096` and passes with the fix.
`overshooting_capacity_asks_for_more_instead_of_wrapping` covers the `usize` wrap.

`cargo fmt --all --check` clean, `cargo clippy --workspace --all-targets -- -Dwarnings`
clean, `cargo test --workspace` 318 unit + 102 integration passing.

Fixes: VECTOR-945
Refs: VECTOR-720


swasik:
> Fixes: [VECTOR-945](https://scylladb.atlassian.net/browse/VECTOR-945)

@ewienik please take a look.

[VECTOR-945]: https://scylladb.atlassian.net/browse/VECTOR-945?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

ewienik (commented):
The root cause is too low free_threshold (currently perf::channel_size). What about increasing this threshold (to perf::channel_size + perf::num_worker o 2xperf::channel_size) to calculate this threshold correctly to the reasonable value rather than dynamically count processing messages?

swasik:
> The root cause is too low free_threshold (currently perf::channel_size). What about increasing this threshold (to perf::channel_size + perf::num_worker o 2xperf::channel_size) to calculate this threshold correctly to the reasonable value rather than dynamically count processing messages?

I completely agree. Fixed.

ewienik (approved):

PR #608, opened 2026-09-17
vector-store: put the DiskANN graph behind a GraphStore trait

Akvear:
Move each node's existence, liveness and adjacency list out of the provider and behind a GraphStore, so the graph's payload can live somewhere other than RAM.

InmemGraphStore is the only implementation, and needs no special locks as it awaits nothing, so its map lock already spans each call.

The behaviour stays identical to master.

Refs: VECTOR-833


crates/vector-store/src/vs_index/diskann/mod.rs
```diff
@@ -129,8 +129,11 @@ impl VsIndexFactory for DiskannIndexFactory {
             ),
             DiskannBackendKind::Scylla => {
                 let source = Arc::new(scylla::BaseTableSource::new(Arc::clone(&table), db_index));
+                let graph = Arc::new(scylla::InmemGraphStore::new(
+                    params.config.max_degree().get(),
+                ));
```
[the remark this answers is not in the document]
Akvear: AFAIK all nodes are removed before RemovePartition is called.

crates/vector-store/src/vs_index/diskann/scylla.rs
```diff
@@ -247,48 +253,238 @@ impl DiskannBackend for ScyllaBackend {
...
+}
 
-    fn is_live(&self) -> bool {
-        matches!(self, Self::Live(_))
+/// A [`GraphStore`] holding the whole graph in RAM.
+#[derive(Debug)]
+pub(super) struct InmemGraphStore {
+    nodes: RwLock<BTreeMap<(PartitionId, PrimaryId), Node>>,
+    max_degree: usize,
+}
```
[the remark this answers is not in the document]
Akvear: AFAIK all nodes are removed before RemovePartition is called.

crates/vector-store/src/vs_index/diskann/scylla.rs
```diff
@@ -247,48 +253,238 @@ impl DiskannBackend for ScyllaBackend {
...
+
 #[derive(Debug)]
-enum Node {
-    Live(AdjacencyList<PrimaryId>),
-    Dead(AdjacencyList<PrimaryId>),
+struct Node {
+    liveness: Liveness,
+    edges: AdjacencyList<PrimaryId>,
 }
```
ewienik, line 354: With this change we revert to the boolean flag solution. Is it needed somewhere?

Akvear: reverted to enum

crates/vector-store/src/vs_index/diskann/scylla.rs
```diff
@@ -279,16 +366,131 @@ impl Node {
...
+        partition_id: PartitionId,
+        id: PrimaryId,
+        edges: &[PrimaryId],
+    ) -> anyhow::Result<()> {
+        let mut nodes = self.nodes.write().unwrap();
+        let Some(node) = nodes.get_mut(&(partition_id, id)) else {
+            return Ok(());
+        };
+        // `extend_from_slice` deduplicates through `push`, which is what keeps
+        // the stored list unique.
```
QuerthDP: I think it should be added right over `list.extend_from_slice(edges)` to have any potential value and not be misleading. Also should be added in `append` since it's the same situation. Having that said I think it is not really needed in either of there at all, and even if, then we should rather care about that behavior in `append`, since in `set` we clear the list anyway each time.

crates/vector-store/src/vs_index/diskann/scylla.rs
```diff
@@ -1139,4 +1324,89 @@ mod tests {
...
+        assert_eq!(edges(&store, node(1)).await, Some(vec![node(2)]));
+    }
+
+    #[tokio::test]
+    async fn create_revives_a_tombstone_with_no_edges() {
+        let store = InmemGraphStore::new(8);
+
+        store.create(part(), node(1)).await.unwrap();
+        store.set(part(), node(1), &[node(2)]).await.unwrap();
+        store.mark_dead(part(), node(1)).unwrap();
```
QuerthDP: I think we should check there if we're really dead afterwards.

crates/vector-store/src/vs_index/diskann/scylla.rs
```diff
@@ -1139,4 +1324,89 @@ mod tests {
...
+
+        store.create(part(), node(1)).await.unwrap();
+        assert!(store.is_live(part(), node(1)));
+        assert_eq!(edges(&store, node(1)).await, Some(vec![]));
+    }
+
+    #[tokio::test]
+    async fn writes_to_an_absent_node_are_dropped() {
```
QuerthDP: What writes and how does this test check that?
I see no asserts that the set or append did fail. The only assert is after clear which should obviously show `None`.

QuerthDP (commented):
Looks generally good. 
Please explain why did we change from futures to async and does it change anything to worry about.

Also verify the `write` test, I think it's wrong.

crates/vector-store/src/vs_index/diskann/scylla.rs
Akvear: not sure what this test is doing at all – probably a claude artifact missed when rebasing.

Akvear:
Changes:
- removed `extend_from_slice` comment
- added assert to `create_revives_a_tombstone_with_no_edges()`
- removed `writes_to_an_absent_node_are_dropped()`

